PR #126, opened 2024-04-24
Remove Mac from Supported Destinations for the FreeAPS target

bjornoleh:
It was noted that Trio can be installed to Mac from TestFlight, and that this might be undesirable. This PR is an attempt to block that.


MikePlante1:
This does work for blocking building to your Mac from Xcode:

| dev | this PR |
| --- | --- |
|   |  |

But it seems to still show up in TestFlight:

| Loop | this PR |
| --- | --- |
|  |  |

bjornoleh:
So it doesn’t look like this PR does what it set out to do (block TF install on Mac).

I did look a little further into this, but it wasn’t successful: 

https://github.com/bjornoleh/Trio/commit/2ae0fa1618aa7306e44f040c983f5808a38eca24

It was probably on the right track, but in fact blocked all installs from TF, including install to iPhone, when I limited the device family to only support iPhone. 

The problem is probably related to this:

https://developer.apple.com/library/archive/qa/qa1623/_index.html

I think I’ll just close this PR for now.

aug0211:
I would prefer that we not block users from doing this.

I deployed Trio to my Mac just yesterday. I needed to develop a rather in-depth Shortcut. Being able to deploy Trio to my Mac meant that Trio Shortcuts became available to me to plug in to the complex Shortcut I was developing on my Mac. This of course only happens if we can build Trio to the Mac.

I created the shortcut on my Mac with a full keyboard, mouse, and large screen. Once I was finished, I sent the Shortcut from my Mac to the phone running Trio for live usage.

Trio being deployed to my Mac saved me many hours of much frustration in this use case.

bjornoleh:
@aug0211 , that’s a great use case, and I don’t see any immediate issues with allowing to install Trio on Mac. Closing this for now, and will let others pick this up later if needed.

PR #177, opened 2024-05-13
Shortcuts - refactoring and add new functionalities

avouspierre:
## Shortcuts Trio 
this PR refactors the shortcuts and adds new shortcuts for Trio. 
### Refactoring 
- Same organisation of files
- Remove junk code
- Add internationalization of shortcuts with the use of catalog string of Xcode 

### List of shortcuts available 
- Override preset selection, application and cancelation
- Add Carbs shortcuts
- Carbs preset list and display all carbs, fat and proteins associated
- Create And Apply a temporary target with the definition of target, duration, date of start
- Apply a preset temporary target 
- Get the state of Trio : BG, IOB, COB 
- Command a bolus

Some examples described in the picture : 

When your iPhone alarm clock is ringing a Profile override is started/canceled.
• When you get to your gym a Profile Override is started/canceled.
• When you step inside your car (which has CarPlay) a profile override is started/canceled.
• When you start an exercise a profile override is started.
• When you end an exercise the active profile override is canceled.
• .

| Select a carb preset, add carbs associated and define a override preset | List of shortcuts available | Cancel temporary preset and/or override preset and adds carbs in the past |
|--------|--------|--------|
| IMG_93A644ADB473-1 | IMG_E62CF95824EE-1 | IMG_F4A8EE137CD2-1 |
| temporary preset with a date in 60 minutes | When I start a fitness activity, start sport override except if my BG > 180, wait 30 min and notify me | When you get a specific sms from your parent a specific profile override is started/canceled |
| IMG_16251E5F9A67-1| RPReplay_Final1714157264| https://github.com/nightscout/Open-iAPS/assets/4339604/5f0a81a8-4445-449a-94cf-6805fbf316b9|

New UI for config shortcuts : 

| config | config |
|--------|--------|
| Simulator Screenshot - iPhone 15 - 2024-05-01 at 18 23 50 | Simulator Screenshot - iPhone 15 - 2024-05-01 at 18 24 00 | 

Detail of the commit (squash all commits about shortcuts) : 
Refactoring the shortcuts with adding :
- same organisation of files
- remove junk code
- add internationalization of shortcuts
- add AddCarbs shortcuts
- add createAndApply temp target
- Add override preset selection, application and cancelation
- Add Carbs preset list and display all carbs, fat and proteins associated
- fix carb list shortcut issue
- fix carb added with fat/protein  shortcut
- add bolus shortcuts with guardrails
- harmonise temporary target labels
- add in Trio a shortcut config allowing to authorize bolus enact and guardrails.
- Add localisation for the UI shortcuts config view and shortcuts
- minor improvements to avoid warning in build


mountrcg:
@avouspierre I made some suggestions to harmonize the texts and also list all shortcuts for the Trio app
https://github.com/mountrcg/Trio/commits/dev%2Bshortcuts/

mountrcg (changes requested):
see commits I proposed in link above

LiroyvH:
Q: I see temp override and temp target. I believe currently they're just the temp targets. Are overrides supposed to be what's called a Profile in-app?

mountrcg:
> Q: I see temp override and temp target. I believe currently they're just the temp targets. Are overrides supposed to be what's called a Profile in-app?

That was one thing I corrected, calling them overrides without temp as they can be also without an end date.

avouspierre:
thanks @mountrcg  ! I updated the PR with your contribution + 1 more fix for wording

mountrcg:
@avouspierre the Bolus shortcut when run as shortcut and not integrated as an action does not ask for bolus amount. Instread throws an error. Carbs works fine this way. Could you add that to Bolus as well
| Bolus | Carbs |
| --- | --- |
| Simulator Screenshot - Dev 12 mini - 2024-05-17 at 09 40 12 | Simulator Screenshot - Dev 12 mini - 2024-05-17 at 09 40 35 |

aug0211:
> @avouspierre the Bolus shortcut when run as shortcut and not integrated as an action does not ask for bolus amount. Instread throws an error. Carbs works fine this way. Could you add that to Bolus as well
> 
> Bolus	Carbs
> Simulator Screenshot - Dev 12 mini - 2024-05-17 at 09 40 12	Simulator Screenshot - Dev 12 mini - 2024-05-17 at 09 40 35

The fix for this is to not mark `bolusQuantity` as optional in the `BolusIntent` parameter declaration (line 28 of `BolusIntent.swift`). After doing this, we must remove the `guard` (and matching `else{...}`) at line 78 of the same file.

Will add these notes in a review.

FreeAPS/Sources/Shortcuts/Bolus/BolusIntent.swift
```diff
@@ -0,0 +1,80 @@
...
+    @Parameter(
+        title: LocalizedStringResource("Amount", table: "ShortcutsDetail"),
+        description: LocalizedStringResource("Bolus amount in U", table: "ShortcutsDetail"),
+        controlStyle: .field,
+        inclusiveRange: (lowerBound: 0, upperBound: 200),
+        requestValueDialog: IntentDialog(LocalizedStringResource(
+            "What is the value of the bolus amount in insulin units ?",
+            table: "ShortcutsDetail"
+        ))
+    ) var bolusQuantity: Double?
```
aug0211, line 28: In order to allow users to input the bolus amount (as noted by @mountrcg [here](https://github.com/nightscout/Trio/pull/177#issuecomment-2116948179)), do not mark this as optional. Achieve this by removing the `?`

So, change this line from:
`) var bolusQuantity: Double?`

To:
`) var bolusQuantity: Double`

We'll then make changes down below at line 50 to remove the `guard` and matching `else {...} ` since this is no longer optional.

See my comment [here](https://github.com/nightscout/Trio/pull/177#issuecomment-2127028568) confirming the same.

FreeAPS/Sources/Shortcuts/Bolus/BolusIntent.swift
```diff
@@ -0,0 +1,80 @@
...
+        }, otherwise: {
+            Summary("Immediately applying \(\.$bolusQuantity) U", table: "ShortcutsDetail") {
+                \.$confirmBeforeApplying
+            }
+        })
+    }
+
+    @MainActor func perform() async throws -> some ProvidesDialog {
+        do {
+            guard let amount: Double = bolusQuantity else {
```
aug0211, line 50: In order to allow users to input the bolus amount (as noted by @mountrcg https://github.com/nightscout/Trio/pull/177#issuecomment-2116948179), we need to make `bolusQuantity` mandatory, and not optional. See the comment above at line 28.

Since `bolusQuantity` is no longer optional, we must remove this `guard` and `else {...}` here.

So, change this line from:
`guard let amount: Double = bolusQuantity else {`

To:
`let amount: Double = bolusQuantity`

We then need to remove the contents of the `else {...}`.

So, delete lines 51, 52, 53, 54, 55, 56, 57:
```
                return .result(
                    dialog: IntentDialog(LocalizedStringResource(
                        "Amount not correctly defined",
                        table: "ShortcutsDetail"
                    ))
                )
            }
```

See my comment https://github.com/nightscout/Trio/pull/177#issuecomment-2127028568 confirming the same.

aug0211 (dismissed):

FreeAPS/Sources/Shortcuts/Bolus/BolusIntent.swift
aug0211: I committed this to the branch and tested on a physical device 👍

FreeAPS/Sources/Shortcuts/Bolus/BolusIntent.swift
aug0211: I committed this to the branch and tested on a physical device 👍

aug0211:
Committed requested changes to the PR and tested on a physical device. Ready to merge in my opinion.

**@marionbarker or @bjornoleh could you give a 👍 and merge if you agree?**

IMG_3537
IMG_3538
IMG_3539
IMG_3540

.github/workflows/build_trio.yml
Sjoerd-Bo3, line 1: Seems like this doesn’t belong i this PR

Trio.xcworkspace/xcshareddata/swiftpm/Package.resolved
Sjoerd-Bo3, line 1: This should also not be in here

Sjoerd-Bo3 (commented):
Didn’t look at the localisations. And do the tables work with crowdin? 
But found 2 foreign changes that should not be in this PR.
For the rest LGTM

bjornoleh:
I merged dev into `dev+shortcuts`locally and built with Xcode.

Then I got the following unstaged changes:

```
% git diff
diff --git a/FreeAPS/Sources/Localizations/ShortcutsDetail/ShortcutsDetail.xcstrings b/FreeAPS/Sources/Localizations/ShortcutsDetail/ShortcutsDetail.xcstrings
index afd4d815..804ff32c 100644
--- a/FreeAPS/Sources/Localizations/ShortcutsDetail/ShortcutsDetail.xcstrings
+++ b/FreeAPS/Sources/Localizations/ShortcutsDetail/ShortcutsDetail.xcstrings
@@ -538,6 +538,7 @@
       }
     },
     "Amount not correctly defined" : {
+      "extractionState" : "stale",
       "localizations" : {
         "de" : {
           "stringUnit" : {
```

The app built ok and is running on my phone, but I haven't done any actual testing yet.

bjornoleh:
Quick comment:

When limiting bolus to insulin recommendation, a more specific info string could be presented in the shortcut.

Tested 1 U bolus:

Recommended was 0,5 U

Response:
"A bolus command of 0,5 U of insulin was sent"

Suggestion: Add info to indicate that the bolus was limited to xx U due to the "Limit by insulin suggestion estimation"

Another comment: this settings string could possibly be improved/shortened: "Limit by insulin suggestion estimation"

bjornoleh:
Regarding the uncommitted changes and sync with dev, I opened this PR:
https://github.com/avouspierre/Trio/pull/1

.github/workflows/build_trio.yml
bjornoleh: Seems like this will be solved by merging https://github.com/avouspierre/Trio/pull/1, which includes a sync with dev.

FreeAPS/Sources/Shortcuts/Carbs/carbPresetResult.swift
```diff
@@ -0,0 +1,42 @@
+import AppIntents
+import Foundation
+
+@available(iOS 16.0, *) struct carbPresetResult: AppEntity, Identifiable {
```
dnzxy, line 4: Please capitalize this `struct` name and the file name to conform to how other `struct`s in the project are named to far, i.e. `carbPresetResult` -> `CarbPresetResult`

FreeAPS/Sources/Shortcuts/Override/overridePresetEntity.swift
```diff
@@ -0,0 +1,33 @@
+import AppIntents
+import Foundation
+import Intents
+import Swinject
+
+@available(iOS 16.0, *) struct overridePreset: AppEntity, Identifiable {
```
dnzxy, line 6: Please capitalize the `struct` names here and the file name to conform to how other `struct`s in the project are named to far, i.e. 
- `overridePresetEntity.swift` -> `OverridePresetEntity.swift`
- `overridePreset` -> `OverridePreset`
- `overridePresetsQuery` -> `OverridePresetsQuery`

FreeAPS/Sources/Shortcuts/Override/overridePresetsIntentRequest.swift
```diff
@@ -0,0 +1,91 @@
+import CoreData
+import Foundation
+
+@available(iOS 16.0, *) final class OverridePresetsIntentRequest: BaseIntentsRequest {
```
dnzxy, line 4: Please capitalize the file name.

FreeAPS/Sources/Shortcuts/TempPresets/tempPresetEntity.swift
```diff
@@ -29,3 +29,21 @@ import Swinject
         return tp
     }
 }
+
+@available(iOS 16.0, *) struct tempPresetsQuery: EntityQuery {
```
dnzxy, line 33: Please capitalize this `struct` name and the file name to conform to how other `struct`s in the project are named to far, i.e. 
-`tempPresetEntity.swift` -> `TempPresetEntity.swift`
- `tempPresetsQuery`-> `TempPresetsQuery`

Trio.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -48,7 +48,7 @@
     {
```
dnzxy, line 48: Package.resolved should not be commited.

FreeAPS/Sources/Localizations/Shortcuts/en.lproj/AppShortcuts.strings
```diff
@@ -0,0 +1,36 @@
+/*
+  AppShortcuts.strings
```
dnzxy, line 2: Given the comment `has to be redone after finalizing the shortcuts` in this file – is this final and okay to merge?

FreeAPS/Sources/Models/FreeAPSSettings.swift
```diff
@@ -233,6 +252,10 @@ extension FreeAPSSettings: Decodable {
             settings.lockScreenView = lockScreenView
         }
 
+        if let bolusShortcut = try? container.decode(BolusShortcutLimit.self, forKey: .bolusShortcut) {
```
dnzxy, line 255: Please catch errors here.

FreeAPS/Sources/Shortcuts/Bolus/BolusIntentRequest.swift
```diff
@@ -0,0 +1,61 @@
...
+import CoreData
+import Foundation
+
+@available(iOS 16.0,*) final class BolusIntentRequest: BaseIntentsRequest {
+    private var suggestion: Suggestion? {
+        fileStorage.retrieve(OpenAPS.Enact.suggested, as: Suggestion.self)
+    }
+
+    func bolus(_ bolusAmount: Double) throws -> LocalizedStringResource {
+        var bolusQ: Decimal = 0
```
dnzxy, line 11: What is a `bolusQ` ?

FreeAPS/Sources/Shortcuts/Carbs/CarbPresetIntentRequest.swift
```diff
@@ -0,0 +1,79 @@
...
+                name: $0.dish ?? "-",
+                carbs: ($0.carbs ?? 0.0) as! Double,
+                fat: ($0.fat ?? 0.0) as! Double,
+                protein: ($0.protein ?? 0.0) as! Double
+            )
+        }
+    }
+
+    func addCarbs(
+        _ quantityCarbs: Double,
```
dnzxy, line 31: Shouldn't we give these function arguments a proper name? When calling this function you would do `addCarbs(20.0, 10.0, 15.0, '2024-06-06 14:35:22 +0000')`and it might be a bit less error prone to know which parameter is which? Not sure, what do you think? :)

FreeAPS/Sources/Shortcuts/Override/overridePresetsIntentRequest.swift
```diff
@@ -0,0 +1,91 @@
...
+            return uuid.contains(id) }
+
+        return UUIDTempTarget.compactMap { override in
+            guard let id = override.id, let name = override.name else { return nil }
+            return overridePreset(id: id, name: name)
+        }
+    }
+
+    func fetchOne(_ uuid: overridePreset.ID) -> overridePreset? {
+        let UUIDTempTarget = overrideList.filter { uuid == $0.id! }
```
dnzxy, line 37: Should the id here be gracefully unwrapped to avoid crashes in case it is `nil`?

FreeAPS/Sources/Shortcuts/Override/overridePresetsIntentRequest.swift
```diff
@@ -0,0 +1,91 @@
...
+        guard let OneTempTarget = UUIDTempTarget.first,
+              let id = OneTempTarget.id,
+              let name = OneTempTarget.name
+        else { return nil }
+
+        return overridePreset(id: id, name: name)
+    }
+
+    func enactTempOverride(_ presetTarget: overridePreset) throws -> Bool {
+        let id_ = presetTarget.id
```
dnzxy, line 47: Why the postfix underscore?

FreeAPS/Sources/Modules/ShortcutsConfig/ShortcutsConfigStateModel.swift
```diff
@@ -0,0 +1,39 @@
...
+//
+import SwiftUI
+
+extension ShortcutsConfig {
+    final class StateModel: BaseStateModel<Provider> {
+        @Published var allowBolusByShortcuts: Bool = false
+        @Published var maxBolusByShortcuts: BolusShortcutLimit = .noAllowed
+
+        override func subscribe() {
+            // allowBolusByShortcuts = (maxBolusByShortcuts != .noAllowed)
```
dnzxy, line 15: Remove unused code

FreeAPS/Sources/Modules/ShortcutsConfig/View/ShortcutsConfigView.swift
```diff
@@ -0,0 +1,68 @@
...
+                    .disabled(!state.allowBolusByShortcuts)
+                }
+            }
+            .onAppear(perform: configureView)
+            .navigationTitle(String(localized: "Shortcuts config", table: "ShortcutsDetail"))
+            .navigationBarTitleDisplayMode(.automatic)
+        }
+
+        private func openShortcutsApp() {
+            let shortcutsURL = URL(string: "shortcuts://")!
```
dnzxy, line 47: In case this does not work, it will cause crashes.

Maybe do something like
```swift
if let shortcutsURL = URL(string: "shortcuts://") {
    if UIApplication.shared.canOpenURL(shortcutsURL) {
        UIApplication.shared.open(shortcutsURL, options: [:], completionHandler: nil)
    }
} else {
    //  handle error
    // maybe some log?
}
```

FreeAPS/Sources/Shortcuts/Bolus/BolusIntentRequest.swift
```diff
@@ -0,0 +1,61 @@
...
+    func getLastBG() throws -> String? {
+        let glucose = glucoseStorage.recent()
+        guard let lastGlucose = glucose.last, let glucoseValue = lastGlucose.glucose else { return nil }
+        let units = settingsManager.settings.units
+
+        return glucoseFormatter
+            .string(from: Double(
+                units == .mmolL ? glucoseValue
+                    .asMmolL : Decimal(glucoseValue)
+            ) as NSNumber)! + " " + units.rawValue
```
dnzxy, line 47: Should we gracefully unwrap here and provide a default value?

dnzxy (commented):
Super small findings, nothing major.

Great job, kudos! This looks awesome overall. I am no Shortcut expert but since this has been in use by multiple users for a long time in-vitro and in-vivo, I trust this will work, so will only keep my review to code.

FreeAPS/Sources/Modules/ShortcutsConfig/View/ShortcutsConfigView.swift
AndreasStokholm: Looks like this has been addressed

dnzxy: Is there a commit I have missed to review again? 😮

FreeAPS/Sources/Localizations/Shortcuts/en.lproj/AppShortcuts.strings
Sjoerd-Bo3: @avouspierre Can you answer this please?

FreeAPS/Sources/Shortcuts/Bolus/BolusIntentRequest.swift
Sjoerd-Bo3: @dnzxy Fixed

FreeAPS/Sources/Modules/ShortcutsConfig/ShortcutsConfigStateModel.swift
Sjoerd-Bo3: @dnzxy Fixed

FreeAPS/Sources/Shortcuts/TempPresets/tempPresetEntity.swift
Sjoerd-Bo3: @dnzxy Fixed

FreeAPS/Sources/Shortcuts/Carbs/carbPresetResult.swift
Sjoerd-Bo3: @dnzxy Fixed

Trio.xcworkspace/xcshareddata/swiftpm/Package.resolved
bjornoleh, line 1: I don't think this file should be deleted, there was just a rather inconsequential change to the file that should not have been committed in the first place.

FreeAPS/Sources/Shortcuts/TempPresets/tempPresetEntity.swift
dnzxy: Looking at the file tree and your recent commits this actually has not been resolved. The file name is still lowercase.

FreeAPS/Sources/Shortcuts/Carbs/carbPresetResult.swift
dnzxy: File name is still lowercase.

Sjoerd-Bo3: @dnzxy Xcode was misbehaving, fixed now

FreeAPS/Sources/Shortcuts/Override/overridePresetEntity.swift
Sjoerd-Bo3: @dnzxy Xcode was misbehaving, fixed now

FreeAPS/Sources/Shortcuts/TempPresets/tempPresetEntity.swift
Sjoerd-Bo3: @dnzxy Xcode was misbehaving, fixed now

FreeAPS/Sources/Shortcuts/Override/overridePresetsIntentRequest.swift
Sjoerd-Bo3: @dnzxy Xcode was misbehaving, fixed now

FreeAPS/Sources/Shortcuts/Override/overridePresetsIntentRequest.swift
Sjoerd-Bo3: @dnzxy Fixed

FreeAPS/Sources/Shortcuts/Carbs/CarbPresetIntentRequest.swift
Sjoerd-Bo3: Fixed - parameterised

FreeAPS/Sources/Shortcuts/Bolus/BolusIntentRequest.swift
Sjoerd-Bo3: fixed

FreeAPS/Sources/Shortcuts/Override/overridePresetsIntentRequest.swift
Sjoerd-Bo3: fixed

FreeAPS/Sources/Localizations/Shortcuts/en.lproj/AppShortcuts.strings
Sjoerd-Bo3: CleanShot 2024-06-14 at 20 03 43@2x
These are actually not used

Sjoerd-Bo3: @dnzxy Can you look into these localised strings for me?

FreeAPS/Sources/Shortcuts/Bolus/BolusIntentRequest.swift
Sjoerd-Bo3: renamed to BolusQuantity for clarity

FreeAPS/Sources/Localizations/Shortcuts/en.lproj/AppShortcuts.strings
avouspierre: update shortcuts strings

Sjoerd-Bo3: Tnx mate!

bjornoleh (changes requested):
I tested briefly in sim mode on an actual phone. All shortcuts work (presets must be defined first). But, there is a critical issue with “New Trio Temptarget”. The TT is not set according to the users units when mmol/L is used. Instead, the mmol/18 value is used. When testing 6 mmol/L as temp target, the result was 0,3 mmol/L

avouspierre:
> I tested briefly in sim mode on an actual phone. All shortcuts work (presets must be defined first). But, there is a critical issue with “New Trio Temptarget”. The TT is not set according to the users units when mmol/L is used. Instead, the mmol/18 value is used. When testing 6 mmol/L as temp target, the result was 0,3 mmol/L

Which unit do you choose in the shortcuts @bjornoleh ?

bjornoleh:
> > I tested briefly in sim mode on an actual phone. All shortcuts work (presets must be defined first). But, there is a critical issue with “New Trio Temptarget”. The TT is not set according to the users units when mmol/L is used. Instead, the mmol/18 value is used. When testing 6 mmol/L as temp target, the result was 0,3 mmol/L
> 
> Which unit do you choose in the shortcuts @bjornoleh ?

It was not clear that I should change units in the shortcut, I only ran it unmodified first. I had the impression it would get the units from Trio. 

But now I also tested to select mmol in the shortcut, and this gave the same result.

avouspierre:
@bjornoleh  Found the issue and fix it (the temp target should be stored only in mgdL).

bjornoleh:
Thanks! I also think the default should be Ask Each Time, otherwise users don’t know about this setting. 

Even better: read units from Trio settings :-)

image

bjornoleh:
> @bjornoleh Found the issue and fix it (the temp target should be stored only in mgdL).

Fix confirmed! 

Tested setting mmol: rendered correctly as mmol with same units in Trio

Also tested setting mg/dL with mmol in Trio, which of course gave 1/18 of the entered target

And changing to mgdL in both Trio and shortcut cave the correct result as well.

bjornoleh (dismissed):

bjornoleh:
> Thanks! I also think the default should be Ask Each Time, otherwise users don’t know about this setting.
> 
> Even better: read units from Trio settings :-)
> 
> image

Would it be possible to let the shortcut read the Trio setting for BG units?

dnzxy (approved):
Based on code review this LGTM now.

Thank you @Sjoerd-Bo3 for addressing the small cosmetical findings.
Thank you @avouspierre for quickly dealing with the TempTarget mmol/L / mg/dl issue.

Let's 🚢  this.

dnzxy (approved):
Re-approving after changing target branch from `dev` to `alpha`.

@bjornoleh @aug0211 please review again and if okay re-approve.

bjornoleh:
I still wonder if the shortcut could get units from the app settings. This is a source of potential issues, where a temp target can be set lower than intended. Since the shortcuts can get the bolus recommendation, I would assume it’s possible to get BG units too.

Sjoerd-Bo3:
> assume

Maybe make an issue for it and target this in a seprate pr?

It works now and your solution is a nice to have.

bjornoleh:
> > assume
> 
> Maybe make an issue for it and target this in a seprate pr?
> 
> It works now and your solution is a nice to have.

We could do that, but I would perhaps say that an effective way to avoid TTs being set at 1/18 of the intended value is a little more than "nice to have", there are safety implications, especially when used with "Low Temptarget Lowers Sensitivity". Every TT would effectively be set to the minimum of 80 mg/dL and typically result in sensitivityRatio being set to 1,5.

bjornoleh (dismissed):
Approving based on testing.

FreeAPS/Sources/Shortcuts/TempPresets/CancelTempPresetIntent.swift
mountrcg, line 1: Cancelling a non-existent TT will result in a success message on iPhone and also a history entry for TT cancelled, even though no TT was active. I would suggest to incorporate message that no TT was active.

mountrcg (commented):
Cancelling a non-existent Override will result in a success message on iPhone I would suggest to incorporate message that no Override was active.

FreeAPS/Sources/Shortcuts/TempPresets/CancelTempPresetIntent.swift
Sjoerd-Bo3: Working on this

Sjoerd-Bo3:
Added this to the TT and Override Cancel actions:
- Displays name of the TT / Override that is canceled
- Display that no TT / Override was active so nothing is canceled
image
image
image
image

dnzxy (approved):

bjornoleh:
Minor mini observation:

Setting a TT of 5,5 mmol, the summary is stating that a TT of 6 is set:

image

It is set correctly in Trio:

image

bjornoleh:
Another observation:

If an override within name is set, the message “No active Override to cancel” is displayed. But the override is correctly cancelled in Trio

An unnamed That is cancelled with the correct message.

Sjoerd-Bo3:
> Another observation:
> 
> If an override with no name is set, the message “No active Override to cancel” is displayed. But the override is correctly cancelled in Trio
> 
> An unnamed temp target is cancelled with the correct message.

Yep true, if the name is empty it will indeed, will see if I can catch the more cleverly

Sjoerd-Bo3:
> Minor mini observation:
> 
> Setting a TT of 5,5 mmol, the summary is stating that a TT of 6 is set:
> 
> image
> 
> It is set correctly in Trio:
> 
> image

I see the Issue, your glucose unit is not correctly recognised, that's why it says '6-'
``` "Are you sure to create a TempTarget with \(targetChoiceName) \(unitName) for \(durationChoiceName) min at \(dateChoiceName) ?"```

**Comment by bjornoleh** [comment]
> > Minor mini observation:
> > Setting a TT of 5,5 mmol, the summary is stating that a TT of 6 is set:
> > image
> 
> I see the Issue, your glucose unit is not correctly recognised, that's why it says '6-' ` "Are you sure to create a TempTarget with \(targetChoiceName) \(unitName) for \(durationChoiceName) min at \(dateChoiceName) ?"`

I can confirm that manually setting units in the shortcut displays the message correctly for mmol:

image

**Comment by bjornoleh** [comment]
However it is strange that the mmol value is correctly converted to mgdL when saved in Trio (target values are always stored as mgdL), but units are not recognised for the message?

**Comment by Sjoerd-Bo3** [comment]
> However it is strange that the mmol value is correctly converted to mgdL when saved in Trio (target values are always stored as mgdL), but units are not recognised for the message?

I dove in to it yesterday and not sure how to fix but it is releated to you not getting prompted the unit type if you just click the message. It picks the right unit during calculation of the action but the step you showed is before that point. So if you didnt manaully chose a unit it is just not filled yet. But gets filled during execution, a step later. And gets also the right one for calculation. Looking into why I cant use dynamic content (searching for the right unit in settings before execution) in the app intents. So will comeback to this today.

**Comment by bjornoleh** [comment]
I am getting these uncommitted changes upon build with Xcode. Probably Xcode which disagrees with something here. Should this be committed?

```
diff --git a/FreeAPS/Sources/Localizations/ShortcutsDetail/ShortcutsDetail.xcstrings b/FreeAPS/Sources/Localizations/ShortcutsDetail/ShortcutsDetail.xcstrings
index 161c608..ed9bf3e 100644
--- a/FreeAPS/Sources/Localizations/ShortcutsDetail/ShortcutsDetail.xcstrings
+++ b/FreeAPS/Sources/Localizations/ShortcutsDetail/ShortcutsDetail.xcstrings
@@ -367,9 +367,6 @@
         }
       }
     },
-    "Activating ${preset}" : {
-
-    },
     "Add Carbs" : {
       "localizations" : {
         "de" : {
@@ -410,9 +407,6 @@
         }
       }
     },
-    "Adding ${carbQuantity} g at ${dateAdded}" : {
-
-    },
     "Allow to add carbs" : {
       "localizations" : {
         "de" : {
@@ -693,9 +687,6 @@
         }
       }
     },
-    "Applying ${bolusQuantity} U " : {
-
-    },
     "Applying ${carbQuantity} g at ${dateAdded}" : {
       "extractionState" : "manual",
       "localizations" : {
@@ -778,9 +769,6 @@
         }
       }
     },
-    "Applying ${preset} override" : {
-
-    },
     "Are you sure to add %@ g of carbs ?" : {
       "localizations" : {
         "de" : {
@@ -1147,15 +1135,6 @@
         }
       }
     },
-    "Choose the Carb preset ${preset}" : {
-
-    },
-    "Choose the temporary override  ${preset}" : {
-
-    },
-    "Choose the TempTarget  ${preset}" : {
-
-    },
     "COB" : {
       "localizations" : {
         "de" : {
@@ -1883,15 +1862,6 @@
         }
       }
     },
-    "Immediately activating ${preset}" : {
-
-    },
-    "Immediately adding ${carbQuantity} g at ${dateAdded}" : {
-
-    },
-    "Immediately applying ${bolusQuantity} U" : {
-
-    },
     "Immediately applying ${carbQuantity} g at ${dateAdded}" : {
       "extractionState" : "manual",
       "localizations" : {
@@ -1974,12 +1944,6 @@
         }
       }
     },
-    "Immediately applying ${preset} override" : {
-
-    },
-    "Immediatly starting ${target} ${unit} TempTarget for ${duration} min" : {
-
-    },
     "IOB" : {
       "localizations" : {
         "de" : {
@@ -3359,9 +3323,6 @@
         }
       }
     },
-    "Starting ${target} ${unit} TempTarget for ${duration} min" : {
-
-    },
     "Starting Date" : {
       "extractionState" : "manual",
       "localizations" : {
```

**Comment by avouspierre** [comment]
the last commit changes some translation sentences and remove others. I'll push a new commit.

**Comment by mountrcg** [comment]
> If an override with no name is set, the message “No active Override to cancel” is displayed. But the override is correctly cancelled in Trio

@Sjoerd-Bo3 

I observe, that the cancellation of the override is some shifted by 1 event.

- apply an override preset with name "xyz"
- use cancel shortcut
   - message no active override to cancel
   - but the cancellation will happen
- apply another override without name
- use cancel shortcut
   - message Override "xyz" canceled <- btw typo, should read cancelled
   - but the cancellation will happen

I somehow cannot reproduce consistently the above behaviour.

Hitting Cancel Override twice in succession will lead to:
Bildschirmfoto 2024-06-18 um 14 20 43
Hitting a third and fourth time will not trigger the error!?!

**Comment by mountrcg** [comment]
So I re-tested all of the above and could not reproduce any of the errors. Could it have been some initial thing as that were the first overrides I activated on a fresh installation?

**Comment by bjornoleh** [comment]
> So I re-tested all of the above and could not reproduce any of the errors. Could it have been some initial thing as that were the first overrides I activated on a fresh installation?

Ok, thanks. Can you verify there are no uncommitted changes on build now after the last commit? If so I can give my approval.

**Comment by bjornoleh** [comment]
> > However it is strange that the mmol value is correctly converted to mgdL when saved in Trio (target values are always stored as mgdL), but units are not recognised for the message?
> 
> I dove in to it yesterday and not sure how to fix but it is releated to you not getting prompted the unit type if you just click the message. It picks the right unit during calculation of the action but the step you showed is before that point. So if you didnt manaully chose a unit it is just not filled yet. But gets filled during execution, a step later. And gets also the right one for calculation. Looking into why I cant use dynamic content (searching for the right unit in settings before execution) in the app intents. So will comeback to this today.

Ref my previous post: Unless there are still tings to do regarding the above.

**Comment by MikePlante1** [comment]
I merged this PR into the [latest alpha](https://github.com/nightscout/Trio/commit/3ec6176db95109463300f5c322b1acdce4286d0c) and tested on a simulator... 

I toggled on `Allows to bolus with shortcuts` and left `Method to limit the bolus amount` set to `Limit by max bolus`. I then opened the Shortcuts app and executed `Trio Bolus` and entered `55`. A popup appeared to confirm I wanted to bolus 55U and I hit `Continue` to which it alerted `A bolus command of 10U of insulin was sent.`

In this circumstance, I think the bolus shortcut should fail and _no_ insulin should be delivered, not the maximum allowed amount.

Also, here's a couple wording suggestions:

current:
> Allows to bolus with shortcuts

suggestion:
> Allow bolusing with shortcuts

current:
> The application lets you create automations using shortcuts. Go to the Shortcuts application to create new automations.

suggestion:
I think this can just be removed entirely. Shortcuts has been in iOS for almost 6 years now, so I don't think it needs an explanation.

I'm a little unsure the best way to phrase the following, but how about

current: 
> Method to limit the bolus amount

suggestion:
> Limit bolus from shortcuts to

current:
> Limit by max bolus

suggestion:
> Max Bolus

current:
> Limit by insulin suggestion estimation

suggestion:
> Insulin recommended

**Comment by aug0211** [comment]
Are there any blockers for this to go forward? Was testing successful?

I see some phrasing recommendations - I recommend those either get fixed (@mike maybe add them in directly?) or touch those up after initial release?

**Comment by MikePlante1** [comment]
@aug0211 Phrasing suggestions aren’t crucial, but I think the initial part of comment about delivering the max when requesting above the max should be addressed.

File: FreeAPS/Sources/Shortcuts/Bolus/BolusIntentRequest.swift [hunk 17]
```diff
@@ -0,0 +1,60 @@
...
+    func bolus(_ bolusAmount: Double) throws -> LocalizedStringResource {
+        var bolusQuantity: Decimal = 0
+        switch settingsManager.settings.bolusShortcut {
+        case .notAllowed:
+            return LocalizedStringResource(
+                "the bolus is not allowed with shortcuts",
+                table: "ShortcutsDetail"
+            )
+        case .limitBolusMax:
+            bolusQuantity = apsManager
```
aug0211, line 19: @MikePlante1 is this where the change is needed for your recommendation? Is the check here to compare the bolusAmount with maxBolus?

I think it's something like this:

```
if bolusAmount > maxBolus {
   return LocalizedStringResource("Maximum bolus exceeded, no bolus administered.")
}
else {
 bolusQuantity = roundBolus(Decimal(bolusAmount))
}
```

Do this down below for the other version - checking for recommended percentage.

aug0211 (commented):
Added a note for @MikePlante1, I think I identified the section of code he is interested in, and I proposed what this could look like.

MikePlante1:
`Trio Carbs` shortcut doesn't seem to work as expected. Testing in a simulator with this PR merged into the latest dev.

| | Input | Response |
| --- | --- | --- |
|  | 1 | `Are you sure to add 1 g of carbs ?` |
|  | 30 | `Are you sure to add 30 g of carbs ?` |
|  | 59 | `Are you sure to add 59 g of carbs ?` |
| ❌ | 60 | `Are you sure to add 1:00 g of carbs ?` |
| ❌ | 100 | `Are you sure to add 1:40 g of carbs ?` |
| ❌ | 200 | `Are you sure to add 3:20 g of carbs ?` |
| ❌ | 201 | `What is the quantity of the carb to add ?` |

MikePlante1:
Also, Carbs seems to work like Bolus when it comes to exceeding the maximum limit. As in, if I set Max Carbs to 30g and enter 44g in the Carbs shortcut, it prompts to confirm an entry of "30g". I think a better workflow would be for it to just alert you that your entry exceeded the maximum and prompt you to edit your entry. Just like how the Meal and Bolus modules work in the Trio app on the iPhone.

MikePlante1:
Another test:
- Created and saved two profile overrides: 110% and 120%
- Confirmed no override is currently being used in Trio
- Activated 110% override from shortcut
- Confirmed 110% override is being used in Trio
- Activated 120% override from shortcut
- Confirmed 120% override is being used in Trio
- Canceled override from shortcut
- It should have prompted me to cancel the 120% override but instead it prompted to cancel the 110% override
- Tapped to confirm the cancelation in the shortcut anyway
- Confirmed no override is being used in Trio
- Tapped the Cancel Override shortcut again
- It prompted me to cancel the 110% override, even though no override was being used
- Tapped to confirm the cancelation in the shortcut anyway
- Tried to tap Cancel Override again, but this time it correctly informed me I did not have any running overrides.

https://github.com/nightscout/Trio/assets/82073483/9bcfff09-b8d2-4dd0-9fc2-57b55cf75f2f

aug0211:
> `Trio Carbs` shortcut doesn't seem to work as expected. Testing in a simulator with this PR merged into the latest dev.
> 
> 
> 
> | | Input | Response |
> 
> | --- | --- | --- |
> 
> |  | 1 | `Are you sure to add 1 g of carbs ?` |
> 
> |  | 30 | `Are you sure to add 30 g of carbs ?` |
> 
> |  | 59 | `Are you sure to add 59 g of carbs ?` |
> 
> | ❌ | 60 | `Are you sure to add 1:00 g of carbs ?` |
> 
> | ❌ | 100 | `Are you sure to add 1:40 g of carbs ?` |
> 
> | ❌ | 200 | `Are you sure to add 3:20 g of carbs ?` |
> 
> | ❌ | 201 | `What is the quantity of the carb to add ?` |

I can reproduce this when running locally on a test phone, in Trio, iAPS 4.X and iAPS 3.X.

They all end up having the actual expected action (they enter the amount of carbs entered), except for the 201 case which never logs anything because it keeps asking for an amount.

So this issue is a "display/confirmation" issue, living somewhere in the confirmation code. Will try to take a look next time I can get time.

aug0211:
> Another test:
> 
> - Created and saved two profile overrides: 110% and 120%
> 
> - Confirmed no override is currently being used in Trio
> 
> - Activated 110% override from shortcut
> 
> - Confirmed 110% override is being used in Trio
> 
> - Activated 120% override from shortcut
> 
> - Confirmed 120% override is being used in Trio
> 
> - Canceled override from shortcut
> 
> - It should have prompted me to cancel the 120% override but instead it prompted to cancel the 110% override
> 
> - Tapped to confirm the cancelation in the shortcut anyway
> 
> - Confirmed no override is being used in Trio
> 
> - Tapped the Cancel Override shortcut again
> 
> - It prompted me to cancel the 110% override, even though no override was being used
> 
> - Tapped to confirm the cancelation in the shortcut anyway
> 
> - Tried to tap Cancel Override again, but this time it correctly informed me I did not have any running overrides.
> 
> 
> 
> https://github.com/nightscout/Trio/assets/82073483/9bcfff09-b8d2-4dd0-9fc2-57b55cf75f2f
> 
> 
> 
> 

Couldn't get video to load here but this seems similar to the carb issue - a shortcut confirmation text issue.

I think:

1. This is an annoying bug, should not stop merging

2. A quick fix is to just remove the reference in the confirmation text to a specific override and just say "are you sure you'd like to cancel the active override?" or something like this - which IMO is what this shouldn't been anyway, making this fix both simpler to implement and simpler to grok when using.

I can work on and test this in an in vitro device.

aug0211:
> > Another test:
> > 
> > * Created and saved two profile overrides: 110% and 120%
> > * Confirmed no override is currently being used in Trio
> > * Activated 110% override from shortcut
> > * Confirmed 110% override is being used in Trio
> > * Activated 120% override from shortcut
> > * Confirmed 120% override is being used in Trio
> > * Canceled override from shortcut
> > * It should have prompted me to cancel the 120% override but instead it prompted to cancel the 110% override
> > * Tapped to confirm the cancelation in the shortcut anyway
> > * Confirmed no override is being used in Trio
> > * Tapped the Cancel Override shortcut again
> > * It prompted me to cancel the 110% override, even though no override was being used
> > * Tapped to confirm the cancelation in the shortcut anyway
> > * Tried to tap Cancel Override again, but this time it correctly informed me I did not have any running overrides.
> > 
> > 
> >   
> >     
> >     
> > 
> >     Screen.Recording.2024-07-01.at.3.50.17.PM.mov
> 
> Couldn't get video to load here but this seems similar to the carb issue - a shortcut confirmation text issue.
> 
> I think:
> 
> 1. This is an annoying bug, should not stop merging
> 2. A quick fix is to just remove the reference in the confirmation text to a specific override and just say "are you sure you'd like to cancel the active override?" or something like this - which IMO is what this shouldn't been anyway, making this fix both simpler to implement and simpler to grok when using.
> 
> I can work on and test this in an in vitro device.

Fixed this one and [committed](https://github.com/avouspierre/Trio/commit/f72f9cb9247bd2eb054ee95979596f2344ef4776).

I kept the originally desired intent of showing the "current" override that was cancelled in the confirmation text.

If we want to switch to the simpler "Override cancelled" confirmation I proposed above, that's a simple 1 line change we can make in the future after merging.

aug0211:
> > `Trio Carbs` shortcut doesn't seem to work as expected. Testing in a simulator with this PR merged into the latest dev.
> > | | Input | Response |
> > | --- | --- | --- |
> > |  | 1 | `Are you sure to add 1 g of carbs ?` |
> > |  | 30 | `Are you sure to add 30 g of carbs ?` |
> > |  | 59 | `Are you sure to add 59 g of carbs ?` |
> > | ❌ | 60 | `Are you sure to add 1:00 g of carbs ?` |
> > | ❌ | 100 | `Are you sure to add 1:40 g of carbs ?` |
> > | ❌ | 200 | `Are you sure to add 3:20 g of carbs ?` |
> > | ❌ | 201 | `What is the quantity of the carb to add ?` |
> 
> I can reproduce this when running locally on a test phone, in Trio, iAPS 4.X and iAPS 3.X.
> 
> They all end up having the actual expected action (they enter the amount of carbs entered), except for the 201 case which never logs anything because it keeps asking for an amount.
> 
> So this issue is a "display/confirmation" issue, living somewhere in the confirmation code. Will try to take a look next time I can get time.

Fixed this and a few other things in [this commit](https://github.com/nightscout/Trio/commit/f5915f5d7aac1b5410f43df8256c7c4b878c09d2).

1. Carb values won't give weird confirmation messages anymore as reported above
2. Max value violations will not be silently capped, they will fail with an alert
3. Users are asked for fat and protein now (previously defaulted to 0 (validation occurs as well)
4. Success confirmation lists carbs, fat, and protein clearly
5. Max value respects user settings configured in Trio
6. Future (out of scope here) - should allow user to select the time
7. Future (out of scope here) - cleaner code option if we get access to settingsManager for max values in the AppIntent

aug0211:
> @aug0211 Phrasing suggestions aren’t crucial, but I think the initial part of my comment about delivering the max when requesting above the max should be addressed.

Fixed in [this commit](https://github.com/nightscout/Trio/commit/6ce9fbeb5a78bbb2aec319d4afadf9f0ff3c7756). Boluses will not be blocked with an alert if they exceed the maximum configured (whether this is pump max bolus or insulin suggestion).

aug0211:
> I merged this PR into the [latest alpha](https://github.com/nightscout/Trio/commit/3ec6176db95109463300f5c322b1acdce4286d0c) and tested on a simulator...
> ...
> ...
> Also, here's a couple wording suggestions:
> 
> current:
> 
> > Allows to bolus with shortcuts
> 
> suggestion:
> 
> > Allow bolusing with shortcuts
> 
> current:
> 
> > The application lets you create automations using shortcuts. Go to the Shortcuts application to create new automations.
> 
> suggestion: I think this can just be removed entirely. Shortcuts has been in iOS for almost 6 years now, so I don't think it needs an explanation.
> 
> I'm a little unsure the best way to phrase the following, but how about
> 
> current:
> 
> > Method to limit the bolus amount
> 
> suggestion:
> 
> > Limit bolus from shortcuts to
> 
> current:
> 
> > Limit by max bolus
> 
> suggestion:
> 
> > Max Bolus
> 
> current:
> 
> > Limit by insulin suggestion estimation
> 
> suggestion:
> 
> > Insulin recommended

Phrasing notes requested in this comment are addressed in [this commit](https://github.com/nightscout/Trio/commit/90d62d2cc65e8142371b1068edffce06d348f9f0).

MikePlante1:
Thanks @aug0211, it's certainly improved now, but still a bit more to fix I think.

***

Regarding canceling of overrides, I think it's fine to just say `Override cancelled` like you proposed. Your commit fixed it displaying the wrong override initially but there are still problems with it. Maybe look at how TT shortcuts are implemented, because I don't run into these problems with the "Cancel Trio TempTarget" shortcut.

Example 1: Start a new profile override (not a saved override) in the Trio app. Then run the "Cancel Trio Override" shortcut. The shortcut says "No active Override to cancel", even though it actually does cancel the override.

Example 2: Start a saved profile override with a set duration. Wait for the override to expire. Run the "Cancel Trio Override" shortcut and it will say `Override 'name' canceled` even though it wasn't running.

Example 3: Create two saved profiles: OA and OB. In the Trio app, start OA and then start OB. Run the "Cancel Trio Override" shortcut and it will say `Override 'OB' canceled`. Checking the Trio app confirms there are now no profile overrides running, but if you run the "Cancel Trio Override" shortcut again it will say `Override 'OA' Canceled`

***

For the "Trio Bolus" shortcut, when `Limit bolus from shortcuts to` is set to `Max bolus`, the shortcut works properly when too big of a bolus is requested by telling the user `The bolus cannot be larger than the pump setting max bolus` and then not delivering any bolus.

Likewise, when I set `Limit bolus from shortcuts to` to `Bolus recommended`, with a max bolus setting of 4U, where Insulin Recommended was 1.3U, and I tried to bolus 3U... the Bolus shortcut told me `The bolus cannot be larger than the suggested insulin.` and did not deliver any bolus.

The check failed, however, when the Insulin recommended equals the Max Bolus. I entered 120g carbs, which caused the Insulin recommended to be 4U. It would have been higher, but it was capped at 4U because the Max Bolus setting is 4U. I then ran the "Trio Bolus" shortcut and requested a 5U bolus. It asked `Are you sure you want to bolus 5 U of insulin ?` and when I selected "Continue" it replied `A bolus command of 4 U was sent`.

***

carbs/fat/protein seem to be properly capped at their max limits in settings. I do find it odd that it seems to pad numbers to one decimal place in the prompts, though. I also don't like how it responds when you don't enter a value for any of the prompts, though. It looks like the prompt should default to 0 if you don't enter anything for carbs, fat, or protein and just hit "done" to move to the next macro prompt, but instead the shortcut just kills itself without any feedback.  So if you enter 10 for carbs, hit Done, then it prompts for fat but it looks like 0g is already pre-populated, so you hit Done, and it crashes.

Maybe there should be two shortcuts: `Trio Add Carbs` and `Trio Add Meal`, the former only prompting for carbs and the latter prompting for carbs, fat, and protein. For the meal entry, though, it should still allow the user to just press Done for any of the values instead of being forced to type 0.

PR #180, opened 2024-05-13
Update README.md: Create ConfigOverride.xcconfig by command line

bjornoleh:


marionbarker:
I recommend adding the BuildTrio.sh instructions below the line that says there is a script available. 

Copy the code embedded here into the Trio README:

https://github.com/loopandlearn/lnl-scripts/tree/main?tab=readme-ov-file#scripts-for-the-trio-app

Code inspection OK of what you added looks good but I can’t test it right now.

README.md
```diff
@@ -24,8 +24,15 @@ git clone --branch=<<branch>> --recurse-submodules https://github.com/nightscout
...
 ```
 
-modify Config.xconfig to replace ##TEAM_ID## with your Apple Developer ID.
+Create an ConfigOverride.xcconfig file that will automate signing of the build targets in Xcode:
 
+```
+echo 'DEVELOPER_TEAM = xxxxxxxxxx' > ConfigOverride.xcconfig
+```
+
+Modify the Config.xconfig to replace `xxxxxxxxxx` with your Apple Developer ID. You can do this from within Xcode, or by using some other text editor.
```
MikePlante1, line 33: How about recommending `nano ConfigOverride.xcconfig` instead of "some other text editor", since it is CLI instructions.

MikePlante1: Also, maybe recommend replacing xxxxxxxxxx with your TEAMID _before_ entering the command in Terminal.

bjornoleh: I was thinking that involving the nano editor would be too complicated, would have to instruct about saving the file (ctrl-o) and existing nano (ctrl-x). Which is possible of course. Is nano always shipped with Mac by the way? I think the default editor might be `vi` 😬

bjornoleh: But instructing to replace xxxxxx with the teamID before pasting the command into Terminal is probably the easiest and most lightweight solution. And if someone gets it wrong, it's just to repeat it, as `>` will replace the line as opposed to append it. Or one can edit in Xcode.

bjornoleh: I pushed that modification now.

MikePlante1 (dismissed):
Did you want to add the build script that Marion mentioned? If you’d prefer that left for a different PR, then LGTM.

bjornoleh:
We can fit that in here if someone has a chance to push it now, otherwise merge as is.

README.md
LiroyvH: All hail nano! I do believe it is shipped with MacOS by default, they even have a support page on it (admittedly slightly to my surprise.).

MikePlante1:
> We can fit that in here if someone has a chance to push it now, otherwise merge as is.

Thoughts on this?

CLI readme

bjornoleh:
> > We can fit that in here if someone has a chance to push it now, otherwise merge as is.
> 
> Thoughts on this?

Looks great!

marionbarker:
I reviewed the entire README file after the last commit/
Please add the missing links that are now known.

Look for TODO: 
* There are 3 of them 
* I know at least one of those is now known.

Tests:
* I tested the addition of the script and it worked.
* I tested the manual instructions and they work.
* Tests performed using a test phone running iOS 17.4.1 with Xcode 15.3

MikePlante1:
> I reviewed the entire README file after the last commit/ Please add the missing links that are now known.
> 
> Look for TODO:
> 
> * There are 3 of them
> * I know at least one of those is now known.

I added a link for Docs (and links to FAX and Oref0), but should probably wait to add the website (`diy-trio.org`) and the Crowdin (until @t1dude says it's ready for contributions)

marionbarker (dismissed):
LGTM

MikePlante1 (dismissed):

LiroyvH:
Note: when you list it like so:
```
git clone --branch=<branch> --recurse-submodules https://github.com/nightscout/Trio.git
cd Trio
```

Then if the default shell is still set to bash, it will automatically execute the first command; which will fail because of the <branch>. This will thus result in an error such as:
"bash: branch: No such file or directory"

I'd suggest writing it as follows, so that when copy/pasting to Terminal the user will always have an opportunity to change the <branch> portion before execution, it auto-terminates if the git command fails:

`git clone --branch=<branch> --recurse-submodules https://github.com/nightscout/Trio.git && cd Trio`

LiroyvH (changes requested):
Note: when you list it like so (as it is in the current proposal):
```
git clone --branch=<branch> --recurse-submodules https://github.com/nightscout/Trio.git
cd Trio
```

Then if the default shell is still set to bash, copy/pasting it to Terminal will automatically execute the first command; which will fail because of the \<branch\>. This will thus result in an error such as:
"bash: branch: No such file or directory"

I'd suggest writing it as follows, so that when copy/pasting directly to Terminal the user will - regardless of which shell they're using - always have an opportunity to change the \<branch\> portion before execution (and it auto-terminates if the git command fails for whatever reason; preventing an error for cd on such an occasion):

`git clone --branch=<branch> --recurse-submodules https://github.com/nightscout/Trio.git && cd Trio`

MikePlante1:
> `git clone --branch=<branch> --recurse-submodules https://github.com/nightscout/Trio.git && cd Trio`

Thanks for the suggestion.

2-line in zsh: success
2-line in bash: fail
1-line in zsh: success
1-line in bash: success

I'll commit that now with you as the author

MikePlante1 (approved):

becksen (approved):
reviewed

LiroyvH (approved):

marionbarker (approved):

Issue #171, opened 2024-05-10
[Feature request]-“Add pump” button when pump has not been configured 


bjornoleh:
We now have an “Add CGM” button when no CGM is configured. Perhaps we should have a similar button for to set up a pump when not configured?

![image](https://github.com/nightscout/Trio/assets/63544115/2b21ff94-7b55-4bf3-9fe2-0308fd7ef4e2)

Tagging @avouspierre

avouspierre:
What do you think with this proposal ? @bjornoleh   

![Simulator Screenshot - iPhone 15 - 2024-05-14 at 15 22 45](https://github.com/nightscout/Trio/assets/4339604/27383500-063d-4d52-8526-9fe027e691a4)

bjornoleh:
Looks pretty good. I don't have a better suggestion for a pump pictogram, but perhaps someone else will weigh in?

PR #184, opened 2024-05-14
“Add pump” button when pump has not been configured

avouspierre:
The PR adds a “pump” icon and “add pump” label in the header of the main view. 
The tap gesture open the view to select a pump. 

Link with #171 to review.
 
Simulator Screenshot - iPhone 15 - 2024-05-14 at 17 14 04

### DIFF
--- a/FreeAPS/Sources/Modules/Home/HomeStateModel.swift
+++ b/FreeAPS/Sources/Modules/Home/HomeStateModel.swift
@@ -185,6 +185,8 @@ extension Home {
                             setupDelegate: self
                         ).asAny()
                         self.router.mainSecondaryModalView.send(view)
+                    } else if show {
+                        self.router.mainSecondaryModalView.send(self.router.view(for: .pumpConfigDirect))
                     } else {
                         self.router.mainSecondaryModalView.send(nil)
                     }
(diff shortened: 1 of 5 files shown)


bjornoleh (approved):
Works as expected during brief testing by adding and removing the sim or Dash pump driver on a non.looping phone. Thanks!

I did not inspect the code.

bjornoleh:
Do you plan on resolving the remaining 2 sonarcloud issues? If so I'll wait with reviewing :-)

avouspierre:
> Do you plan on resolving the remaining 2 sonarcloud issues? If so I'll wait with reviewing :-)

That's the problem with a "minor" fix required to review again. I don't think to resolve the 2 sonarcloud issues because all the (old) code is based on the double  or triple closures in swiftUI. Not sure this rule is very useful and improve the maintenance of the code.

marionbarker:
Code review - some of this is beyond my skill level, but all looks reasonable.

## Testing:

Configuration:
* SE phone, iOS 17.5, Trio dev branch (commit 56a3f878), no pump attached (graphic on left)
* `git apply `trio-pr-184.patch` and rebuild (graphic on right)

pr-184-success

marionbarker (approved):

bjornoleh (approved):
Still works as expected by testing on the phone. Merging after two approving reviews.

Issue #136, opened 2024-04-29
Minor iAPS left-overs


marionbarker:
These are nits - just providing this list so they are available for update when and if desired.

I am testing only alpha branch (soon to be dev).
Configuration: commit 99effeb7

This is a plist file so a little different:
* Open-iAPS/FreeAPS/Resources/Info.plist
* contains `<string>com.artificial-pancreas-iaps</string>`

The rest of these are in *.swift files and indicate the line numbers where hard-coded "iAPS" exists:
* FreeAPS/Sources/Shortcuts/State/
    * ListStateIntent.swift, line 6, 12, (not sure about the `StateiAPSResults`, see next file)
    * StateIntentRequest.swift (I suspect this should be left alone - I do not know what it does)
    * ListStateView.swift (same comment)

Note: many of the rest of these should (later) be modified to use appName instead of hard coded string
   * I did not list all the places with hard-coded Open-iAPS but there are a lot
   * In the short term - simplest solution is to replace the hard-coded "iAPS" with hard-coded "Open-iAPS"
   * Next update, we can modify to using appName instead - that takes more time and effort

Hard-coded string from iAPS to Open-iAPS
* Open-iAPS/FreeAPS/Sources/Application/FreeAPSApp.swift, line 58
* Open-iAPS/FreeAPS/Sources/Modules/HealthKit/View/AppleHealthKitRootView.swift line 16 
* Open-iAPS/FreeAPS/Sources/Modules/NightscoutConfig/View/NightscoutConfigRootView.swift lines 106 and 132
* Open-iAPS/FreeAPS/Sources/Modules/NotificationsConfig/View/NotificationsConfigRootView.swift line 95
* Open-iAPS/FreeAPS/Sources/Modules/PreferencesEditor/PreferencesEditorStateModel.swift lines 44, 240, 381
* Open-iAPS/FreeAPS/Sources/Modules/PreferencesEditor/View/PreferencesEditorRootView.swift line 25
* Open-iAPS/FreeAPS/Sources/Services/WatchManager/GarminManager.swift line 160
* Open-iAPS/FreeAPS/Sources/Services/Calendar/CalendarManager.swift line 96
* Open-iAPS/FreeAPSWatch WatchKit Extension/ComplicationController.swift lines 11, 55

Hard-coded string from FAX to Open-iAPS
* Open-iAPS/FreeAPS/Sources/Modules/PreferencesEditor/PreferencesEditorStateModel.swift line 150
[...]

PR #197, opened 2024-05-15
LiveActivity.swift: widget URL string "Trio://"

bjornoleh:
Replaces widgetURL(URL(string: "freeaps-x://"))

Related to #136

### DIFF
--- a/LiveActivity/LiveActivity.swift
+++ b/LiveActivity/LiveActivity.swift
@@ -288,7 +288,7 @@ struct LiveActivity: Widget {
                     label.fontWidth(.compressed)
                 }
             }
-            .widgetURL(URL(string: "freeaps-x://"))
+            .widgetURL(URL(string: "Trio://"))
             .keylineTint(Color.purple)
             .contentMargins(.horizontal, 0, for: .minimal)
             .contentMargins(.trailing, 0, for: .compactLeading)


JeremyStorring (approved):

marionbarker (commented):
I am not sure how the  widgetURL is used. 
The modification looks reasonable from code review.
Does this mean a shortcut on the phone requires an update?

bjornoleh:
Tapping the Live activity widget was unaffected by the change (app opened before and after). But Trio:// is the correct url for whatever it's used for.

MikePlante1 (approved):

MikePlante1:
I concur with bjornoleh. Tested opening the app from both the lockscreen widget and the Dynamic Island both before and after applying this PR on a real phone and it worked as usual. Not really sure what (if anything) it affects.

Should it maybe use APP_URL_SCHEME from Config, though? Just a thought. Hardcoding to Trio seems to be fine, though.

bjornoleh:
I didn't quickly figure out how to use APP_URL_SCHEME here, so went with the hardcoded way instead. Shone more than good enough for this anyways.

MikePlante1:
Actually... I even just deleted that line completely and it still seems to work fine. 🤷 So not sure what it's even for.

MikePlante1:
Merging this with several approvals and it seems it could even be deleted.

Issue #146, opened 2024-05-04
Submodule tests are not run automatically, failure in CalibrationsTest


marionbarker:
**Describe the bug**
1. When running Tests in Xcode, expect submodule tests to run automatically
2. The CalibrationsTests in FreeAPSTest is failing and should be fixed

**To Reproduce**
1. Open Xcode
3. Build to a simulator
4. Run Tests

**Expected behavior**
1. All tests should run (including submodules.
2. All tests should pass

The 4 XCTAssert line that fail are: lines 34, 36, 38 and 42 for `FreeAPSTests/CalibrationsTests.swift`

marionbarker:
I kept meaning to test this but life got in the way.
This was fixed with PR #206 
I confirmed once again that automatic tests are working just fine; they all passed.

bjornoleh:
Marking as completed based on Marion’s test.

PR #206, opened 2024-05-17
Restore tests (replaces PR #147)

marionbarker:
There are 2 commits to this PR.
* The first one restores the tests which were changed to "skip".
* The second one should be reviewed - it modifies `testCreateMultipleCalibration` test so it will pass instead of fail

These are notes from @itsmojo:

Regarding having to comment out the 4 tests in CalibrationsTests, the issue seems to be that the expected values are only returned on the first Test run for a given iPhone simulator. After the first run, the values seem to be a consistent values of 0.9500000000000307, 15.999999999996717, 91.99999999999918 & 2 (instead of the 0.8, 37, 101 & 1 in the initial run that are being tested for). So this seems like this is a stateful test which really should be reworked to have consistent results on each run. Also, using XCTAssertTrue() for these tests is not as helpful as using XCTAssertEqual() so that you can see both values on a failure and so an optional accuracy parameter can be used to deal with non exact floating point values like 15.999999999996717.

Note:

The original PR #147 was automatically closed when the Trio repository was made public. This replaces that PR.

### DIFF
--- a/FreeAPS.xcodeproj/xcshareddata/xcschemes/Trio.xcscheme
+++ b/FreeAPS.xcodeproj/xcshareddata/xcschemes/Trio.xcscheme
@@ -263,7 +263,7 @@
             </BuildableReference>
          </TestableReference>
          <TestableReference
-            skipped = "YES">
+            skipped = "NO">
             <BuildableReference
                BuildableIdentifier = "primary"
                BlueprintIdentifier = "43CABDFC1C3506F100005705"
@@ -273,7 +273,7 @@
             </BuildableReference>
          </TestableReference>
          <TestableReference
-            skipped = "YES">
+            skipped = "NO">
             <BuildableReference
                BuildableIdentifier = "primary"
                BlueprintIdentifier = "C17F50CD291EAC3800555EB5"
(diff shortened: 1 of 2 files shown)


avouspierre:
the testCreateMultipleCalibration fails if you run only this one. testCreateSimpleCalibration and testCreateMulitpleCalibration use the same data so the result of calibration could be different in regard of the order of the execution. Best to launch `calibrationService.removeAllCalibrations()` at the end of each test to have a clear calibration datasource.

marionbarker:
Thanks @avouspierre. I believe the last commit makes this a "real" fix.
Ready for review.

itsmojo (approved):

itsmojo (approved):

bjornoleh (commented):
I’m not quite qualified to review this, I’m afraid.

avouspierre (approved):

MikePlante1 (approved):
I'm submitting this review as a proxy for @itsmojo since his review should count for this PR. I did do a test test, and a test build to a simulator.

marionbarker:
This closes Issue #146

Issue #198, opened 2024-05-16
Toggle option to Disable Fetching of Nightscout Carbs in Trio


Sjoerd-Bo3:
**Is your feature request related to a problem? Please describe.**
Currently, even if remote commands are disabled in Trio, carbs saved to Nightscout by anyone with Careportal access are still pulled into Trio and used in dosing decisions. This creates a security vulnerability, as it is not possible to fully disable remote control over carb data input. This issue was highlighted by Dylan Sutton, who noted that simply disabling remote commands does not prevent carb data from being fetched and used, making the system potentially unsafe.

**Describe the solution you'd like**
Introduce a toggle option in Trio settings to enable or disable the fetching of carb data from Nightscout. This setting should default to not allow fetching treatments to ensure better security. This would involve adding a basic toggle and the necessary logic to respect this setting, ensuring that when disabled, Trio does not fetch carb data from Nightscout.

**Describe alternatives you've considered**
- Disabling the Careportal in Nightscout as a security measure. However, this does not fully solve the problem as treatments can still be added through the API.
- Commenting out the Nightscout fetch code entirely in a custom build, but this approach is not user-friendly and lacks flexibility.

**Additional context**
This issue was discussed in detail, and it was noted that the most dangerous remote command is the announcement for bolusing, which can be disabled with the remote control toggle. However, to completely block unwanted entries and ensure higher security, a specific setting to disable fetching treatments from Nightscout is necessary. Daniel provided a temporary solution by commenting out the fetch code, and Sjoerd offered to implement the toggle feature, seeking guidance from Daniel's commit.

---

**Note:** This request aligns with the need to enhance the security of Trio by giving users more control over the data inputs used in dosing decisions, ultimately making the system safer and more reliable.

Sjoerd-Bo3:
*Comment from Daniel:*
I added an "allow NS download toggle" to my personal iAPS build a couple of weeks ago. Just made a commit to a fresh Trio dev testing branch with the same changes. Feel free to use and improve this as you like (Super simple code implementation, but it works)
https://github.com/dsnallfot/Trio/commit/b8d9a3160a31a417561c602e9e38915777853f35

dsnallfot:
I will make a PR based on the commit posted

PR #221, opened 2024-05-20
Add toggle to allow downloads from Nightscout (And improve NS Config UI)

dsnallfot:
Linked to issue #198 

Adds a **Download** toggle (Default setting false) that needs to be flipped on to allow any Nightscout download of Treatments (Carbs, Temp Targets) and Announcements (Remote commands for pump, basal, overrides and bolus).

When the toggle is disabled the functions fetchCarbs, fetchTempTargets and fetchAnnouncements is blocked from running (the same way that the toggle for **Upload** blocks upload and deletion of Treatments, preferences and settings.

Header renamed from "Allow Uploads" to "Allow up- and downloads"

IMG_3761

This PR do **not** change the current **Allow remote control of Trio** toggles since there could be room for discussion if disabling bolus commands maybe should have its own toggle (ie that you want to block Remote bolus announcements but still want to allow fetching carbs and temp targets).

### DIFF
--- a/FreeAPS/Sources/Modules/NightscoutConfig/View/NightscoutConfigRootView.swift
+++ b/FreeAPS/Sources/Modules/NightscoutConfig/View/NightscoutConfigRootView.swift
@@ -24,12 +24,34 @@ extension NightscoutConfig {
                 NavigationLink("Fetch and Remote Control", destination: NightscoutFetchView(state: state))
 
                 Section(
-                    header: Text("Import Settings"),
-                    footer: Text(
-                        "Importing settings from Nightscout will overwrite these settings in Trio Settings -> Configuration:\n • DIA (Pump settings)\n • Basal Profile\n • Insulin Sensitivities\n • Carb Ratios\n • Target Glucose"
-                    )
+                    header: Text("Import Settings from Nightscout"),
+                    footer: VStack(alignment: .leading, spacing: 2) {
+                        Text(
+                            "Importing settings from Nightscout will overwrite these settings in Trio Settings -> Configuration:"
+                        )
+                        HStack {
+                            Text("•")
+                            Text("DIA (Pump settings)")
+                        }
+                        HStack {
+                            Text("•")
+                            Text("Basal Profile")
+                        }
+                        HStack {
+                            Text("•")
+                            Text("Insulin Sensitivities")
+                        }
+                        HStack {
+                            Text("•")
+                            Text("Carb Ratios")
+                        }
+                        HStack {
+                            Text("•")
+                            Text("Target Glucose")
+                        }
+                    }
                 ) {
-                    Button("Import settings from Nightscout") {
+                    Button("Import settings") {
                         importAlert = Alert(
                             title: Text("Import settings?"),
                             message: Text(
--- a/FreeAPS/Sources/Modules/NightscoutConfig/View/NightscoutConfigRootView.swift
+++ b/FreeAPS/Sources/Modules/NightscoutConfig/View/NightscoutConfigRootView.swift
@@ -29,26 +29,11 @@ extension NightscoutConfig {
                         Text(
                             "Importing settings from Nightscout will overwrite these settings in Trio Settings -> Configuration:"
                         )
-                        HStack {
-                            Text("•")
-                            Text("DIA (Pump settings)")
-                        }
-                        HStack {
-                            Text("•")
-                            Text("Basal Profile")
-                        }
-                        HStack {
-                            Text("•")
-                            Text("Insulin Sensitivities")
-                        }
-                        HStack {
-                            Text("•")
-                            Text("Carb Ratios")
-                        }
-                        HStack {
-                            Text("•")
-                            Text("Target Glucose")
-                        }
+                        Text(" • ") + Text("DIA (Pump settings)")
+                        Text(" • ") + Text("Basal Profile")
+                        Text(" • ") + Text("Insulin Sensitivities")
+                        Text(" • ") + Text("Carb Ratios")
+                        Text(" • ") + Text("Target Glucose")
                     }
                 ) {
                     Button("Import settings") {
--- a/FreeAPS/Sources/Modules/NightscoutConfig/View/NightscoutConfigRootView.swift
+++ b/FreeAPS/Sources/Modules/NightscoutConfig/View/NightscoutConfigRootView.swift
@@ -67,8 +67,9 @@ extension NightscoutConfig {
                     if state.isUploadEnabled {
                         Toggle("Glucose", isOn: $state.uploadGlucose).disabled(!state.changeUploadGlucose)
                     }
+                    Toggle("Download", isOn: $state.isDownloadEnabled)
                 } header: {
-                    Text("Allow Uploads")
+                    Text("Allow Up- and downloads")
                 }
 
                 Section {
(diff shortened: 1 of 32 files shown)


dsnallfot:
@Sjoerd-Bo3 @aug0211 @MikePlante1 
I cannot add reviewers for the PR, so I just tag you here.

This is just a PR for the download treatments toggle. No refactoring of the "allow remote bolus" toggles that was also discussed in discord. That could be of course be added, but I realized that its probably good to discuss if we want to keep the possibility to allow carbs and temp targets downloads, but block remote commands for blousing for instance. Then we need to keep the allow remote control-toggle (but maybe rename it somehow)

aug0211:
@dsnallfot what happens if we enable downloads but disable the other  remote treatments toggle? I haven’t fully grokked how these work together yet.

dsnallfot:
@aug0211 if the download toggle is off, no treatments or announcements gets fetched.

If the download toggle is on, treatments get fetched. But announcements get fetched only if the remote control toggle also is on. 

To make this better we could disable and hide the "remote control" toggle when download is off? What do you think?

dsnallfot:
Converted to draft until decided if I should also hide the remote control toggle/section when download is off.

dsnallfot:
Demo after latest commit 6e187c1

https://github.com/nightscout/Trio/assets/72826201/7390963e-34ed-420d-af98-3d9b0df166ce

FreeAPS/Sources/Modules/NightscoutConfig/View/NightscoutFetchView.swift
```diff
@@ -0,0 +1,35 @@
...
+                    "The Fetch Treatments toggle enables fetching of carbs and temp targets entered in Careportal or by another uploading device than Trio."
+                )
+            }
+            Section {
+                Toggle("Remote Control", isOn: $state.allowAnnouncements)
+                    .disabled(!state.isDownloadEnabled)
+            } header: { Text("Allow Remote control of Trio")
+            } footer: {
+                Text(
+                    "Fetch Treatments needs to be allowed to be able to toggle on Remote Control.\n\nWhen enabled you allow these remote functions through announcements from Nightscout:\n • Suspend/Resume Pump\n • Opening/Closing Loop\n • Set Temp Basal\n • Enact Bolus."
```
dnzxy, line 29: I‘d suggest splitting these multiple texts into several text elements. Will only comment this for this LOC, applicable for the other info texts as well.

dnzxy (commented):
Looks good to me overall. Great work. Maybe add the new `isDownloadEnabled` settings to the defaults file as well? Small comment regarding multiline text.

FreeAPS/Sources/Modules/NightscoutConfig/View/NightscoutFetchView.swift
dsnallfot: Good point! You mean something like this?:
```swift
footer: VStack(alignment: .leading, spacing: 2) {
                        Text(
                            "Importing settings from Nightscout will overwrite these settings in Trio Settings -> Configuration:"
                        )
                        Text("• DIA (Pump settings)")
                        Text("• Basal Profile")
                        Text("• Insulin Sensitivities")
                        Text("• Carb Ratios")
                        Text("• Target Glucose")
                    }
```

dsnallfot:
> Looks good to me overall. Great work. Maybe add the new `isDownloadEnabled` settings to the defaults file as well? Small comment regarding multiline text.

Ok. I thought I had added it everywhere? (Used the existing `isDownloUploadEnabled` as guidance. Can you please point me to where I find the default file?

FreeAPS/Sources/Modules/NightscoutConfig/View/NightscoutFetchView.swift
dnzxy: Yes exactly what I meant. You might also be able to use string literals here so that you can make use of the existing translations for all those setting (section) names 🤔

dnzxy:
I meant this file https://github.com/dsnallfot/Trio/blob/dev-test-stuff/FreeAPS/Resources/json/defaults/freeaps/freeaps_settings.json but just realized I missed that you had added it already. Please disregard that comment 😅

FreeAPS/Sources/Modules/NightscoutConfig/View/NightscoutFetchView.swift
dsnallfot: Will that work with my pretty bullets • ? Hm... I guess I could make HStacks for each line and put the bullet in separat text elements (since the bullet don't needs to be translated)

dnzxy: You should be able to either do something like 

`Text("• " + NSLocalizedString("…"))` or you can also concatenate multiple texts into one by doing 

`Text("…") + Text("…")`

Either one should definitely work with your lovely bullet points 😊

dsnallfot: Great! Also realized that I don't need to use the NSLocalizedString("xxx", comment: yyy) stuff to make translations work for text within footers/headers. However. Not all text have existing translations as is (And should of course be added, but I really need help from someone who knows how to efficiently update alla localization files without having to copy+paste text 100s of times)

Will make all text split changes and push a new commit in a little while.

dnzxy: My understanding is that you need to use it when you are using string literals (so stuff like `"\(myvar)"` and when concatenating strings within a single `Text("…")`. The way you did it now should be fine without the NSLocalized 👍😊

FreeAPS/Sources/Modules/NightscoutConfig/View/NightscoutConfigRootView.swift
```diff
@@ -94,50 +58,38 @@ extension NightscoutConfig {
...
-                        DecimalTextField("", value: $state.localPort, formatter: portFormater)
-                    }
-                } header: { Text("Local glucose source") }
+                        .alert(isPresented: $importedHasRun) {
+                            Alert(
+                                title: Text((fetchedErrors.first?.error ?? "").count < 4 ? "Settings imported" : "Import Error"),
+                                message: Text(
+                                    (fetchedErrors.first?.error ?? "").count < 4 ?
+                                        NSLocalizedString(
+                                            "\nNow please verify all of your new settings thoroughly: \n\n • DIA (Pump settings)\n • Basal Profile\n • Insulin Sensitivities\n • Carb Ratios\n • Target Glucose\n\n in Trio Settings -> Configuration.\n\nBad or invalid profile settings could have disastrous effects.",
```
dnzxy, line 67: Same thing with splitting the texts here as well, please. Rest looks splendid 😊

dsnallfot: That text block is already localized as it is (look in the localizations files), that's why I didn't give it the split treatment. Sure we want to change that??

dnzxy: You‘re right. Dang. I guess let’s leave it for now? Not sure, as we need to do a lot of translations anyway and we *could* adjust those translations by hand as the \n is a good delimiter. Up to you, I‘m okay with either 👍

dsnallfot: Nah. Let's leave it for now, need to go bug hunting elsewhere 😊 . Thanks you for a good review and feedback.

dnzxy (approved):

aug0211 (approved):
Approved!

I was a bit thrown off by how many localization updates were in here, but they are

1. trivial in nature (capitalization)
2. consistent/intentional
3. related to the NS options focused on across this PR

Let's go! 🚢

PR #235, opened 2024-05-24
Enable editing of preset profile overrides

dsnallfot:
- Enable edit of all settings and/or name for already stored profile overrides
- Swipe left to edit or delete
- Get list of all settings included in the override when saving a new one or editing an existing one

Suggested solution for issue #234 

- Extra useful when overrides are used with shortcuts (no need to re add changed overrides in shortcuts since id remains unchanged after edit - even if you change name on the override)

**NOTE!** This is not an attempt to redesign the entire UI. This is just to add edit functionality with minimal changes.

**Screenshots**
| Edit or Delete swipe | Settings summary before saving |
|--------|--------|
| Simulator Screenshot - iPhone 13 mini+ - 2024-05-25 at 01 55 40 | Simulator Screenshot - iPhone 13 mini+ - 2024-05-25 at 01 56 39 |

Issue #234, opened 2024-05-24
Allow Editing of Existing Profile Overrides


aug0211:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I need to make a modification to an existing profile override, but not enough to justify creating an all new override. It is toilsome to need to delete the existing override and re-create a new override from scratch that matches it in all aspects except for the minor tweak I'd like to make.

For example, I have an override to tamp down sizes of SMBs in the early morning hours, from roughly 3 AM to 7 AM. As we navigate life, I sometimes need to make tweaks such as (1) how many minutes the overrides lasts for by default or (2) how many minutes of basal I want to allow to be applied via SMB or perhaps (3) try a modified target for this time period or even (4) adjust the % slider during this period.

To do this today, I must delete the existing profile override and create a new one, with everything the same except for the one thing I'd like to tweak and save as part of the profile override going forward.

This is especially toilsome when working with Shortcut Automations, as I often find that I need to enter the Automation and re-select the Profile Override from the automation to make sure it triggers (even if re-created with an identical name).

Of course, this all applies to Temp Targets as well, but I would prioritize this for Profile Overrides ahead of temp targets.

**Describe the solution you'd like**
An "Edit" option for existing Profile Overrides. This would allow us to change any settings in the Profile Override, and click save to persist the changes.

**Describe alternatives you've considered**
The best alternative I've discovered so far is to completely delete the existing Profile Override and re-create a new one from scratch, matching the old Profile Override in all ways except for whichever small tweak I'd like to make. I then go into any Shortcut Automations referencing the Profile Override and re-select the same Profile Override from the Shortcut Automation dropdown list to be sure it works.

For Temp Targets, the workaround I've identified is to edit the config through settings, where we can get access to the JSON. This is not ideal for most users but is fine for me. I don't think this is an option for Profile Overrides due to the storage mechanism.

One other alternative which would be the best could be that this is already doable today, and I just don't know how to do it :)

Issue #154, opened 2024-05-06
Remove the experimental in Temp target view


avouspierre:
**Is your feature request related to a problem? Please describe.**
The temp target view allows to activate a experimental slider to "simulate" a target in regard of the % basal insulin (HBT) . Now, the implementation is done with Override profil and not useful to stay here
![CleanShot 2024-05-06 at 13 41 56](https://github.com/nightscout/Open-iAPS/assets/4339604/8da14b41-2e14-42a6-9617-c03a5800bc0e)

**Describe the solution you'd like**
remove this part in the view and all code associated in HomeView and others...

**Describe alternatives you've considered**
no object

**Additional context**

mountrcg:
There is not just Loop users migrating to Trio, but also AAPS users who are familiar with TT and exercise mode. This gives them a great usability and flexibility for TempTargets.

bjornoleh:
I don't think it was proposed to remove temp targets as such, only the slider. But there might be a case for keeping it still? Your implementation seems to take target and percent insulin as input, and adjust halfBasal target?

avouspierre:
I don't want to remove TT but remove the slider and the logic behind. I don't understand the difference between the experimental slider and the override function. For me, it is a duplicate function and override allows to do more options. I know @mountrcg that you don't implement override profil but Trio has the functionality.

mountrcg:
Yes I am aware that it is not about removing TTs. But adjusting HBT is what makes TTs flexible as it adjusts the sensitivity change impact for any given TT. So the slider lets you do that quite easily. Its valuable for exercise at different intensities. TT only adjusts sensitivity. Overrides adjust a whole lot of things including CR I think.

mountrcg:
> Your implementation seems to take target and percent insulin as input, and adjust halfBasal target?

yes Jons slider adjusted HBT and the target and calculated the sena ratio. That results in unnatural boundaries for HBT as you need that for the slider.
The slider in autoISF calculates hbt with the input of desired target ans sens ratio. Both have easily defined boundaries.
What is left to do, is that the calc result replaces the current hbt setting in profile, while at the moment it injects the result into oref overriding the hbt that comes from the profile - not good.

bjornoleh:
The CR adjustment is optional in profiles. In addition, SMB can be disabled permanently or on schedule for the duration of the profile/override.

avouspierre:
The difference between temp target with experimental and override is two different algos : 
[...]

PR #236, opened 2024-05-25
Enable editing of preset temp targets

dsnallfot:
- Enable editing Name, Target and Duration for preset temp targets
- Swipe left to edit or delete
- Replaced display of "Low target - High target" in presetView since only low target is used
- Edit popup shows current settings, and editing of name, target and duration is done in the popup

Connected to issue #234 and PR #235 

Extra useful when preset temp targets are used with shortcuts (no need to re add changed presets in shortcuts since id remains unchanged after edit - even if you change name on the preset)

**NOTE!** This is not an attempt to redesign the entire UI. This is just to add edit functionality with minimal changes.

**Screenshots**
| Edit or Delete swipe | Edit View |
|--------|--------|
| Simulator Screenshot - iPhone 13 mini+ - 2024-05-25 at 09 46 13 | Simulator Screenshot - iPhone 13 mini+ - 2024-05-25 at 09 46 19 | 

**Video demo**

https://github.com/nightscout/Trio/assets/72826201/72ac5941-af02-4232-8e6f-e49f366b54ca

### DIFF
--- a/FreeAPS/Sources/Modules/AddTempTarget/AddTempTargetStateModel.swift
+++ b/FreeAPS/Sources/Modules/AddTempTarget/AddTempTargetStateModel.swift
@@ -3,7 +3,7 @@ import SwiftUI
 
 extension AddTempTarget {
     final class StateModel: BaseStateModel<Provider> {
-        @Injected() var storage: TempTargetsStorage!
+        @Injected() private var storage: TempTargetsStorage!
         @Injected() var apsManager: APSManager!
 
         let coredataContext = CoreDataStack.shared.persistentContainer.viewContext
@@ -93,6 +93,32 @@ extension AddTempTarget {
             }
         }
 
+        private func convertAndRound(_ value: Decimal) -> Decimal {
+            if units == .mmolL {
+                return Decimal(round(Double(value.asMgdL)))
+            } else {
+                return Decimal(round(Double(value)))
+            }
+        }
+
+        func updatePreset(_ preset: TempTarget, low: Decimal) {
+            let roundedLow = convertAndRound(low)
+
+            if let index = presets.firstIndex(where: { $0.id == preset.id }) {
+                presets[index] = TempTarget(
+                    id: preset.id,
+                    name: newPresetName.isEmpty ? preset.name : newPresetName,
+                    createdAt: preset.createdAt,
+                    targetTop: roundedLow,
+                    targetBottom: roundedLow,
+                    duration: duration,
+                    enteredBy: preset.enteredBy,
+                    reason: newPresetName.isEmpty ? preset.reason : newPresetName
+                )
+                storage.storePresets(presets)
+            }
+        }
+
         func save() {
             guard duration > 0 else {
                 return
@@ -103,18 +129,15 @@ extension AddTempTarget {
                 lowTarget = Decimal(round(Double(computeTarget())))
                 saveSettings = true
             }
-            var highTarget = lowTarget
 
-            if units == .mmolL, !viewPercantage {
-                lowTarget = Decimal(round(Double(lowTarget.asMgdL)))
-                highTarget = lowTarget
-            }
+            let roundedLow = convertAndRound(lowTarget)
 
             let entry = TempTarget(
                 name: newPresetName.isEmpty ? TempTarget.custom : newPresetName,
                 createdAt: Date(),
-                targetTop: highTarget,
-                targetBottom: lowTarget,
+                targetTop: roundedLow,
+                targetBottom: roundedLow,
+
                 duration: duration,
                 enteredBy: TempTarget.manual,
                 reason: newPresetName.isEmpty ? TempTarget.custom : newPresetName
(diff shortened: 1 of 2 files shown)


dsnallfot:
Todo shortly: just figured out that md/dl values are saved as decimals instead of integers, which causes unwanted decimals in temptargets_presets.json file when mmol is converted to mg/dl and not rounded.

dnzxy:
Please see my comment on #235, the same is applies for the UI of this proposed change.

dsnallfot:
Closed since we should wait with all UI/UX changes until after 1.0 release.

aug0211:
I like this UX, it’s more intuitive than the override version (editing in the popup is more natural to other experiences on iOS)

I think swiping is clear enough, better than a long tap! I guess a dedicated “Edit” mode could be added in the top corner; maybe that’s more iOS standard? I like the swipe for options though, personally. I could see additional swipe functionality being added eventually, too, such as “duplicate” to save into a new entry with identical settings as the starting point to modify.

I’m very eager to get people who are ready and able to do UX and UI work unblocked - UI/UX is perhaps the weakest point of Trio currently (maybe behind Caregiver experience) 🙌

dsnallfot:
Reopen and put in the "idea box" until after 1.0 release

dsnallfot:
The mg/dl decimal rounding bug should be fixed now

avouspierre:
The edit mode doesn't allow to use the experimental approach. Probably not useful as asked in #154

dsnallfot:
> The edit mode doesn't allow to use the experimental approach. Probably not useful as asked in #154 

Thanks Pierre, Good spotted! You can edit temp targets set by the slider, but only target and duration (no slider in edit mode. That was a compromise I choose to make but didn't communicate clearly in the notes)

But I also noticed when testing right now this that I need to fix the mgdl-mmol conversion for targets set by slider to be able to edit them correctly. Right now only mgdl works as intended. Will fix tonite regardless of what happens with PR 154

dnzxy:
There’s quite a lot of discussion in #154 and I don’t think the slider should be removed but rather fixed (as proposed by @mountrcg ) . In going with that, the slider would need to make a reappearance here for the TT edit view.

mountrcg:
@dsnallfot can I make a proposal for the slider, which will not be based on changing HBT with slider as currently, but changing the ratio / insulin percentage as also depicted on the slider.
refer to: https://github.com/mountrcg/Trio/commit/7974cc252ca0d66070f7403410dd3f2ba92dc11f

mountrcg:
What would be great to save HBT in settings instead of pushing it as an extra variable to oref. I don't like this setup as it basically contradicts what is written in a preferences. In prefs currently the 160mg/dL resides and for an adv. TT oref gets a second HBT value that overrides the current setting. Not so good. What do you guys think?

dnzxy:
Please do @mountrcg but in a separate PR. Daniel’s PR here can introduce the editability of presets and slightly change deletion behavior (alert, swipe action), but leaving logic as-is. You can fix / propose a change around the slider either in #154 or in a new PR 😊 let’s please not scopes here.

dsnallfot:
Demo with latest (last?) changes.
Experimental slider now available in edit mode

**Video**

https://github.com/nightscout/Trio/assets/72826201/129d71a3-3cb1-48cd-8132-73dfbc6882e3

bjornoleh (approved):
Tested on a phone, works as expected.

aug0211 (approved):
Testing on physical device for 24h - looks good!

FreeAPS/Sources/Modules/AddTempTarget/AddTempTargetStateModel.swift
```diff
@@ -189,5 +195,46 @@ extension AddTempTarget {
             }
             return Decimal(Double(target))
         }
+
+        func computePercentage(target: Decimal) -> Decimal {
+            let c = Decimal(hbt - 100)
```
dnzxy, line 200: Consider using more descriptive variables names instead of `c`.

dnzxy (approved):
Tested on simulator device, looks good and works as expected. Very minor comment on a single variable name; I‘m okay with moving forward as-is though 😊

FreeAPS/Sources/Modules/AddTempTarget/AddTempTargetStateModel.swift
bjornoleh: For context, this is probably copied from oref0:

https://github.com/nightscout/trio-oref/blob/d1dfb7099d59e4f7779e5fc5c75f8ca299d91ef2/lib/determine-basal/determine-basal.js#L920-L926

dsnallfot: I used the “c” to be consistent with the other existing HBT computation. Is it ok to leave it like that for now (we will likely need to redo this any year in @mountrcg PR)

dnzxy: Yes totally okay. I just wanted to note it 😊  The readability of the oref code is not really the best (as we know). No criticism on your work.

Sjoerd-Bo3 (approved):
Tested this with a live device for several days, edited some TT's during this period. And only found a thing with Experimental Setting, but that was due to my understanding of that. Using it a lot and works great.

dnzxy:
Merging with 4 approvals.

PR #243, opened 2024-05-27
Translations from Crowdin

Sjoerd-Bo3:
Translations from crowdin

# Please Squash Merge!


t1dude:
Crowdin token and project ID DM'ed on Discord.

Sjoerd-Bo3:
@bjornoleh How can we testrun this? Do you know?

bjornoleh:
> @bjornoleh How can we testrun this? Do you know?

I suppose you could create a dev-test branch, and edit the branch name in L5 to match? Then push something to that branch? You would also need to temporarily change the default branch to dev-test, otherwise the workflow is not available.

Sjoerd-Bo3:
> > @bjornoleh How can we testrun this? Do you know?
> 
> I suppose you could create a dev-test branch, and edit the branch name in L5 to match? Then push something to that branch? You would also need to temporarily change the default branch to dev-test, otherwise the workflow is not available.

Yeah I agree on that last part indeed. Will do tonight when I have some time.

t1dude:
This PR should be removed. Translations of this version of Trio are no longer maintained.

Issue #175, opened 2024-05-11
Duplicated carb entries


pheltzel:
--Added 30g of carbs at 5:58 a.m. on May 11 (log attached).
--But 2 entries appeared in the History screen in Trio

![IMG_9008](https://github.com/nightscout/Trio/assets/1165092/7aa70970-936e-46ed-a7b4-d7b32a631cb5)
![IMG_9009](https://github.com/nightscout/Trio/assets/1165092/e4996f15-a33f-4b82-939e-a1c4840acb6e)

I deleted the second errant entry. So sorry, don't have a screenshot for that. Should have captured that first. 

--Running Trio v0.1.0
--Branch: dev6f862aa
--Omnipod Dash, G7
 
[log-dupe-carb-entry.txt](https://github.com/nightscout/Trio/files/15283019/log-dupe-carb-entry.txt)

This may be an intermittent bug -- and possibly only a display issue. Same issue was reported here in the iAPS Github Issues area and may be related to writing to Health:

https://github.com/Artificial-Pancreas/iAPS/issues/450

A potential fix was suggested here:
https://github.com/Artificial-Pancreas/iAPS/pull/564

Could be user error, but I'm pretty sure I didn't enter this twice. 

I still have iAPS 4 on my iPhone 15, not sure if that could be related. I've been seeing how it differs from what Trio advises.

But I've experienced the bug reported above (Issue 450 in iAPS), and it was thought to be Apple Health related, so I turned it off. I've turned it off in Trio now as well.

itsmojo:
Not sure if this is another example of this problem or something else. On 5/14/24 I had an incident where I wanted to log some food with carbs, fats & proteins, but hitting Save and continue wasn’t taking the app to the next screen after multiple attempts. It appeared as if the Trio app was locked up. I evidently got out of that screen (using Close? I believe, but I’m not sure), but then I found that I had created many duplicate carb entries (I’d guess 70 - ~7 with my future carbs for each of my maybe 10 attempts to try to Save and continue with no response) that had to all be manually deleted. Unfortunately I was in a conversation & in a bit of a rush to recover when this happened and didn’t grab any screen shots or take careful mental notes of the details. I was running a version of Trio that was cloned on May 11 at 2 pm PDT at the time  of this incident.

pheltzel:
Thanks @itsmojo (also nice to see you again!) This absolutely be user error. 

[...]

PR #248, opened 2024-05-30
Ensure meal is only saved once

MikePlante1:
- Disables the `Save and continue` button after it's pressed until the AddCarbs module is reopened.
- Renames `saved` flag to `noteSaved` for clarity.
- Addresses #175

### DIFF
--- a/FreeAPS/Sources/Modules/AddCarbs/View/AddCarbsRootView.swift
+++ b/FreeAPS/Sources/Modules/AddCarbs/View/AddCarbsRootView.swift
@@ -8,7 +8,8 @@ extension AddCarbs {
         @StateObject var state = StateModel()
         @State var dish: String = ""
         @State var isPromtPresented = false
-        @State var saved = false
+        @State var noteSaved = false
+        @State var mealSaved = false
         @State private var showAlert = false
         @FocusState private var isFocused: Bool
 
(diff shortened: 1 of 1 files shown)


aug0211:
- Multiple days of testing in vitro and in vivo, 0 issues encountered
- Button grays out as expected immediately on tap
- Attempted to recreate instigating issue a number of times in vivo with a looping pump (and CGM) connected by entering carbs during loop cycle
- 0 instances resulted in duplicate carbs, despite my best efforts to break it
- 0 other issues observed over the course of multiple meal entries (both in vitro and in vivo)

aug0211 (approved):
See comments, tested in vitro and in vivo with 0 issue and was unable to reproduce initial issue (which is a good thing in this case).

dnzxy (approved):
Basing approval on code review only. LGTM. Precise, intentional changes. We have two test accounts by experienced testers.

dnzxy:
Merging with 2 testers (in-vitro and in-vivo) and 3 approvals.

MikePlante1:
From the code, I'm pretty confident the button won't be able to be pressed twice, but I've still been spamming the "save" button every time I log meals the past few days and haven't had it trigger twice, even when purposely entering during a loop cycle.

Issue #245, opened 2024-05-28
Nightscout CGM Import Settings fails with error


kskandis:
**Describe the bug**
I know Nightscout is not recommended for CGM but I thought I would test it. Trio connects fine to my Nightscout instance but Nightscout Config - Import Settings fails with the following error:

"The data couldn't be read because it isn't in the correct format."

**To Reproduce**
1. Go to Settings
2. Click on CGM
3. Scroll down to Nightscout under Type
4. After entering URL and Secret, confirm Connect
5. Tap on Import Settings, accept prompt to allow overwrite of existing settings
6. See error

**Expected behavior**
Nightscout settings should be imported.

**Screenshots**

**Smartphone (please complete the following information):**
iPhone SE 3rd Generation iOS 17.4.1

**Setup Information (please complete the following information):**
* Pump type: none yet, plan to use OmniPod Dash
* CGM type and CGM app: Nightscout 15.0.2
* Trio version, branch and git reference (see Trio Settings): dev 5faf162 Trio v0.1.0

**Additional context**
Unable to provide screenshot as the Nightscout CGM cannot be deleted and re-added.  Configuring Nightscout seems to be a one-time only process.  It would be nice to have a Delete option to allow a re-configuration of Nightscout CGM.

aug0211:
Does your selection of mg/dl vs mmol match your NS configuration?

My hunch is perhaps your Trio preferences still has mmol set (default starting point) and your NS is using mg/dl.

If this is the case, the fix is to set your preferences in Trio to match NS. It would also indicate that the error message in Trio isn’t clear enough.

kskandis:
> Does your selection of mg/dl vs mmol match your NS configuration?

Yes, Preferences - Glucose Units is mg/dL which matches my NS configuration.  I retried again after following your instructions to Delete the Nightscout service and then re-added it, connected, then Import Settings.  It still shows:

Import Error
The data couldn't be read because it isn't in the correct format.

dsnallfot:
Very strange. I have trouble locating that particular or similar Import error text string  in the Trio code ("The data couldn't be read because it isn't in the correct format."). 

A screenshot of the Import error alert would be of great help for further troubleshooting so I can see the exact phrasing.

And also, you wrote "Pump type: none yet, plan to use OmniPod Dash". Do you run a pump simulator? (The reason im asking is that there needs to be a pump simulator or actual pump set up for the import settings to work, since basal settings are written to the pump. If that fails, it should however show another import error message.

[...]

PR #249, opened 2024-05-30
Remove unused Import Settings in JSON that cause errors decoding JSON

kskandis:
Resolves [Nightscout CGM Import Settings fails with error](https://github.com/nightscout/Trio/issues/245) by:

1. Removing unused fields from the JSON struct which do not get saved or written to settings/pump.  These unused fields used Types that were inconsistent with Loop JSON.
2. Checks if JSON Profile was entered by Loop to ensure proper Default record is loaded.

### DIFF
--- a/FreeAPS/Sources/Models/RawFetchedProfile.swift
+++ b/FreeAPS/Sources/Models/RawFetchedProfile.swift
@@ -4,16 +4,12 @@ struct FetchedNightscoutProfileStore: JSON {
     let _id: String
     let defaultProfile: String
     let startDate: String
-    let mills: Decimal
     let enteredBy: String
-    let store: [String: ScheduledNightscoutProfile]
-    let created_at: String
+    let store: [String: FetchedNightscoutProfile]
 }
 
 struct FetchedNightscoutProfile: JSON {
     let dia: Decimal
-    let carbs_hr: Int
-    let delay: Decimal
     let timezone: String
     let target_low: [NightscoutTimevalue]
     let target_high: [NightscoutTimevalue]
--- a/FreeAPS/Sources/Modules/NightscoutConfig/NightscoutConfigStateModel.swift
+++ b/FreeAPS/Sources/Modules/NightscoutConfig/NightscoutConfigStateModel.swift
@@ -133,7 +133,9 @@ extension NightscoutConfig {
                 {
                     do {
                         let fetchedProfileStore = try jsonDecoder.decode([FetchedNightscoutProfileStore].self, from: data)
-                        guard let fetchedProfile: ScheduledNightscoutProfile = fetchedProfileStore.first?.store["default"]
+                        let loop = fetchedProfileStore.first?.enteredBy.contains("Loop")
+                        guard let fetchedProfile: FetchedNightscoutProfile = fetchedProfileStore.first?
+                            .store[loop! ? "Default" : "default"]
                         else {
                             error = "\nCan't find the default Nightscout Profile."
                             group.leave()


bjornoleh:
Thanks for your contribution. The changes look good, but I am not overly familiar with these details, and don’t have other looping data to test from. 

Hoping someone with fresh profile data uploaded to NS from both Loop and AAPS will have a chance to test. Import of manual NS profile entries would also be good to test here.

dsnallfot (approved):
No comments regarding the code. Seems reasonable to exclude what's  not needed from the fetch.

**Successfully tested by following these steps:**
1. Applied suggested changes from PR In simulator build
2. Cleared out basal/isf/cr/target and dia settings from simulator build
3. Found old MongoDB Loop profile settings entry, cloned that in MongoDB and applied current date as startDate
4. Saved this profile in Nightscout 
5. Run "Import settings" in Trio
6. Successfully fetched all Loop profile setting (basals/isf/cr/dia and target)
7. Deleted the cloned loop profile in MongoDB
8. Saved most current Trio profile (before the loop clone)  in Nightscout again
9. Run "Import settings" in Trio
10. Successfully fetched all Trio profile settings (basals/isf/cr/dia and target)

Works as expected regarding Loop Profile import. Doesn't break iAPS/Trio profile imports. ie Looks good to me. Approved

marionbarker (approved):
I tested this change - relying on others for code review.
I use Loop and have never been able to import settings into Trio before.
After applying the patch from this PR, I was able to import my Loop settings from my "real" nightscout site. I confirmed the imported settings in the Trio app of my test phone match my Loop settings: DIA, Basal Profile, ISF, CR and (low-end of) Target Range.

trio-pr-249

MikePlante1 (approved):
Code review looks good to me. I couldn't find any references to the deleted variables. Looking at the profiles in Nightscout I can see that Loop uses `Default` wheres Trio uses `default`, so that section of the code looks good too.

I tested in a simulator by uploading settings from Loop to a spare Nightscout account, then trying to import settings into Trio, which failed. After applying this PR, though, the settings were properly imported into Trio.

My only suggestion to consider is maybe setting target to the middle of the target range from Loop rather than the bottom target, since Trio only uses a single target.

Sjoerd-Bo3:
And maybe we need to make a separate PR for this?

> My only suggestion to consider is maybe setting target to the middle of the target range from Loop rather than the bottom target, since Trio only uses a single target.

@kskandis are you also up for that?

bjornoleh:
And possibly AAPS-generated profiles? @t1dude ?

bjornoleh:
> Could someone also quickly check that NS-generated profiles are also imported correctly?

I got to test this now, and can confirm that changes made from the NS profile editor are correctly imported. For target, only the lower bound is imported, the upper bound is ignored, as noted by others above. 

For me this is LGTM. Would be interesting to know if AAPS profiles are supported too, but that’s a little out of scope for this PR.

bjornoleh (approved):

dnzxy:
I will merge this with 4 approvals and 4 testers. 

@t1dude if you could please test this with AAPS-uploaded profiles after the fact; if a fix for AAPS-profiles is necessary, we will record this as a new issue ticket.

@MikePlante1 comment regarding not using just the lower bound of Loop's correction range can be handled via #252 .

Thank you everyone for testing and reviewing. Thank you @kskandis for your contribution to Trio.

kskandis:
> And maybe we need to make a separate PR for this?
> 
> > My only suggestion to consider is maybe setting target to the middle of the target range from Loop rather than the bottom target, since Trio only uses a single target.
> 
> @kskandis are you also up for that?

That makes sense to me.  Yes, I'll look into it.

Issue #252, opened 2024-05-31
NS Settings Import: Set median of lower and upper correction range bound when importing Loop settings


dnzxy:
## Is your feature request related to a problem? Please describe.
When importing settings from NS that were uploaded to NS from Loop, there are upper and lower correction range limits stored in NS. This is downloaded to Trio, but the lower end is set as target, as Trio does not have a correction or target range.

## Describe the solution you'd like
Do not set the lower bound of the correction range as target, but the median of upper and lower. 

## Additional context
This came up when testing PR #249 which solves issue #245

## User Impact
Ensures a safer Trio target, as Loop already aims to get to "center" of correction range; thus this is the target new Trio users will want to use upon importing their settings.

Sjoerd-Bo3:
@kskandis This is the one I was talking about. All yours if you want to!

PR #253, opened 2024-05-31
Set median of lower and upper correction range bound when importing L…

kskandis:
Resolves [NS Settings Import: Set median of lower and upper correction range bound when importing Loop settings](https://github.com/nightscout/Trio/issues/252) by:

Sets Trio's target correction low and high to median of lower and upper target range when NS profile's enteredBy is Loop.  Rounds median value to 0 or 1 significant digits depending on Glucose unit (0 for mg/dL, 1 for mmol/L).

EDIT:
This issue probably applies to AAPS, too, since it also uses a low and high target.  [examples.json](https://github.com/nightscout/AndroidAPS/blob/3df83535f9f8f3246308121900c2785c50915fa8/core/nssdk/src/main/kotlin/app/aaps/core/nssdk/remotemodel/examples.json#L332) shows its `profileJson` with `target_low` and `target_high`.  I cannot test it w/out having an actual "openaps://AndroidAPS" profile though.

I think merged PR, Remove unused Import Settings in JSON that cause errors decoding JSON #249, also applies to AAPS.

### DIFF
--- a/FreeAPS/Sources/Modules/NightscoutConfig/NightscoutConfigStateModel.swift
+++ b/FreeAPS/Sources/Modules/NightscoutConfig/NightscoutConfigStateModel.swift
@@ -241,7 +241,8 @@ extension NightscoutConfig {
                                 let median = loop! ? self.getMedianTarget(
                                     lowTargetValue: target.value,
                                     lowTargetTime: target.time,
-                                    highTarget: fetchedProfile.target_high, units: self.units
+                                    highTarget: fetchedProfile.target_high,
+                                    units: self.units
                                 ) : target.value
                                 return BGTargetEntry(
                                     low: median,
--- a/FreeAPS/Sources/Modules/NightscoutConfig/NightscoutConfigStateModel.swift
+++ b/FreeAPS/Sources/Modules/NightscoutConfig/NightscoutConfigStateModel.swift
@@ -84,6 +84,24 @@ extension NightscoutConfig {
             return NightscoutAPI(url: url, secret: secret)
         }
 
+        private func getMedianTarget(
+            lowTargetValue: Decimal,
+            lowTargetTime: String,
+            highTarget: [NightscoutTimevalue],
+            units: GlucoseUnits
+        ) -> Decimal {
+            if let idx = highTarget.firstIndex(where: { $0.time == lowTargetTime }) {
+                let median = (lowTargetValue + highTarget[idx].value) / 2
+                switch units {
+                case .mgdL:
+                    return Decimal(round(Double(median)))
+                case .mmolL:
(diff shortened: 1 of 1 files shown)


FreeAPS/Sources/Modules/NightscoutConfig/NightscoutConfigStateModel.swift
```diff
@@ -220,9 +238,14 @@ extension NightscoutConfig {
 
                         let targets = fetchedProfile.target_low
                             .map { target -> BGTargetEntry in
-                                BGTargetEntry(
-                                    low: target.value,
-                                    high: target.value,
+                                let median = loop! ? self.getMedianTarget(
+                                    lowTargetValue: target.value,
+                                    lowTargetTime: target.time,
+                                    highTarget: fetchedProfile.target_high, units: self.units
```
JeremyStorring, line 244: Apologies for being picky - can we put units on it's own line to be consistent?

kskandis: Thanks for the review. Done.

bjornoleh:
I think perhaps AAPS targets should be handled the same way as Trio targets, using only / mainly the lower target, instead of the median of lower/upper target range, as seems to be the case in this PR currently. We use the same algorithm, and I would expect the same behaviour regarding targets.

From https://androidaps.readthedocs.io/en/latest/Configuration/Config-Builder.html#closed-loop:
Please note: In closed loop mode a single target instead of target range (i.e. 5,5 mmol or 100 mg/dl instead of 5,0 - 7,0 mmol or 90 - 125 mg/dl) is recommended.

FreeAPS/Sources/Modules/NightscoutConfig/NightscoutConfigStateModel.swift
dnzxy: Resolving this conversation as it was addressed with commit `8b8645a`

dnzxy (approved):
LGTM (based on code review only, as I do not use Loop and cannot test Loop-imported settings).

dnzxy:
@marionbarker or @bjorkert – could you please test this, and if good, give your approval please?

marionbarker:
### Summary
Success: This PR worked as expected.

### Testing Details
I applied this patch to dev (commit 4d6e102a).
* Previously I had imported settings from Nightscout (Loop user).
* My Loop range is 90 to 100 mg/dL.
* My imported setting using this dev is 90 mg/dL.

After applying the patch and building again, I imported my settings again.
* My imported setting is now 95 mg/dL.

marionbarker (approved):

bjornoleh:
Merging with two approvals

PR #275, opened 2024-06-05
Add an option to confirm boluses faster on Apple Watch

MikePlante1:
Adds an option to reduce the amount you need to spin the crown on an Apple Watch to confirm a bolus.

⚙️>`Watch`>`Confirm Bolus Faster`
Confirm Bolus Faster

Has been in iAPS-dev since 2024-02-09: https://github.com/Artificial-Pancreas/iAPS/pull/508

### DIFF
--- a/FreeAPS/Resources/json/defaults/freeaps/freeaps_settings.json
+++ b/FreeAPS/Resources/json/defaults/freeaps/freeaps_settings.json
@@ -42,5 +42,6 @@
   "rulerMarks" : true,
   "maxCarbs": 250,
   "displayFatAndProteinOnWatch": false,
+  "confirmBolusFaster": false,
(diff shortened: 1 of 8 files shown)


bjornoleh (approved):
Functionality confirmed by building to Xcode simulator for iPhone 15 and S7 watch simulator.

The default was "off", and enabling the Confirm bolus faster feature did exactly that.

The code looks reasonable after a quick glance.

Given that this has also been used for a good while in iAPS, I'd say this is LGTM (unless someone objects to the implementation of such a feature).

dnzxy (approved):
LGTM. Been using this feature since forever since it got into iAPS.

Sjoerd-Bo3 (approved):

Issue #231, opened 2024-05-22
Pump/CGM state reverts on phone restart


dnzxy:
**Describe the bug** 
Upon phone restart, the app seldom reverts to old device state for CGM or pump, thus losing device connection. For pods this inherently leads to loss of device and need for replacement. This looks a lot like an old Loop bug is still present in Trio, but that mustn’t be the case as changes were migrated to FreeAPS and Trio in the past.

**Screenshots**
| Pod error | &nbsp; |
|--------|--------|
| ![IMG_0752](https://github.com/nightscout/Trio/assets/48965855/76b4c08a-3704-4e77-87bf-0594b2116fb8) | ![image0](https://github.com/nightscout/Trio/assets/48965855/d86b0b62-ba4d-4da5-ae23-91c9a0c4c8c0) | 

screenshots by user @Sjoerd-Bo3 showing loss of pod connection and pairing info in bluetooth settings

**Setup Information (please complete the following information):**
* all pump types 
* all CGMs
* Trio version 0.1.0

**Additional context**
Loop had a very similar issue, if not this issue, and the fix to this issue was moving device state info from UserDefaults to binary file (plist). See https://github.com/LoopKit/Loop/pull/1702

kskandis:
> This looks a lot like an old Loop bug is still present in Trio, but that mustn’t be the case as changed where migrated to FreeAPS and Trio in the past.

It looks like the https://github.com/LoopKit/Loop/pull/1702 fix is not imported to Trio.  Trio imports https://github.com/LoopKit/LoopKit but not https://github.com/LoopKit/Loop where the fix to this is implemented.  Some of the Trio corresponding code to LoopKit/Loop lives under https://github.com/nightscout/Trio/tree/dev/FreeAPS/Sources/APS, so Loop's fix needs to be merged into Trio's code, such as [Loopkit/Loop/Managers/DeviceDataManager.swift](https://github.com/LoopKit/Loop/pull/1702/files#diff-c17002aeaed37d4b71df1a6e2326b47b4938d7b6d4c86e93a92b731740510fd1),

dnzxy:
@avouspierre had hinted on Discord that these changes were already imported to FAX and in turn to iAPS, so also Trio. I only recorded the issue here. I also thought this was still present in iAPS/Trio. 

Quoting, as of 23/05/2024
> I quickly checked but trio (freeaps in fact) seems implemented the plist strategy for the persistent wrapper.. so  the fix of loop is probably already here..

[...]

PR #277, opened 2024-06-06
Port over pump loss fix via LoopKit/Loop#1702

dnzxy:
Ports over Loopkit/Loop#1702 to fix the loss of pumps and CGMs by no longer storing them as UserDefaults entry but via binary plist files. Storage of CGM info was already done via PR #15 back before Trio was opened for beta. This PR also migrates information storage for pump information.

Quoting from Loop PR:
> It's an attempt to fix the issue […] where Loop appears to revert to an old version of device state.
> Migration is supported coming from **previous versions of Loop that used UserDefaults**, but **reverting to a previous version of Loop will mean loss of Pump, CGM** […] as the previous version of Loop will not know to load the data from the binary plist files.

Thank you to @kskandis and @avouspierre for the great collaboration.

Addresses and solves #231

### DIFF
--- a/FreeAPS/Sources/Modules/Settings/View/SettingsRootView.swift
+++ b/FreeAPS/Sources/Modules/Settings/View/SettingsRootView.swift
@@ -61,12 +61,16 @@ extension Settings {
                                     .frame(maxWidth: .infinity, alignment: .trailing)
                                     .buttonStyle(.borderedProminent)
                             }
-                            HStack {
-                                Text("Delete Pump State nlist")
-                                Button("Delete") { state.resetLoopDocuments() }
-                                    .frame(maxWidth: .infinity, alignment: .trailing)
-                                    .buttonStyle(.borderedProminent)
-                            }
+// Commenting this out for now, as not needed and possibly dangerous for users to be able to nuke their pump pairing informations via the debug menu
+// Leaving it in here, as it may be a handy functionality for further testing or developers.
+// See https://github.com/nightscout/Trio/pull/277 for more information
+//
+//                            HStack {
+//                                Text("Delete Stored Pump State Binary Files")
+//                                Button("Delete") { state.resetLoopDocuments() }
+//                                    .frame(maxWidth: .infinity, alignment: .trailing)
+//                                    .buttonStyle(.borderedProminent)
+//                            }
                         }
                         Group {
                             Text("Preferences")
--- a/FreeAPS/Sources/Modules/Settings/View/SettingsRootView.swift
+++ b/FreeAPS/Sources/Modules/Settings/View/SettingsRootView.swift
@@ -61,9 +61,9 @@ extension Settings {
                                     .frame(maxWidth: .infinity, alignment: .trailing)
                                     .buttonStyle(.borderedProminent)
                             }
-// Commenting this out for now, as not needed and possibly dangerous for users to be able to nuke their pump pairing informations via the debug menu
-// Leaving it in here, as it may be a handy functionality for further testing or developers.
-// See https://github.com/nightscout/Trio/pull/277 for more information
+                            // Commenting this out for now, as not needed and possibly dangerous for users to be able to nuke their pump pairing informations via the debug menu
+                            // Leaving it in here, as it may be a handy functionality for further testing or developers.
+                            // See https://github.com/nightscout/Trio/pull/277 for more information
 //
 //                            HStack {
 //                                Text("Delete Stored Pump State Binary Files")
--- a/FreeAPS/Sources/APS/DeviceDataManager.swift
+++ b/FreeAPS/Sources/APS/DeviceDataManager.swift
@@ -43,10 +43,6 @@ private let staticPumpManagersByIdentifier: [String: PumpManagerUI.Type] = [
     MockPumpManager.pluginIdentifier: MockPumpManager.self
 ]
 
-// private let staticPumpManagersByIdentifier: [String: PumpManagerUI.Type] = staticPumpManagers.reduce(into: [:]) { map, Type in
-//    map[Type.managerIdentifier] = Type
-// }
-
 private let accessLock = NSRecursiveLock(label: "BaseDeviceDataManager.accessLock")
 
 final class BaseDeviceDataManager: DeviceDataManager, Injectable {
@@ -78,7 +74,8 @@ final class BaseDeviceDataManager: DeviceDataManager, Injectable {
         didSet {
             pumpManager?.pumpManagerDelegate = self
             pumpManager?.delegateQueue = processQueue
-            UserDefaults.standard.pumpManagerRawValue = pumpManager?.rawValue
+            rawPumpManager = pumpManager?.rawValue
+            UserDefaults.standard.clearLegacyPumpManagerRawValue()
             if let pumpManager = pumpManager {
                 pumpDisplayState.value = PumpDisplayState(name: pumpManager.localizedTitle, image: pumpManager.smallImage)
                 pumpName.send(pumpManager.localizedTitle)
@@ -105,6 +102,8 @@ final class BaseDeviceDataManager: DeviceDataManager, Injectable {
         }
     }
 
+    @PersistedProperty(key: "PumpManagerState") var rawPumpManager: PumpManager.RawValue?
+
     var bluetoothManager: BluetoothStateManager { bluetoothProvider }
 
     var hasBLEHeartbeat: Bool {
@@ -123,7 +122,11 @@ final class BaseDeviceDataManager: DeviceDataManager, Injectable {
     }
 
     func setupPumpManager() {
-        pumpManager = UserDefaults.standard.pumpManagerRawValue.flatMap { pumpManagerFromRawValue($0) }
+        if let pumpManagerRawValue = rawPumpManager ?? UserDefaults.standard.legacyPumpManagerRawValue {
+            pumpManager = pumpManagerFromRawValue(pumpManagerRawValue)
+        } else {
+            pumpManager = nil
+        }
     }
 
     func createBolusProgressReporter() -> DoseProgressReporter? {
@@ -163,20 +166,6 @@ final class BaseDeviceDataManager: DeviceDataManager, Injectable {
                 self.updateUpdateFinished(true)
             }
         }
-
-//        pumpUpdateCancellable = Future<Bool, Never> { [unowned self] promise in
-//            pumpUpdatePromise = promise
-//            debug(.deviceManager, "Waiting for pump update and loop recommendation")
-//            processQueue.safeSync {
-//                pumpManager.ensureCurrentPumpData { _ in
-//                    debug(.deviceManager, "Pump data updated.")
-//                }
-//            }
-//        }
-//        .timeout(30, scheduler: processQueue)
-//        .replaceError(with: false)
-//        .replaceEmpty(with: false)
-//        .sink(receiveValue: updateUpdateFinished)
     }
 
     private func updateUpdateFinished(_ recommendsLoop: Bool) {
@@ -186,11 +175,6 @@ final class BaseDeviceDataManager: DeviceDataManager, Injectable {
             warning(.deviceManager, "Loop recommendation time out or got error. Trying to loop right now.")
         }
 
-        // directly in loop() function
-//        guard !loopInProgress else {
-//            warning(.deviceManager, "Loop already in progress. Skip recommendation.")
-//            return
-//        }
         self.recommendsLoop.send()
     }
 
@@ -319,7 +303,8 @@ extension BaseDeviceDataManager: PumpManagerDelegate {
     }
 
     func pumpManagerDidUpdateState(_ pumpManager: PumpManager) {
-        UserDefaults.standard.pumpManagerRawValue = pumpManager.rawValue
+        rawPumpManager = pumpManager.rawValue
+        UserDefaults.standard.clearLegacyPumpManagerRawValue()
         if self.pumpManager == nil, let newPumpManager = pumpManager as? PumpManagerUI {
             self.pumpManager = newPumpManager
         }
@@ -536,29 +521,6 @@ extension BaseDeviceDataManager: DeviceManagerDelegate {
     ) {}
 
     func recordRetractedAlert(_: Alert, at _: Date) {}
-
-//    func scheduleNotification(
-//        for _: DeviceManager,
-//        identifier: String,
-//        content: UNNotificationContent,
-//        trigger: UNNotificationTrigger?
-//    ) {
-//        let request = UNNotificationRequest(
-//            identifier: identifier,
-//            content: content,
-//            trigger: trigger
-//        )
-//
-//        DispatchQueue.main.async {
-//            UNUserNotificationCenter.current().add(request)
-//        }
-//    }
-//
-//    func clearNotification(for _: DeviceManager, identifier: String) {
-//        DispatchQueue.main.async {
-//            UNUserNotificationCenter.current().removeDeliveredNotifications(withIdentifiers: [identifier])
-//        }
-//    }
 
     func removeNotificationRequests(for _: DeviceManager, identifiers: [String]) {
         DispatchQueue.main.async {
--- a/FreeAPS/Sources/APS/Extensions/PumpManagerExtensions.swift
+++ b/FreeAPS/Sources/APS/Extensions/PumpManagerExtensions.swift
@@ -2,6 +2,8 @@ import LoopKit
 import LoopKitUI
 
 extension PumpManager {
+    typealias RawValue = [String: Any]
+
     var rawValue: [String: Any] {
         [
             "managerIdentifier": pluginIdentifier, // "managerIdentifier": type(of: self).managerIdentifier,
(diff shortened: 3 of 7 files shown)


avouspierre:
I tried with iphone 8 + RPI Dash simulator with last code : 
- Complete new pump installation + OK : ✅
- kill the app after a complete installation - reopen OK with the pod : ✅
- kill the app during the installation of a new pod -after pod insertion - :  Message "Finish setup" > Ok (after restarting with pod -q only) :  ✅
- kill the app during the installation of a new pod -after pod pair - :  Message "Finish setup" > Ok (after restarting with pod -q only) :  ✅

avouspierre (approved):

marionbarker:
### Problems with PR:
I think we are missing something. I believe this should remember the pump when going from dev to the new branch (but the reverse should not be true.) (Error 1)

Also, with the PR in place, deleting the insulin delivery system doesn't seem to get rid of it. (Error 2) This is for Omnipod ONLY. Did not happen for Medtronic.

### Configuration
* iPhone SE
* dev is commit: 14c76432
* pump-loss-fix-migration is commit: c9e747a6

Noticed the transition problem - which I thought was wrong.
Deleted all apps on phone with shared app group and started fresh, same problem.

### Testing - DASH simulator
* build dev, attach rPi DASH simulator
* (Error 1): build dnzxy:pump-loss-fix-migration, no pump seen in main screen
* build dev, pod is there again
   * deactivate pod
   * main screen shows no pod
* build dnzxy:pump-loss-fix-migration
   * (Error 1): no pump seen in main screen
   * add Omnipod DASH pump
   * connect to rPi DASH simulator
   * build dev over app, it says no pod (doesn't loose the pump selection)
      * I expect this direction to lose the pod
   * Return to dnzxy:pump-loss-fix-migration, connects to the rPi with no issues
   * Deactivate pod, switch to other insulin delivery
   * (Error 2): but main screen still shows no pod, even though tapping on pump shows no pump is attached

### Testing - Medtronic
* build dev, attach medtronic (MDT 515 pump)
* (Error 1): build dnzxy:pump-loss-fix-migration, no pump seen in main screen
* build dev, MDT 515 is there again and still connected. 
   * confirm that removing Medtronic pump shows no pump icon in Main screen
* build dnzxy:pump-loss-fix-migration, no pump seen in main screen
   * add MDT 515
   * (Error 2 - does NOT happen for MDT pump): delete MDT pump, main screen shows add pump

bjornoleh:
Test:

Not fresh install, various branches had been build in the past, but:

- Build dev [14c7643](https://github.com/nightscout/Trio/commit/14c7643275e3f351448d7623432f0d1778b65af4)
- Attach old, expired Dash pod (NXP)
- Build [pump-loss-fix-migration](https://github.com/dnzxy/Trio/tree/pump-loss-fix-migration)
- Dash pod was connected and worked as it should, except that it had gone silent. I expect this to be related to the fact that the pod expired in 2021, and unrelated to this PR

Testing in reverse, going from [pump-loss-fix-migration](https://github.com/dnzxy/Trio/tree/pump-loss-fix-migration) to dev, the pod was lost (and the sim pump had been automatically selected for some reason).

kskandis:
Thank you for your thorough testing!

> Also, with the PR in place, deleting the insulin delivery system doesn't seem to get rid of it. (Error 2) This is for Omnipod ONLY. Did not happen for Medtronic.

This issue pre-exists in dev, too, so it isn't related to this PR.  We should open a separate issue for it?

For other issues, I think once this PR is installed on the iPhone, the nlist pump state persists and will be the first choice as the method of retrieval.  So changing pump state on a Dev installation will not be supported once this PR is installed.  Is this the way Loop also works?  Loop does have [ResetLoopManager](https://github.com/LoopKit/Loop/blob/6cf0285e13f6523339b6a4758558cd66d0153c2e/Loop/Managers/ResetLoopManager.swift#L25) which gets called from LoopAppManager on launch.  We could use this to clear nlists documents for testing.

dnzxy:
Can only join @kskandis and thank you all for rigid testing (and even burning pods). Thank you! To address the feedback and question:

The expected behaviour should be 
1. move from `dev` (old setup) to PR code (new setup) -> pump is kept
2. move from PR code to `dev` or any branch that has old setup -> pump is lost 

It looks like that @marionbarker seems to have lost the pod (rPi) on both cases, while @bjornoleh's test showed the expected behaviour. I'm unsure how to proceed here.

@marionbarker can you please check if this may be a UI glitch, where the home screen shows no pump ("Add pump"), but when you tap it, it is actually already connected? I have experienced this behaviour with the in-app simulator before and reported it on Discord (no issue so far); just to make sure if the actual pairing is lost, or if the UI just is not updated properly.

bjornoleh:
### Failure during testing:

- Built Trio dev [14c7643](https://github.com/nightscout/Trio/commit/14c7643275e3f351448d7623432f0d1778b65af4) from scratch to a different phone (SE2), with no Loop, iAPS, Trio or xDrip4iOS apps installed
- Parired NXP Dash pod
- Built [pump-loss-fix-migration](https://github.com/dnzxy/Trio/tree/pump-loss-fix-migration) on top
  - **lost the pod**

dnzxy:
@kskandis and I will be looking at this more closely and be working towards a solution hopefully later today.

I think it's fair to summarize: we've encountered two issues here (which we will look at):
1. Pump is lost when going FROM UserDefault stored pump pairing TO binary file storage pump pairing
    a. Loss of pump pairing upon REVERSAL from binary file storage to UserDefaults storage is expected and deemed okay.
3. After pump removal, the UI shows "No pod" (= no paired device) but should actually show "Add pump" (=no selected insulin delivery method)

dnzxy:
With the recent changes introduced by @kskandis I can now report the following:

I can now successfully
- pair pod on current feature branch w/ applied changes
- switch back to dev and rebuild
- pump is lost now (expected)
- pair new pod to dev
- successfully paired pod on dev ->  switch to feature branch w/ applied patch -> rebuild -> pump still connected

I've force closed, restarted phone, rebuild feature branch multiple times, pod stays paired. Ultimately, pod persists going from dev -> feature; is lost feature -> dev; works as designed.

There is also a (temporary ❓ ) debug feature to clean the binary files for paired pumps hidden under the debug toggle. This cleans out the files; may be useful for testing this PR and possibly beyond.

marionbarker:
### Status
Success for Omnipod DASH
Comments:
* transition from dev to to pump-loss-fix-migration works as expected
* however, when you return to dev from pump-loss-fix-migration two things happen:
   * pump is lost and must be added back (this is expected)
   * subsequent transition from dev to to pump-loss-fix-migration loses the pod unless you delete the app before returning to dev - in that case, expected smooth transition works

This comment has gotten pretty long. Do the Medtronic testing in a new comment.

### Configuration
*  pump-loss-fix-migration is commit 8a6e8f18 - call this fix for short
* dev is commit 14c76432 (same as yesterday)
* iPhone SE (yesterday wiped out all programs with shared app group, only Trio exists on phone)

### Testing - DASH

#### Test 1: Success: pod maintained going from dev to fix
* initial condition: dev with rPi DASH paired
* build fix: rPi is still connected (bolus command worked)
* NOTE - come back to this later in test 4

#### Test 2: Success: pod lost going from fix to dev
* build dev: 
  * As expected - lose pump:
  * Trio says Error: pump not set
  * main screen says Add Pump
  * add pump and it works (for dev)

#### Test 3A: Fail: pod NOT maintained going from dev to fix
* return to fix and see that pod is maintained
* but says comms are lost and I'm not able to restore
   * See Test 4
* discard pod and pair fresh

#### Test 3B: Success - fix works for all pod commands
* now test every pod command before moving forward
   * pair and insert (interrupt each and rebuild with successful behavior)
   *  suspend, resume
   * set and cancel temp basal
   * modify confidence reminders and notifications
   * all pod diagnotics
   * bolus
   * loop is green

#### Test 4: Success: Try the dev to fix transition again
When this was done for Loop, you had to delete the app to go backwards.
Delete Trio (no other apps with shared app group are on the phone)
* build dev and import settings from NS
* attach rPi DASH
* check that all functions work (see Test 3B)
* build fix on phone, pod is still there and communications seem to work
* test every pod command

#### Test 5: repeat the dev - fix test
Try again building dev over fix without deleting first, then return to fix
* pod is lost as expected - main screen says add pump
* pair rPi DASH
* This time, ^C the rPi before the rebuild.
* build fix over dev
* restart rPi and the issue seen in Test 3A persists.

marionbarker:
### Status
MDT Test #1 a success. Will do the rest of the MDT testing later.

### Configuration
*  pump-loss-fix-migration is commit 8a6e8f18 - call this fix for short
* dev is commit 14c76432 (same as yesterday)
* iPhone SE - no app on phone (all programs with shared app group previously deleted)

### Testing - Medtronic

#### Test 1: Success: MDT maintained going from dev to fix
* initial condition: build dev onto blank phone
* add Nightscout and import settings
* pair with MDT 515 pump
* wait for green loop, suspend/resume delivery and bolus
* build fix over dev
* maintains connection to MDT
* bolus, suspend/resume, green loop

Need to take a break. I'll do the return to dev testing for MDT later. I'll leave the test phone running closed loop with MDT and fix branch running.

kskandis:
Thanks for your detailed comments and testing!!

> however, when you return to dev from pump-loss-fix-migration two things happen:
> * pump is lost and must be added back (this is expected)
> * subsequent transition from dev to pump-loss-fix-migration loses the pod unless you delete the app before returning to dev - in that case, expected smooth transition works

In the case where you are moving from PR to Dev, you should use the new Settings - Debug Option "Delete Pump State nlist" Button to remove the persistent nlist doc.  Otherwise when you re-install the PR, it will still exist and the PR will use that (rather than the legacy, Dev pump settings) for the pump setttings.

avouspierre:
> In the case where you are moving from PR to Dev, you should use the new Settings - Debug Option "Delete Pump State nlist" Button to remove the persistent nlist doc. Otherwise when you re-install the PR, it will still exist and the PR will use that (rather than the legacy, Dev pump settings) for the pump setttings.

Tested with simulator with success. But probably to remove when merged because no reason to open this option...avoid to delete plist associated with a pump without any reason...

avouspierre (approved):

dnzxy:
> ### Status
> Success for Omnipod DASH Comments:
> 
> * transition from dev to to pump-loss-fix-migration works as expected
> * however, when you return to dev from pump-loss-fix-migration two things happen:
>   
>   * pump is lost and must be added back (this is expected)
>   * subsequent transition from dev to pump-loss-fix-migration loses the pod unless you delete the app before returning to dev - in that case, expected smooth transition works

I cannot reproduce this behaviour. 

Here's what I did:
- Clean phone, no other self-built app on phone, fresh iPhone build (reset to factory settings
- Checkout Trio `dev`
- Build to test device
- Pair pod using rPi pod simulator
- Pairing process successful
- Play test beeps to verify
- Checkout feature-branch on commit `8a6e8f1``
- Rebuild from Xcode to phone (NOTE: pod was still paired, not removed, no nothing!)
- Pod is still paired just fine, UI showing reservoir and battery
- Play test beeps to verify connection
- See screenshot for rPi logs
image
- Checkout `dev` again
- Rebuild to test phone -> going from feature-branch commit `8a6e8f1` to `dev`
- Pod is completely lost
- UI shows `Add Pump` and when tapped guides me through the initial setup of the Omnipod DASH (or any other pump of my choosing)
- rPi shows it's still `*** Waiting for the next command ***` and `Listening for commands`

I did not (!) have to use the debug reset functionality. This worked for me out-of-the-box.

How do we want to proceed with the PR? I agree with Pierre, that the debug option should only be kept for testing and must be removed for proper use of the app – nuking pod details is just too big of a risk even when hidden behind the debug toggle.

kskandis:
> * I did not (!) have to use the debug reset functionality.

You only need to use the debug reset functionality if you return back to Dev after installing PR, and then, re-installing PR, so multiple round trips of installation which would a user would not do.  I know confusing!

Dev -> PR -> Dev -> PR, so multiple tests.  If Debug nlist Reset is NOT performed, the 2nd PR install will not see the 2nd Dev's pump.  It will see the 1st PR's pump because nlist is still active!

So install Dev (add pod) -> install PR (see Dev's Pod) -> Tap Debug nlist Reset -> install Dev (Add Pump is displayed, add a pump) -> install PR (see Dev's pump)

marionbarker:
### Summary - Success
* Thanks for the information about the debug option
* Used that prior to building dev over fix, this resolved the requirement to delete the app before building dev
   * See Medtronic Test 2 and Omnipod Test 6 for successful demonstration

### Configuration
* pump-loss-fix-migration is commit 8a6e8f18 - call this fix for short
* dev is commit 14c76432
* iPhone SE - only Trio on phone

### Testing - Medtronic (continued)

#### Test 2: build dev on top of fix
* while running fix, **use debug option to delete pump state nlist**
* build dev over fix
* pump is gone main screen shows add pump (as expected)
* add MDT back
* confirm MDT is working, suspend, resume, bolus, green loop

#### Test 3: transition from dev to fix
* now build fix over dev again
* pump is still there but is showing signal loss (not uncommon for MDT)
* these functions work: bolus, suspend, resume
* wait to see if signal loss indicator gets restored over time - takes 2 cycles, but green loop does get resolved.

### Testing - DASH, part 2

#### Test 6 (for DASH): build dev over fix and then return to fix - revisited
* start with fix (from Medtronic test immediately preceding this one)
* delete medtronic pump
* add DASH and pair rPi
* exercise the Omnipod functions - accidentally got into the state that needs OmniBLE fix (see ### OmniBLE Test below)
* restore proper operation, confirm all commands work
* while running fix, **use debug option to delete pump state nlist**
* build dev over fix - lose pod and pump selection as expected
* attach new rPi pod to dev, make sure it works
* build fix over dev, pod is still there and it works correctly
* verify all pod functions are good, have a green loop


### OmniBLE Test for OmniBLE PR 125
While running the DASH testing to first suspend and then resume above:
* got into a state where pod state of rPi does not match what Trio thinks
* time stamp is just before 13:00 PDT
   * hx="1724b075240a1d08001f100000000fff0000"
   *   extended_bolus_active =   False,   immediate_bolus_active =   False,   temp_basal_active =   False,   basal_active =   False
   * but Trio thinks pod resumed successfully and is running scheduled basal of 0.7 U/hr
   * modify the OmniBLE submodule to point at improved_delivery_status
   * rebuild the app
   * it now comes up saying insulin suspended - this is correct! (Add this log to OmniBLE PR 123 as proof this fix works)
   * upload the log file and csv file under that PR to demonstrate the code worked as expected.

marionbarker (approved):
I approve this as is. There is some question as to whether the debug option to delete (nlist or plist) should be left in. I have no strong opinion on this.

I tested for DASH and Medtronic and both worked as expected.

bjornoleh:
It seems like this PR is coming together now. 

Question: I had different experiences transitioning from dev to fix depending on having either the all previously installed, or more likely having other apps with the shared app group (xdrip4iOS and possibly other instances of Trio or iAPS). Will the fix work correctly also in these cases now?

FreeAPS/Sources/Modules/Settings/View/SettingsRootView.swift
```diff
@@ -61,6 +61,12 @@ extension Settings {
                                     .frame(maxWidth: .infinity, alignment: .trailing)
                                     .buttonStyle(.borderedProminent)
                             }
+                            HStack {
+                                Text("Delete Pump State nlist")
```
bjornoleh, line 65: Let’s remove the debug option for deleting plist before merging, to reduce risk of accidental removals. 

Question: is nlist synonymous with plist?

bjornoleh (changes requested):

FreeAPS/Sources/Modules/Settings/View/SettingsRootView.swift
dnzxy: Commented out with a comment pointing to this PR :)

marionbarker:
Test again with latest commit (comment out the debug: delete pump button).

### Status

### Configuration
This continues directly from last comment I made in this PR.
* Trio branch pump-loss-fix-migration is at commit 3ac4c317 (delete pump button is commented out)

### Testing
* Trio has rPi-4 DASH attached with earlier commit of fix branch
* build new commit of fix branch, no issue, pod still is communicating
   * deactivate pod from Trio
* Delete MDT pump from Loop
* Add MDT pump to Trio
   * wait for green loop
 * Build dev over fix (expect pump to be lost) - success - pump is lost but don't add the pump back
 * Build fix over dev - success (pump is still there because the information in plist is still valid for MDT)
    * Confirm I can suspend/resume the MDT

marionbarker (dismissed):
Tested before the final formatting change.

marionbarker (approved):
This modification was requested because Xcode was auto-formatting the files to make the comments line up and that resulted in changes in the local clone with respect to GitHub.

Code review of final commit.

- This is only a change to the format for some comment lines.
- There is no need to retest functionality of the code.
- I built with Xcode and confirmed there are no modifications to files automtically made by Xcode

bjornoleh (approved):

bjornoleh (approved):
Approving based on initial testing (partial success) and review of Marion's successful testing of the last changes after my tests with actual pods.

Verified that the app builds without uncommitted changes.

PR #288, opened 2024-06-08
Crowdin integration in GH Actions and configuration for Crowdin itself

Sjoerd-Bo3:
Template only, needs to be configured

- [x] Waiting for GH PAT in secrets from @Tornado-Tim 
- [x] Waiting for Crowdin token and project ID in secrets from @t1dude 
- [x] Configuring of the Action with @t1dude @dnzxy @Sjoerd-Bo3
- [ ] Testing of the Action @Sjoerd-Bo3 
- [ ] Look into Squashing of Crowdin commits (1 Commit per PR)

### DIFF
--- a/.github/workflows/crowdin.yml
+++ b/.github/workflows/crowdin.yml
@@ -16,13 +16,13 @@ jobs:
         uses: crowdin/github-action@v1.20.4 
         with:
           upload_sources: true
-          upload_translations: false
+          upload_translations: true
           download_translations: true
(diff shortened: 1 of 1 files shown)


t1dude (dismissed):

t1dude:
This topic should be revisited after 1.0 is released. I'm not sure we should automate Crowdin -> GitHub, but this should be decided after an open discussion.

PR #296, opened 2024-06-10
[DanaKit] Progress update 10-06-2024

bastiaanv:
## Describe the bug
In the current dev/alpha branch, whenever you do a bolus, the `lastStatusPumpDateTime`-state doesn't get updated. This state is used to fetch the history from the pump. If this value doesn't get updated after a bolus, the history sync will re-fetch the bolus. This will cause a double insulin entry for every bolus/SMB.

### Screenshot of issue:
IMG_4587 png
(provided by rig22 in the iAPS discord -> Dana channel)

## Note:
This bug makes the DanaKit project very unstable/unusable, therefore it would be nice to merge this asap


AndreasStokholm:
Having a quick glance at the code, I can't help but notice the same chunk copy/pasted 3 times throughout the code. Wouldn't it make more sense to extract this code to a function to ensure it doesn't get out of sync in the future?

bastiaanv:
@AndreasStokholm Fair point, updated the code to use a single function to fetch the pump's time!

AndreasStokholm (approved):
LGTM 👏

bjornoleh (approved):
Verified that this builds ok with no uncommitted changes, otherwise trusts the reviews of Andreas and the author.

Merging into the dana feature branch with two approvals

PR #298, opened 2024-06-11
Include `high` and `low` values in TIR in Statistics

MikePlante1:
This changes the statistics to include the `high` and `low` values in the normal range instead of the high and low ranges. This would match the colors of the glucose dots on the Statistics screen.

For this example in a simulator, I kept `high`/`low` to the default `70`/`180` and manually entered these values:

glucose values

| Before this PR (standing) | Before this PR (laying) |
| --- | --- |
| old standing | old laying |
|  After this PR (standing) |  After this PR (laying) |
| new standing | new laying |

### DIFF
--- a/FreeAPS/Sources/Modules/Stat/View/ChartsView.swift
+++ b/FreeAPS/Sources/Modules/Stat/View/ChartsView.swift
@@ -114,7 +114,7 @@ struct ChartsView: View {
                 type: NSLocalizedString(
                     "Low",
                     comment: ""
-                ) + " (≤\(low.formatted(.number.grouping(.never).rounded().precision(.fractionLength(1)))))",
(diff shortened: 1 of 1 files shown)


bjornoleh (approved):
### Code review:

The code changes looks good and are limited to the required changes for the scope of this PR.

### Test build to iPhone 13 mini:

Builds: ✅ 
TIR definition ok both for standing and laying chart ✅ 
mmol/L tested ok: ✅ 
- 3,8 below range
- 3,9 within range
- 10,0 within range
- 10,1 above range

No uncommitted auto linting: ✅

AndreasStokholm (approved):
Tested in sim and on a test phone with a Pi sim. Works and shows values as expected in both mg/dL and mmol/L.

Also, big plus for removing the ligatures 👏

bjornoleh:
Merging with two 👍

Issue #265, opened 2024-06-02
Extra messages exchanged with Omnipod


marionbarker:
## Describe the bug
The number of messages exchanged for the same actions with Trio is greater than with Loop. The extra exchanges with Trio appear to be getStatus followed by status response. These should not require a lot of power, but some people are complaining of pod faults with Trio. Extra messages can mean more battery drain and thus increase the likelihood of faults.

## Attach a Log
I will add some specific comparisons in a comment. I specifically observed this when looking at a suspend resume action. I will evaluate additional actions when preparing my comment.

## To Reproduce
*Steps to reproduce the behavior:*
I will include the specific steps when I add some specific comparisons in a comment.

## Expected behavior
The number of messages and message types exchanged between the phone and the pod were carefully adjusted during Loop development to be as small as possible while doing the required task. 

(There was some relaxation of this design with the advent of BLE communications, but most of the efficiencies have been reinstated for Loop and the OmniBLE/Kit submodules.)

## Screenshots
N/A

### Smartphone:
N/A

### Pump:
* Manufacturer: Insulet
* Model: probably both Eros and DASH, use the rPi DASH simulator for testing

### CGM:
N/A

### Trio Version:
The explicit version / commit numbers will be listed in the comment with the comparisons

## Technical Details
@itsmojo suggests that the extra get status calls observed when suspending and resuming in Trio come from the `self.updateStatus()` call in `clearBolusReporter()` in `FreeAPS/Sources/APS/APSManager.swift`. He wants to reserve his focus for the OmniBLE/Kit submodules, so is turning analysis of extra messages over to the Trio developers.

## Additional context
*Add any other context about the problem here.*

marionbarker:
### Summary
* I established exactly which pod buttons are sending too many getStatus commands to the pod
* Based off the hint provided by @itsmojo, I commented out some code (see end of this comment) and repeated the tests
* This needs someone else to generate the preferred code change in a PR
* I'll be happy to test the PR when it is prepared

### Configuration

* iPhone 8 running Trio alpha, commit 217d1d65
* iPhone 7+ running Loop dev, commit `c4b4588`

### Method

Examine the response to each command in the rPi DASH simulator. 
Compare to Loop.

For each command, there is an 0x1d message returned from the pod.
* Each command / response is parsed as needed using a python parser
* The first 0x1d message shows the correct state
* The additional 0x0e (getStatus) commands are completely unnecessary
[...]

PR #301, opened 2024-06-13
exclude cancelling from bolusTrigger, comment out updateStatus

kskandis:
This PR resolves this issue by excluding `BolusState.canceling` from sending a `bolusTrigger.send(false)` to APSManager which would tripper a `bolusReporter?.removeObserver`.

This is because if the `bolusReporter` observer is prematurely removed then the cancel completion will not be observed. APSManager separately observes [CancelBolus](https://github.com/nightscout/Trio/blob/217d1d65d7e40753d2c8988fef517d2687acdb07/FreeAPS/Sources/APS/APSManager.swift#L463) and then will remove the `bolusReporter` so there is no reason to include `BolusState.canceling` in the trigger.

I tested this in Mock and Cancel Bolus succeeds. I also commented out `updateStatus()`.  Once PR is tested, `updateStatus()` func should be removed.

### DIFF
--- a/FreeAPS/Sources/APS/APSManager.swift
+++ b/FreeAPS/Sources/APS/APSManager.swift
@@ -921,7 +921,7 @@ final class BaseAPSManager: APSManager, Injectable {
         bolusReporter = nil
         processQueue.asyncAfter(deadline: .now() + 0.5) {
             self.bolusProgress.send(nil)
-            self.updateStatus()
+            // self.updateStatus()
         }
     }
 }
--- a/FreeAPS/Sources/APS/APSManager.swift
+++ b/FreeAPS/Sources/APS/APSManager.swift
@@ -470,6 +470,7 @@ final class BaseAPSManager: APSManager, Injectable {
                 debug(.apsManager, "Bolus cancelled")
             }
 
+            self.updateStatus()
             self.bolusReporter?.removeObserver(self)
             self.bolusReporter = nil
             self.bolusProgress.send(nil)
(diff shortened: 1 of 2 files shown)


marionbarker:
The patch you posted in #265 was not successful with a rPi DASH simulator. 
I'm going to return to the Issue to post results in comments.
I also have some more suggestions from @itsmojo for things to look at which I will also post under the issue.
I definitely need and appreciate the help.

marionbarker:
@kskandis Please reopen this PR. Once we land an update to OmniBLE and OmniKit, this will be relevant again. Only the first commit in this PR is needed (the second commit actually introduces additional getStatus).

@itsmojo - your OmniBLE modification worked
* configuration:
   * latest dev (commit bd566014)
   * applied the suggestion from your PM along with changes to Trio in this PR (first commit only)
   * testing with an rPi using the modifications in the PR
   * I will do additional testing with both Trio and Loop, but this looks very promising
   * no extra getStatus command observed and can bolus again as soon as the previous bolus finishes or if previous bolus is canceled.
   * setting and canceling a Manual Temp Basal also appeared to work.

I will update the truth table in #265 and let this run in closed loop to examine the log files.

kskandis:
> Please reopen this PR. Once we land an update to OmniBLE and OmniKit, this will be relevant again. Only the first commit in this PR is needed (the second commit actually introduces additional getStatus).

I don't seem to have permission to re-open this PR (reopen is greyed out).  I can create a new one ...

marionbarker:
That is fine. I think there was a patch in #265 we should look at. May as well wait for the OmniBLE PR.

kskandis:
> That is fine. I think there was a patch in #265 we should look at. May as well wait for the OmniBLE PR.

Okay.  I just created the new PR [here](https://github.com/nightscout/Trio/pull/351) though if it is useful... I'll leave it open for now.

PR #310, opened 2024-06-15
Live activity improvements

10nas:
- Add back live previews for the simple live activity view.
- Refactor `LiveActivityAttributes.ContentState`.
- Change `LiveActivity` structure for readability.
- Fix cases where text got clipped in the simple live activity. Live previews are really helpful to find these kinds of issues. Please don't delete the live previews again in the future.
- Add a disclaimer "Live Activity Expired" when the live activity expires on the lock screen. Important improvement that has also been requested on Discord before.
- Delete unused file in the live activity folder.


FreeAPS/Sources/Services/LiveActivity/LiveActivityBridge.swift
```diff
@@ -65,35 +65,44 @@ extension LiveActivityAttributes.ContentState {
...
+        let detailedState: LiveActivityAttributes.ContentAdditionalState?
 
-        let conversionFactor: Double = settings.units == .mmolL ? 18.0 : 1.0
-        let convertedChartBG = chartBG.map { Double($0) / conversionFactor }
+        switch settings.lockScreenView {
+        case .detailed:
+            let chartBG = chart.map(\.glucose)
 
-        let chartDate = chart.map(\.date)
+            let conversionFactor: Double = settings.units == .mmolL ? 18.0 : 1.0
```
marionbarker, line 74: Just noting that other parts of Trio use:
```
let conversionFactor = 0.0555
```

Whereas this code is using (not quite) the inverse of that. It may be out of scope of this PR, but units conversion should be handled consistently.

marionbarker (approved):
Other than the conversionFactor noted, this code built on an SE (no LA available) and an iPhone 15. Testing it without using Trio.

I did not see the clipping that was supposed to be fixed either before or after applying the patch.

dnzxy:
> I did not see the clipping that was supposed to be fixed either before or after applying the patch.

If you want to reproduce clipping, simulate a 3 digit glucose with a double arrow up, that should be clipped to something like `10…` instead of `100 <arrow>` (before this fix).

marionbarker (dismissed):
Approving again after base branch was changed. (Same comments as before about conversion - but that is out of scope of this PR.)

marionbarker:
As requested. Lock screen graphics when CGM data is stale and when LA has expired. 

IMG_6899
IMG_6900

dnzxy (dismissed):
Approval based on code review. LGTM.

FreeAPS/Sources/Services/LiveActivity/LiveActivityBridge.swift
bjornoleh: Was this fixed?

mountrcg: not in this commit yet.

dnzxy: @bjornoleh it's out of scope for his PR, as also mentioned by @marionbarker in her testing comment. hence the resolution :)

bjornoleh: Ok, if the divide by 18 was not introduced here, then I agree. If it was, then I suppose it should be fixed. 

Not sure if it would matter for the end result, but there are rounding issues when comparing CGM values between other apps, hopefully we are not introducing differences in rounding within this app, as this can throw people onto the troubleshooting wagon for no reason (that’s my main concern).

dnzxy: @bjornoleh please take a peek at the git diff in the PR. This was not introduced but only moved (from line 70-71 to now inside a `switch case`). 
image

bjornoleh: Perfect, thanks, that’s all I needed to know, it was not so easily understood from looking at it on the phone (have probably missed this being pointed out before too). I’m not giving a formal review now, but I have no objections to merging.

10nas:
> As requested. Lock screen graphics when CGM data is stale and when LA has expired.

For future reference, the expired screen can be easily tested by changing this line to `true`: [LiveActivityBridge.swift:105](https://github.com/nightscout/Trio/blob/73afe846da67bd898cbdeafabd722fed0e6d4947/FreeAPS/Sources/Services/LiveActivity/LiveActivityBridge.swift#L105) The live activity will then always show the expired message (only the "simple" live activity).

10nas:
> IMG_6900

I noticed the strange background here and fixed it. No other changes in my newest commit. Also synced up with latest dev branch while at it.

dnzxy:
So this only removes the square background behind the text in the rounded shape, correct, @10nas ?

10nas:
> So this only removes the square background behind the text in the rounded shape, correct, @10nas ?

Correct

dnzxy (approved):
Re-approving after minor UI change around expired LAs.
Thanks for the update, nas!

MikePlante1:
I reproduced the clipping in Dynamic Island with an iPhone 15 simulator. The clipping occurred when the sgv was 3 digits and there was a double arrow. It only occurred when Trio was only using one Dynamic Island slot because the other was being used by a different app.

Interesting to note that none of the examples displayed 2 arrows like they should, but neither did the main screen of Trio. Two arrows were shown for both Simple and Detailed lockscreen widgets and the notifications, though.

Also interesting that the arrow is only red when using one Dynamic Island slot.

| Notes | Screenshot |
| --- | --- |
| dev 1 slot | 1 slot dev |
| dev 2 slot | 2 slot dev |
| PR 1 slot | 1 slot with PR |
| PR 2 slot | 2 slot with PR |

MikePlante1 (approved):
LGTM from testing and light code review.

Sjoerd-Bo3 (approved):
Tested in vivo and looked for the expire changes and such. They look and function great. Clipping didnt test (mmol, so never had any issue anyway)

Code review looks great!

Sjoerd-Bo3:
Merging with 3(+expired from marion) approvals

PR #320, opened 2024-06-17
Eventual BG with 1 decimal in the app and in the watch app

### LINKED ISSUE #319: Eventual BG for mmol/L is presented with irrelevant precision (2 decimal places)
## Describe the bug
Eventual BG for mmol/L is presented with irrelevant precision (2 decimal places). The eventual BG estimate is highly uncertain, and presenting this value at such an apparent level of precision might give the wrong idea about the nature of this estimate.

## Expected behavior
Eventual BG for mmol/L should at most be presented with a single decimal digit.

## Screenshots
App and watch in Xcode iOS simulator showing eventual BG of 13,76 mmol/L:
image

For mg/dL, the eventual BG value is presented as integer values, which is fine (+/- 1 mg/dL = +/- 0,06 mmol/L, or approximately +/- 0,1 mmol/L)

### Trio Version:
* Version Number: 0.1.0
* Repo: *nightscout/trio*
* Branch: alpha in this example, but applies to all previous versions
* Git Reference: 82f2d01adfb1071d4f9b4916ee9ab879827bdf3a

bjornoleh:
- Use `targetFormatter` instead of `numberFormater` for `eventualBG` in the app, since this has the appropriate `maximumFractionDigits = 1`

- Change eventualFormatter to use `maximumFractionDigits = 1` instead of `2` on the watch

Resolves #319

### DIFF
--- a/FreeAPS/Sources/Modules/Home/View/HomeRootView.swift
+++ b/FreeAPS/Sources/Modules/Home/View/HomeRootView.swift
@@ -344,7 +344,7 @@ extension Home {
 
                     if let eventualBG = state.eventualBG {
                         Text(
-                            "⇢ " + numberFormatter.string(
+                            "⇢ " + targetFormatter.string(
                                 from: (state.units == .mmolL ? eventualBG.asMmolL : Decimal(eventualBG)) as NSNumber
                             )!
                         )
--- a/FreeAPS/Sources/Services/WatchManager/WatchManager.swift
+++ b/FreeAPS/Sources/Services/WatchManager/WatchManager.swift
@@ -216,7 +216,7 @@ final class BaseWatchManager: NSObject, WatchManager, Injectable {
     private var eventualFormatter: NumberFormatter {
         let formatter = NumberFormatter()
         formatter.numberStyle = .decimal
-        formatter.maximumFractionDigits = 2
+        formatter.maximumFractionDigits = 1
         return formatter
     }
 


bjornoleh:
Eventual BG as presented with this patch:

52,8 mmol/L to the left, 951 mg/dL to the right (random numbers from sim phone/watch/CGM and pump)
image

mountrcg:
@bjornoleh there is a glucoseformatter also in ./FreeAPS/Sources/Modules/Home/View/Header/CurrentGlucoseView.swift which should fit perfectly for this issue

mountrcg:
> But I suppose my solution is a minimal and acceptable change, isn't it?

Totally!

As per discussion of targets and rounding errors leading to changed targets after multiple unit switches, I suspect that using https://github.com/nightscout/Trio/blob/20ba334b2c3b62ba296e67695bc015c4e5c868d4/FreeAPS/Sources/Modules/Home/View/Header/CurrentGlucoseView.swift#L13 would not lead to these. I think @polscm32 already centralized all formatters in Helpers at one point in some branch, which I guess is something we should also do. It only popped into my mind when viewing this PR as there are some inconsistencies in how rT.variables are handed over from oref to swift. Some are "convert_bg functioned" some always in md/dL - I think I am partly responsible for that and started fixing it, so I stumbled across glucoseformatter just this week.

bjornoleh:
@mountrcg , perhaps you can write up an issue about this for later?

Sjoerd-Bo3 (dismissed):
LGTM and tested on watch

Sjoerd-Bo3 (approved):

AndreasStokholm (approved):
Looks good and tested in sim with iOS 16.4, iOS 17.5, watchOS 10.4

PR #322, opened 2024-06-18
feat(issuemanagement): automatically adding the project to the issue types bug and needs triage

Sjoerd-Bo3:
bug and needs triage labels are automatically added the Trio project item


.github/workflows/add_to_project.yml
```diff
@@ -0,0 +1,20 @@
...
+  issues:
+    types:
+      - opened
+
+jobs:
+  add-to-project:
+    name: Add issue to project
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@RELEASE_VERSION
```
bjornoleh, line 13: You probably want to specify the release version here. Latest release is v1.0.1. Set it to v1 or v1.0 perhaps?

https://github.com/actions/add-to-project/releases

bjornoleh (changes requested):

bjornoleh:
Looks good I guess. Did you get to test this somehow? Screenshots?

.github/workflows/add_to_project.yml
```diff
@@ -0,0 +1,20 @@
...
+    name: Add issue to project
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@v1.0.2
+        with:
+          # You can target a project in a different organization
+          # to the issue
+          project-url: https://github.com/orgs/nightscout/projects/2
+          github-token: ${{ secrets.ADD_TO_PROJECT_PAT }}
+          labeled: bug, needs-triage
```
bjornoleh, line 19: I looked at the documentation for actions/add-to-project, and the current configuration will only add issues opened with labels bug OR needs-triage. Meaning enhancements are not included (not sure if we use any other labels). 

https://github.com/actions/add-to-project?tab=readme-ov-file#examples

Sjoerd-Bo3: correct, this only targets bug issues for now.
If we do the features here aswell we need a different strategy for that anyway.

bjornoleh: Ok, I guess this is fine for now. 

Also noticing use of separate GH token secrets here and in PR 323. I guess they could be the same, so the secrets doesn’t have to be entered in duplicate?

Sjoerd-Bo3: Nope, different PAT's with different rights They are already in the secret cage

bjornoleh (approved):

bjornoleh (approved):

AndreasStokholm (approved):
This works for me!

bjornoleh (approved):

bjornoleh:
Merging with two approvals

PR #323, opened 2024-06-18
Stale Issues

Sjoerd-Bo3:
Stale issues will get Stale notifications after 30 days and closed after 14 days
Stale issues that 'needs-triage' will get Stale notifications after 30 days and closed after 30 days


marionbarker:
This got used with Loop and it cleaned up the backlog, but you need an option to mark issues as longer-term items that need to stay open.

Sjoerd-Bo3:
> This got used with Loop and it cleaned up the backlog, but you need an option to mark issues as longer-term items that need to stay open.

Good idea! Will implement! Also the featurerequests will not be on GH in the future.

.github/workflows/stale_triage_issues.yml
bjornoleh, line 1: Is this empty file needed? 😀

Sjoerd-Bo3: Whoops, no ofc not. Opted to use 1 file instead of 2

bjornoleh (approved):

bjornoleh (approved):

AndreasStokholm (approved):
👏

bjornoleh:
Merging with two approvals

PR #328, opened 2024-06-20
Set Default for Dexcom Upload to Remote Service to be true

marionbarker:
There were complaints early on that the default setting for the Dexcom CGM modules (in LoopKit) for uploading to remote service was set to false.

That cannot be changed in the LoopKit GitHub username, but the same restriction does not apply in the loopandlearn username.

Modify that default to true for both the CGMBLEKit and G7SensorKit repositories in loopandlearn trio branch.

In addition, the main_rc branch of CGMBLEKit, previously used for Trio (before this PR) had an earlier version of the PR for handling future dates for G6 (which caused issues with the SAGE upload in Loop-dev). The trio branch for loopandlearn CGMBLEKit was updated to use the final form of that PR. See [LoopKit: CGMBLEKit PR 191](https://github.com/LoopKit/CGMBLEKit/pull/191).
* The difference is the final version adds [extra printout to Xcode error log](https://github.com/LoopKit/CGMBLEKit/pull/191/commits/bbdda7b1077f15a0093e2748ccf176509726c3d4)


marionbarker:
## Summary
Both G6 and G7 have upload reading enabled for fresh install

## Testing
* I deleted the Trio app from the test phone and built fresh.
* I confirmed both G6 and G7 and consequently Nightscout have upload Glucose set to true.
* I did not do any Tidepool testing.

marionbarker:
@MikePlante1 Thanks for pointing out my mistake. I missed a cherry pick to get trio back in line with loopandlearn dev_rc after creating trio from LoopKit/CGMBLEKit dev branch.

marionbarker:
Closing this because it is not as simple as changing the default to be true in the 2 submodules. In fact, iAPS still has the default in CGMBLEKit as false, but there are other changes in iAPS that tie together the Nightscout upload glucose enable flag with the cgm manager upload enable flags.

Issue #199, opened 2024-05-16
Add Referral Text for Upload Glucose Icon in Nightscout Config


Sjoerd-Bo3:
### Is your feature request related to a problem? Please describe.
Users are often confused by the grey "upload glucose" icon in Nightscout config, not knowing where to go to enable glucose uploading. This confusion could be avoided with better guidance. Currently, there's no indication in the Nightscout config on how to enable this feature, leading to potential frustration and delays in data uploading.

### Describe the solution you'd like
Add a referral text next to the grey "upload glucose" icon in the Nightscout config. The text should guide users on where to go to enable glucose uploading, such as "Please go to [specific location] to enable uploading glucose." This would improve user experience by providing clear instructions.

### Describe alternatives you've considered
- Adding a button that directly opens the CGM module instead of just referral text. *(Submodules changes needed)*
- Implementing a notification system to alert users on transmitter change that they need to re-enable glucose uploading. *(Would need a separate Issue?)*
- Exploring the possibility of defaulting the glucose upload setting to "on" for new users or after transmitter swaps. *(Submodules changes needed and discussed at other places to not do this)*

### Additional context
The need for this improvement was discussed extensively, with several users noting the inconvenience caused by the current setup. Mike Plante mentioned the repetitive need to toggle the setting after every transmitter swap, which is particularly annoying for G6 users. Dan and Sjoerd noted that while the setting is connected, it’s not changeable from the Nightscout settings. Implementing this small change would significantly enhance the user experience and reduce confusion.

![image](https://github.com/nightscout/Trio/assets/2100083/2a876e12-8854-46a4-915d-083e2cc941c6)

---

**Note:** This request aims to enhance the usability of the Nightscout config by providing clear guidance on enabling glucose uploads, ultimately making the system more user-friendly and reducing the likelihood of user errors.

Sjoerd-Bo3:
Comment from Daniel:
Hi again,

I added an "allow NS download toggle" to my personal iAPS build a couple of weeks ago. Just made a commit to a fresh Trio dev testing branch with the same changes. Feel free to use and improve this as you like (Super simple code implementation, but it works)
https://github.com/dsnallfot/Trio/commit/b8d9a3160a31a417561c602e9e38915777853f35

MikePlante1:
[...]

PR #338, opened 2024-06-29
direct user where to flip Upload Glucose toggle

MikePlante1:
Since the `Upload Glucose` toggle is disabled when using a plugin for the cgm source (Dexcom or LibreTransmitter), display text to direct the user to where they can flip the toggle. If they aren't using a plugin for their cgm source, then this text won't be displayed.

> To flip the Upload Glucose toggle, go to ⚙️ > CGM > CGM Configuration

Screenshot 2024-06-29 at 1 36 51 PM

Addresses issue https://github.com/nightscout/Trio/issues/199

### DIFF
--- a/FreeAPS/Sources/Modules/NightscoutConfig/View/NightscoutUploadView.swift
+++ b/FreeAPS/Sources/Modules/NightscoutConfig/View/NightscoutUploadView.swift
@@ -13,6 +13,10 @@ struct NightscoutUploadView: View {
                         "The Upload Treatments toggle enables uploading of carbs, temp targets, device status, preferences and settings."
                     )
                     Text("\nThe Upload Glucose toggle enables uploading of CGM readings.")
+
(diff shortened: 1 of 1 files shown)


Sjoerd-Bo3:
closes #199

marionbarker:
Should this be a request to `dev` or to `alpha`

I confirmed that this works as indicated, but it is still really confusing.
The extra message showed up in the NS Upload screen when either G6 or G7 was selected as the CGM.
When I had xDrip4iOS selected as the CGM, the NS glucose toggle worked directly in the NS Upload screen.

It's still really confusing, but I think this is a good-enough work-around for RC1.

marionbarker (approved):
Works as a bandaid fix.

Sjoerd-Bo3 (approved):
LGTM

Sjoerd-Bo3 (approved):

marionbarker (approved):

PR #348, opened 2024-07-10
Update Insulin Peak Time

tmhastings:
Correct identifications of Ultra-rapid and Rapid-acting insulins in the Insulin Peak Time Popup

### DIFF
--- a/FreeAPS/Sources/Modules/DataTable/View/DataTableRootView.swift
+++ b/FreeAPS/Sources/Modules/DataTable/View/DataTableRootView.swift
@@ -45,7 +45,7 @@ extension DataTable {
                 formatter.minimumFractionDigits = 0
                 formatter.maximumFractionDigits = 1
             }
-            formatter.roundingMode = .down
+            formatter.roundingMode = .halfUp
             return formatter
         }
 


MikePlante1:
> Does this need an update to localize strings to show up in the pop-up in the app? [Line 2034](https://github.com/nightscout/Trio/blob/dev/FreeAPS/Sources/Localizations/Main/en.lproj/Localizable.strings#L2033C1-L2034C140)

To show in the app, the change needs to be added to [PreferencesEditorStateModel.swift#L382](https://github.com/nightscout/Trio/blob/9672da256c317a314acc76d6e4f6e82cc174d133/FreeAPS/Sources/Modules/PreferencesEditor/PreferencesEditorStateModel.swift#L382) and in order for localizations to work, all 23 **Localizable.strings** files need to be updated. Otherwise all translated apps would just show the untranslated text from **PreferencesEditorStateModel.swift**

***

> Novolog is used in Loop. In some places (Canada, for one) it's referred to as Novorapid.
> 
> Not sure about the timing.

The options for insulin in Trio come from Loop and are Apidra, Humalog, Novolog, Fiasp, and Lyumjev... so I think it's best to just leave this tooltip as `Novolog`. I do think it could be beneficial to update the description in LoopKit in [InsulinType.swift](https://github.com/LoopKit/LoopKit/blob/873b3b7c406cfc982f9061afb5f5e27e88d9208d/LoopKit/InsulinKit/InsulinType.swift#L56) as well as all 11 **Localizable.strings** for LoopKit

I'd suggest something like:

> NovoLog (insulin aspart) is a rapid-acting insulin made by Novo Nordisk. It also goes by the name NovoRapid

FreeAPS/Sources/Localizations/Main/en.lproj/Localizable.strings
```diff
@@ -2031,7 +2031,7 @@ Enact a temp Basal or a temp target */
...
+"Time of maximum blood glucose lowering effect of insulin, in minutes. Beware: Oref assumes for ultra-rapid (Lyumjev & Fiasp) & rapid-acting (Novolog/Novorapid, Humalog, & Apidra) curves minimal (35 & 50 min) and maximal (100 & 120 min) applicable insulinPeakTimes. Using a custom insulinPeakTime outside these bounds will result in issues with Trio, longer loop calculations and possible red loops." = "Time of maximum blood glucose lowering effect of insulin, in minutes. Beware: Oref assumes for ultra-rapid (Lyumjev & Fiasp) & rapid-acting (Novolog/Novorapid, Humalog, & Apidra) curves minimal (35 & 50 min) and maximal (100 & 120 min) applicable insulinPeakTimes. Using a custom insulinPeakTime outside these bounds will result in issues with Trio, longer loop calculations and possible red loops.";
```
marionbarker, line 2034: I still think this is horribly confusing. There is an Insulin Type selected in Pump; Insulin Curve selected in Preferences and then this. And what the heck is bi-linear (Insulin Curve in preferences)? 

Also the exponential model (for rapid-acting and ultra-rapid) has 3 parameters - this is only changing one.

What about something like this, where you make sure the limits are right and fix the {pump/insulin curve} section:

Custom adjustment for time of maximum blood glucose lowering effect of insulin, in minutes. Most people should use the insulin model as selected in {the pump or preferences: Insulin curve}. If custom adjustment is desired, limit the range to be no shorter than 35 minutes and no longer than 120 minutes.

 Beware: Oref assumes for ultra-rapid (Lyumjev & Fiasp) & rapid-acting (Novolog/Novorapid, Humalog, & Apidra) curves minimal (35 & 50 min) and maximal (100 & 120 min) applicable insulinPeakTimes. Using a custom insulinPeakTime outside these bounds will result in issues with Trio, longer loop calculations and possible red loops." = "Time of maximum blood glucose lowering effect of insulin, in minutes. Beware: Oref assumes for ultra-rapid (Lyumjev & Fiasp) & rapid-acting (Novolog/Novorapid, Humalog, & Apidra) curves minimal (35 & 50 min) and maximal (100 & 120 min) applicable insulinPeakTimes. Using a custom insulinPeakTime outside these bounds will result in issues with Trio, longer loop calculations and possible red loops.

dnzxy: To answer this in short: the insulin curve setting can be deleted, it’s useless. Insulin is selected via pump manager UI from LoopKit.

The text reads good!

mountrcg:
Btw, if we could apply the boundary concept to this it would help a lot. I had raised an [issue with oref](https://github.com/openaps/oref0/issues/1436) at the time. However I guess that needs splitting peak time into 3 peak times for the three insulin models.

PR #352, opened 2024-07-17
Prevent crash when holding backspace

MikePlante1:
Currently, entering text into `Note` while entering a meal and then holding the backspace button would crash the app. This PR prevents that crash.

### DIFF
--- a/FreeAPS/Sources/Views/TextFieldWithToolBar.swift
+++ b/FreeAPS/Sources/Views/TextFieldWithToolBar.swift
@@ -290,20 +290,24 @@ extension TextFieldWithToolBarString.Coordinator: UITextFieldDelegate {
         shouldChangeCharactersIn range: NSRange,
         replacementString string: String
     ) -> Bool {
-        if let maxLength = parent.maxLength {
-            // Get the current text, including the proposed change
-            let currentText = textField.text ?? ""
-            let newLength = currentText.count + string.count - range.length
-            if newLength > maxLength {
-                return false
-            }
+        guard let currentText = textField.text as NSString? else {
+            return false
         }
 
+        // Calculate the new text length
+        let newLength = currentText.length + string.count - range.length
+
+        // If there's a maxLength, ensure the new length is within the limit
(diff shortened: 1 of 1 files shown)


MikePlante1:
Force updated to use `decimalFormatter.groupingSeparator` instead of just `","` to allow it to work for all regions.

MikePlante1:
Updated to use decimalFormatter.groupingSeparator instead of just "," to allow it to work for all regions.

marionbarker (dismissed):
## Summary
LGTM to fix the crash

## Code Review
* code changes looks appropriate
* Can someone check localization?
   * I tried different languages and did not see the decimal/thousands place delimiters I expected
   * If confirmed, that should be a separate issue.  

## Test Details

### Confirm problem
* build with dev, 8303f0a8
* confirm 1 - crashed the app with backspace in Notes field
* confirm 2 - entered in carbs field - see the bug

### Confirm fix
* apply patch from PR 352
* confirm fix 1 - able to backspace in notes field, app does not crash
* confirm fix 2 - entered `126555` in carb field, displays 126,555 (with > max warning symbol)
    * delete one character at a time
        * 12,655
        * 1,265
        * 126
        * 12 (no longer > max carbs, symbol gone)
* confirm maxLength works
* confirm clear works
* confirm proper thousands' place behavior in temp target, bolus

MikePlante1:
> * Can someone check localization?
>   
>   * I tried different languages and did not see the decimal/thousands place delimiters I expected
>   * If confirmed, that should be a separate issue.

You have to change the Region of the phone, not the Language. Each of the following produces a different grouping separator:
US 9,999,999,999.5
UK 9.999.999.999,5
Norway: 9 999 999 999,5
India: 9,99,99,99,999.5

MikePlante1:
I removed the grouping separator fix from this PR, as a similar solution has been added to https://github.com/nightscout/Trio/pull/362

dnzxy (approved):

marionbarker (approved):
Tested again with the latest commit.

Issue #269, opened 2024-06-04
[Bug] SAGE not updating in NightScout


TymHuckin:
![image](https://github.com/nightscout/Trio/assets/92105120/75f20aa3-a914-4221-8d2c-e6030ac178b9)

I've noticed that the sensor is doesn't always update in Nightscout. I change the sensor yesterday (G6/Anubis) and it's still showing as over 13 days old. Had this before but then suddenly it updated.
0.1.0  56a3f87 updated on my phone 4 June 2024.

aug0211:
Noticed this with G7 as well.

Let’s all observe closely and gather data on our next sensor changes to provide details and logs so we can help get to the bottom of this.

aug0211:
Observed again today.

1. Existing G7 had issues and stopped getting readings
2. Inserted a new sensor for a “background warmup” while hoping the old sensor would come back online
3. At 25m into “background warmup”, old sensor had still not come online - decided to cut over to new sensor
4. Used the “replace sensor” option in G7 app
5. Scanned QR code in G7 app to pair new sensor
6. Confirmed in G7 app that sensor was already applied
7. “Forgot” old sensor in iOS Bluetooth settings
8. G7 app picked up new sensor
9. Trio began receiving new readings
10. Observed that SAGE in Nightscout was not updated 
11. Observed that no “CGM Sensor Start” (or similar) event had been created in Nightscout by Trio
12. Manually (via Shortcut) added “CGM Sensor Start” event to Nightscout via iOS shortcut
13. Observed SAGE updating in Nightscout after manually running the shortcut

bjornoleh:
The sensor details are stored in a file on your phone, probably in the `Trio/monitor`directory. Could you please look that up and post it here?

aug0211:
Definitely, will do when kiddo is home and I can look at his phone - most likely this evening.

bjornoleh:
For some background, SAGE uploads to Nightscout was implemented by merging this branch:

https://github.com/trixing/freeaps/commits/patch/sage/

After some refactoring, SAGE was later fixed for G6 by this PR:

https://github.com/Artificial-Pancreas/iAPS/pull/212

Then apparently SAGE was broken by a refactor after the transition to submodules in Trio, where relevant files were removed:

https://github.com/nightscout/Trio/commit/47fa01786a7457b30e2307a11e02908719fa0aa9

[...]

PR #355, opened 2024-07-19
Upload SAGE to NS for Dexcom

MikePlante1:
Address Issue https://github.com/nightscout/Trio/issues/269 when using Dexcom G5, G6 (ONE), or G7 (ONE+) as a CGM source. Note this does not address using xDrip4iOS as a cgm source.

It expands on (and slightly refactors) the LibreTransmitter fix in https://github.com/nightscout/Trio/pull/112 . Please reference the comments on that PR about adding the LibreTransmitter library (this PR adds the CGMBLEKit and G7SensorKit libraries).

The only testing I've done for this PR is with my in-vivo phone.
* iPhone 15 ProMax
* iOS 17.5.1
* Dexcom G6 via official G6 app
* Nightscout hosted on GoogleCloud
* Medtronic 722 + OrangeLink

```diff
- Testing still needs to be done using G7
```

Before building this PR, my phone was at the tip of dev: https://github.com/nightscout/Trio/commit/9672da256c317a314acc76d6e4f6e82cc174d133
* I checked `Files > Trio > Monitor` and there was no file `cgm-state.json`
* On my NS, I generated a 6-month report for `Treatments` and checked ☑️ `Event Type: CGM Sensor Start`
* The most recent entries for `CGM Sensor Start` are 2024-04-01 and 2024-03-22 but those were from when I used xDrip4iOS and directly uploaded glucose to NS from xDrip4iOS instead of Trio/iAPS.
prePR NS record xD
* I did have an entry on 2024-03-07 marked as from iAPS, but if I remember correctly, it was from this repo before it was rebranded as Trio and before the change to submodules here https://github.com/nightscout/Trio/commit/47fa01786a7457b30e2307a11e02908719fa0aa9
prePR NS record

After receiving my first reading from G6 after building this PR to my phone:
* SAGE on Nightscout was updated correctly.
postPR NS
* CGM Sensor Start showed up in Treatment report on the correct day
postPR NS record
* FIle `cgm-state.json` now exists in `Files > Trio > Monitor`
IMG_5114

### DIFF
--- a/FreeAPS/Sources/APS/CGM/PluginSource.swift
+++ b/FreeAPS/Sources/APS/CGM/PluginSource.swift
@@ -1,5 +1,7 @@
+import CGMBLEKit
 import Combine
 import Foundation
+import G7SensorKit
 import LibreTransmitter
 import LoopKit
 import LoopKitUI
@@ -162,11 +164,29 @@ extension PluginSource: CGMManagerDelegate {
         case let .newData(values):
 
             var sensorActivatedAt: Date?
+            var sensorStartDate: Date?
             var sensorTransmitterID: String?
-            /// specific for Libre transmitter and send SAGE
+
+            /// SAGE
             if let cgmTransmitterManager = cgmManager as? LibreTransmitterManagerV3 {
-                sensorActivatedAt = cgmTransmitterManager.sensorInfoObservable.activatedAt
-                sensorTransmitterID = cgmTransmitterManager.sensorInfoObservable.sensorSerial
+                let sensorInfo = cgmTransmitterManager.sensorInfoObservable
+                sensorActivatedAt = sensorInfo.activatedAt
+                sensorStartDate = sensorInfo.activatedAt
+                sensorTransmitterID = sensorInfo.sensorSerial
+            } else if let cgmTransmitterManager = cgmManager as? G5CGMManager {
+                let latestReading = cgmTransmitterManager.latestReading
+                sensorActivatedAt = latestReading?.activationDate
+                sensorStartDate = latestReading?.sessionStartDate
+                sensorTransmitterID = latestReading?.transmitterID
+            } else if let cgmTransmitterManager = cgmManager as? G6CGMManager {
+                let latestReading = cgmTransmitterManager.latestReading
+                sensorActivatedAt = latestReading?.activationDate
+                sensorStartDate = latestReading?.sessionStartDate
+                sensorTransmitterID = latestReading?.transmitterID
+            } else if let cgmTransmitterManager = cgmManager as? G7CGMManager {
+                sensorActivatedAt = cgmTransmitterManager.sensorActivatedAt
+                sensorStartDate = cgmTransmitterManager.sensorActivatedAt
+                sensorTransmitterID = cgmTransmitterManager.sensorName
             }
 
             let bloodGlucose = values.compactMap { newGlucoseSample -> BloodGlucose? in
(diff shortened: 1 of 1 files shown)


marionbarker:
## Status
Everything I could test seemed fine.
I will add an approval knowing that someone else will be checking this.

## Code review
Code review looks good - although I wonder if an enum plus case statements is more in keeping with the style of the app.
* Others need to answer that

## Test
* started with latest dev (no other patches applied), commit 8303f0a8
* applied patch from this PR
* Built to test phone with no issues
* Built to real phone on which I use Trio for LiveActivities triggering off my G7 (I run Loop as my OS-AID) 
   * Trio missed the most recent glucose - probably because I was building, so wait a few more cycles
   * Trio captured the next few CGM readings - no more misses, showing up in dynamic island and lock screen widget

marionbarker (approved):

MikePlante1:
> Code review looks good - although I wonder if an enum plus case statements is more in keeping with the style of the app.

It looks like the enum for which cgm is being used was just changed to `.plugin` for all of these, so I just copied the code for Libre. If there's a better way I'm definitely open to hearing it.

> ## Test
> * started with latest dev (no other patches applied), commit [8303f0a](https://github.com/nightscout/Trio/commit/8303f0a881fab13690522ea02dba9944b6528f6e)
> * applied patch from this PR
> * Built to test phone with no issues
> * Built to real phone on which I use Trio for LiveActivities triggering off my G7 (I run Loop as my OS-AID)
>   
>   * Trio missed the most recent glucose - probably because I was building, so wait a few more cycles
>   * Trio captured the next few CGM readings - no more misses, showing up in dynamic island and lock screen widget

Is your Trio app connected to a Nightscout account to see if it updates SAGE properly for your G7? One whose SAGE was either out of date or non-existent? I can give you credentials for a spare NS if you need.

aug0211:
Is starting a new CGM required in order to test this, can we test this with an existing/already started CGM?

I suppose we may be able to just forget/remove the existing G7 and re-add in Trio after applying this commit?

MikePlante1:
> Is starting a new CGM required in order to test this, can we test this with an existing/already started CGM?
> 
> I suppose we may be able to just forget/remove the existing G7 and re-add in Trio after applying this commit?

Is SAGE currently out-of-date in your NS? If yes, apply this PR and after receiving 1 or 2 readings from G7, check NS to see if SAGE is now updated properly.

avouspierre:
Even the code introduced specific lib in main code, currently no other simple solution to collect sage information. Probably could be a very good solution to have a generic API / method in loop submodules to retrieve this info and avoid specific code.

aug0211:
> > Is starting a new CGM required in order to test this, can we test this with an existing/already started CGM?
> > I suppose we may be able to just forget/remove the existing G7 and re-add in Trio after applying this commit?
> 
> (Edit: existing G7 is fine)
> 
> Is SAGE currently out-of-date in your NS? If yes, apply this PR and after receiving 1 or 2 readings from G7, check NS to see if SAGE is now updated properly.

Not currently out of date because I use a shortcut to set it.

It sounds like this can be tested with an exciting sensor though, just by clearing the previous sensor date from the backend in NS.

Thanks, will take a look!

aug0211 (approved):
- Tested with G7 in vivo
- Success
- Deleted previous sensor change date from Nightscout
- Confirmed latest sensor age from Nightscout was gone
- Deployed this code over my live instance of Trio
- After 1 loop, nothing updated
- After 2nd loop, the SAGE updated correctly
- I now also see the sensor insertion date in Nightscout history, noting the sensor bluetooth ID which is nice

Old manual sensor start entry:
01 - Previous Manual Entry
==========
Delete the old entry:
02 - Delete
==========
Confirm the old entry is gone:
03 - Gone
==========
Viola, the new G7 start entry automatically appears:
04 - Back Successfully
==========

dnzxy:
Merging with 3 approvals. Thank you @MikePlante1

Issue #339, opened 2024-07-01
Unable to log glucose entry > 14mmol/L


dsut4392:
## Describe the bug
When logging finger stick glucose entries with the app set to use mmol/L as units, values greater than 14mmol/L cannot be saved. 

## Attach a Log
*Tap the Trio settings icon at the bottom of the screen, then tap 'Share logs' on the bottom of the list 

## To Reproduce
*Steps to reproduce the behavior:*
1. Set units preference to mmol/L
2. *Click on on centre of chart, then Glucose tab
3. Click on “Log Glucose +”
4. Enter any value greater than “14” and the “Save” button is greyed out, the entry cannot be saved 

## Expected behavior
* values greater than 14mmol/L should be able to be saved. 

## Screenshots
*If applicable, add screenshots to help explain your problem.*
![Uploading image.png…]()

## Setup Information (please complete the following information):

Dev aa26049

### Smartphone:
* Hardware: iPhone 15 Pro Max
* OS Version: iOS 17.5.1

### Pump:

* Model:  Omnipod Dash

### CGM:
* Device: Dexcom G6
* Manager app: Dexcom App

### Trio Version:
* Version Number: Dev 0.1.0
* Repo: *nightscout/trio*
* Git Reference: aa26049

## Technical Details
*If applicable, provide any technical details that might help in diagnosing the problem. This could include logs, error messages, or relevant configuration details.*

## Additional context
*switching app to mg/dL allows entry of BG values up to 720. Should this limit be lowered to reduce safety risks? Which BG meters even report a value this high?

marionbarker:
I confirmed this issue. 

I used the same SHA (which is the most recent dev version). 
The dev code is identical to the most recent alpha branch. (The SHA is different, but code is the same).

MikePlante1:
I found the offending line in [DataTableRootView.swift](https://github.com/nightscout/Trio/blob/a364b62c7d1cd94397be1a3ad9e9e2e3e0971d48/FreeAPS/Sources/Modules/DataTable/View/DataTableRootView.swift#L157):

Looks like post https://github.com/nightscout/Trio/pull/140 : `0.8-14 mmol/L` and `40-720 mg/dL`
Pre-PR140 it was `2.2-21 mmol/L` & `40-380 mg/dL`

If we base it on the range of blood sugar meters, the popular Contour Next reads `20-600 mg/dL` & `0.6-33.3 mmol/L`. Strange that it can read significantly lower in mmol/L than mg/dL.

[...]

PR #356, opened 2024-07-19
Fix manual glucose entry for mmol/L

MikePlante1:
* Prevents mmol/L entry from being forced into having 1 decimal place
* Fixes typo that swapped high/low limits from mg/dL and mmol/L
* Round down for entry instead of to nearest

Addresses Issue https://github.com/nightscout/Trio/issues/339

Before this PR:

https://github.com/user-attachments/assets/f0651253-9f78-4a15-8cb7-9717c3245d56

After this PR:

https://github.com/user-attachments/assets/371ad5b7-8de6-465d-a690-18d9953cfedb

### DIFF
--- a/FreeAPS/Sources/Modules/DataTable/View/DataTableRootView.swift
+++ b/FreeAPS/Sources/Modules/DataTable/View/DataTableRootView.swift
@@ -37,6 +37,18 @@ extension DataTable {
             return formatter
         }
 
+        private var glucoseEntryFormatter: NumberFormatter {
+            let formatter = NumberFormatter()
+            formatter.numberStyle = .decimal
+            formatter.maximumFractionDigits = 0
+            if state.units == .mmolL {
+                formatter.minimumFractionDigits = 0
+                formatter.maximumFractionDigits = 1
+            }
+            formatter.roundingMode = .down
+            return formatter
+        }
+
         private var dateFormatter: DateFormatter {
             let formatter = DateFormatter()
             formatter.timeStyle = .short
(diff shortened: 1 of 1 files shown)


marionbarker (approved):
## Status
LGTM to fix the mmol/L Glucose entry problem
Then we can resolve whether there is a better method for fixing this with more consistent units handling

## Code Review
* code changes looks appropriate
* Can someone check localization?
   * I tried different languages and did not see the decimal/thousands place delimiters I expected
   * If confirmed, that should be a separate issue.  

## Test

### Confirm problem
* build with dev, 8303f0a8
* confirm cannot enter mmol/L values > 9.9 on glucose screen

### Confirm fix
* apply patch from PR 356
* confirm can now enter mmol/L values up to 40 mmol/L on glucose screen
* no error message is give when > 40 mmol/L is attempted, but the save button is inactive so that's ok for now
* It is a little disconcerting to enter 40.0 or 10.0 and have it modify the display to 40 or 10 respectively
    * The display on the log includes the single decimal point for mmol/L
* Switch to mg/dL - confirm min/max range is now 14 mg/dL to 720 mg/dL
* Switch to mmol/L to test the low range (log changes to 0.8 mmol)
    * confirm min/max range is 0.8 to 40.0 mmol/L

MikePlante1:
> * Can someone check localization?
>   
>   * I tried different languages and did not see the decimal/thousands place delimiters I expected
>   * If confirmed, that should be a separate issue.



> * It is a little disconcerting to enter 40.0 or 10.0 and have it modify the display to 40 or 10 respectively
>   
>   * The display on the log includes the single decimal point for mmol/L

I agree that it's not great, but unfortunately having it auto-format to a single decimal place is what messed up being able to enter double digits. I played around with it a bunch to make it so minimumFractionDigits defaults to 0, is set to 1 when the user types a decimal point, and is reset to 0 if the decimal point is deleted... but couldn't get it to work right.

dnzxy (approved):
LGTM

PR #359, opened 2024-07-22
set default state and image for simulatorPump

kskandis:
This fixes an issue with "Add Pump" not displaying pump after Simulator Pump is added per @dnzxy.

### DIFF
--- a/FreeAPS/Sources/APS/DeviceDataManager.swift
+++ b/FreeAPS/Sources/APS/DeviceDataManager.swift
@@ -94,6 +94,9 @@ final class BaseDeviceDataManager: DeviceDataManager, Injectable {
                     }
                     pumpExpiresAtDate.send(endTime)
                 }
+                if let simulatorPump = pumpManager as? MockPumpManager {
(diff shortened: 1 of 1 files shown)


MikePlante1:
Both with and without this PR applied, I notice with a fresh install on a simulator, adding a simulator pump still shows the "add pump" icon after selecting it. Sending a bolus changes the "add pump" icon to a "100% battery" icon, though, both with and without this PR. Removing the pump instantly changes the icon to "add pump", and if I then select "simulator" again for pump it shows the "100% battery" icon as soon as it's selected without having to issue another bolus.

Am I testing the wrong thing? What steps should I take?

kskandis:
> Am I testing the wrong thing? What steps should I take?

No, you are correct.  There is still an issue with this PR that needs further investigation.  I'll convert this PR to a draft 'til we isolate the additional issues.

dnzxy (approved):
Can confirm that `40f9b33` fixes the issue. I had that locally as well, just a bit more down in `DeviceDataManager` and not as part of the `didSet`. Thanks to @kskandis for fixing this 🙏 

LGTM!

MikePlante1 (approved):
Repeated the tests with the latest commit and everything LGTM.

PR #360, opened 2024-07-22
Remove feature to save autotune basal profile to pump

MikePlante1:
Based on [this issue report](https://github.com/Artificial-Pancreas/iAPS/issues/711) on iAPS, this feature is buggy and potentially dangerous anyway.


aug0211:
Is this a proposed bandaid while a true fix for the issue is worked on, or are you proposing killing this altogether?

jdtoppin:
Is this reproducible in trio? I used this feature somewhat often when I first started iAPS and it was nice to save time - I imagine that was the intent when it was introduced.

PR #365, opened 2024-07-27
Don't allow entry when entry is already at maximumFractionDigits

MikePlante1:
This makes it so there isn't any weird rounding when entering more decimal places than allowed. Now it just prevents entering more decimal places once the maximum has been entered.

### DIFF
--- a/FreeAPS/Sources/Views/TextFieldWithToolBar.swift
+++ b/FreeAPS/Sources/Views/TextFieldWithToolBar.swift
@@ -135,6 +135,26 @@ public struct TextFieldWithToolBar: UIViewRepresentable {
                 textField.moveCursorToEnd()
             }
         }
+
+        // Helper method to calculate the number of decimal places in a string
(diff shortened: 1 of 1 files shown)


marionbarker:
## Summary
This fixes the "weird rounding"

## Configuration
* Build Trio dev commit c870190d to SE phone running iOS 17.5.1
* Find a "weird rounding" example.
* Then apply the patch and test.

## Tests

### Example:
Type a value with multiple decimal places into
* bolus entry with straight dev:
   *  type 1.555555 and it provides: 1.56, 1.57, 1.58, 1.59, 1.6 etc
* bolus entry with patch
   *  type 1.555555 and it provides: 1.55 and that's it

marionbarker (approved):

dnzxy (approved):
Approving on code review.  +1 for using nice, descriptive and readable variable names. 

Thanks Mike!

bjornoleh:
Merging with two approvals.

PR #366, opened 2024-07-27
update submodules OmniXXX: prevent bogus pod suspended message 

marionbarker:
This fixes an intermittent error where user is incorrectly told their pod is suspended.
It has been reported several times for Loop and a Loop Report file was captured one time.
This enabled analysis that this spurious message occurs when attempting to bolus following a comms error.
This change was determined to be correct from code inspection.

Please see: [Bug: Incorrect pod suspended message when bolusing after comms error](https://github.com/LoopKit/Loop/issues/2198#top)


dnzxy (approved):

aug0211 (approved):

dnzxy:
Merging with 2 approvals.

Issue #371, opened 2024-08-03
Excessive Log Size in rc1


marionbarker:
## Describe the bug
It is known that the log sizes are fairly big - in the range of 5 to 10 MB per day.
* for the case where Apple Health is enabled and then Dynamic Features are enabled, the log is more than doubled in size

## Attach a Log
*Tap the Trio settings icon at the bottom of the screen, then tap 'Share logs' on the bottom of the list and attach it to this ticket.*
* The log attached below did not have Apple Health enabled until 05:18 am.
* Apple Health was later disabled at 22:00
* The file size was 21.6 MB before it was compressed.

## To Reproduce
*Steps to reproduce the behavior:*
1. Run Trio
2. Enable Apple Health
3. Turn on Dynamic Features
4. Share logs

## Expected behavior
These `HealthKitManager`debug statements were added as part of trouble shooting and do not need to be present in rc1

## Configuration:
* Trio dev branch, commit 08eba438 (tag: 0.2.0-rc1)
* SE phone running Nightscout as CGM and rPi DASH as pump

## Log:
[log_prev.txt.zip](https://github.com/user-attachments/files/16481070/log_prev.txt.zip)

marionbarker:
Closed because PR #372 was merged

PR #372, opened 2024-08-03
Decrease HealthKitManager debug logging

marionbarker:
* disable some debug logging when AppleHealth is enabled

See Issue #371 for explanation for this PR.

## Additional comments

* most of the excessive lines came in as debug logging in PR #22
* other debug log lines have been removed over time in various commits
* these particular lines were identified by @itsmojo as being unnecessary when he analyzed results

## Testing

* I confirmed the extra lines appeared when Apple Health and Dynamic Features are enabled
* I tested the patch supplied by @itsmojo and confirm the excessive debug logging (for this case) is removed

### DIFF
--- a/FreeAPS/Sources/APS/DeviceDataManager.swift
+++ b/FreeAPS/Sources/APS/DeviceDataManager.swift
@@ -459,7 +459,6 @@ extension BaseDeviceDataManager: PumpManagerDelegate {
         completion: @escaping (_ error: Error?) -> Void
     ) {
         dispatchPrecondition(condition: .onQueue(processQueue))
-        // debug(.deviceManager, "New pump events:\n\(events.map(\.title).joined(separator: "\n"))")
 
         // filter buggy TBRs > maxBasal from MDT
         let events = events.filter {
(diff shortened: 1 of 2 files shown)


AndreasStokholm:
Seeing as PR 22 is merged quite a while in the past and it seems to be stable. Maybe the debugging I added can be removed completely instead of just commenting it out.

marionbarker:
If I remove the code with comments, then, as an example, I get this:
```
            healthKitStore.save(bolusSamples + basalSamples) { (success: Bool, error: Error?) -> Void in
                if success {
                } else {
                    debug(.service, "Failed to store insulin entry in HealthKit Store!")
                    debug(.service, error?.localizedDescription ?? "Unknown error")
                }
            }
```

To my novice eye - that seems like it needs a different rewrite - happy to accept direction and I'll fix all three places where this would occur.

marionbarker:
Ready for review again. I removed the debug lines instead of commenting them out.

itsmojo (approved):

aug0211 (approved):

MikePlante1 (approved):
I agree those logs aren't needed and the PR only removes `debug()` lines. Successful build to a simulator to confirm lint-free.

Issue #373, opened 2024-08-04
The current version of build_trio.yml will not work in main branch


marionbarker:
## Describe the bug
Following the release of main (in the future), Trio build script will have issues similar to that seen following the release of Loop 3.4. This required the following LoopWorkspace [PR 164](https://github.com/LoopKit/LoopWorkspace/pull/164).

## Additional context
Modify the build script for Trio before the creation of the main branch associated with Trio release.
This will avoid this issue and save everyone a lot of time.

As of commit 0efde61f of the dev branch, the version of build_trio.yml will fail when found in the main branch.

PR #374, opened 2024-08-04
Fix browser build sync and alive behavior

marionbarker:
This PR is in response to Trio Issue #373. In addition to fixing the main build for GitHub actions, this also brings in these other changes:
* update macos up to version 14
* explicitly set Xcode version to 15.4
* modify `# Patch Fastlane Match to not print tables` to not print user information in the clear
   * this was identified as a privacy issue in LoopWorkspace (happened in macos-14, not seen previously when using macos-13)  

This PR uses changes to LoopWorkspace [PR 164](https://github.com/LoopKit/LoopWorkspace/pull/164) following release of Loop version 3.4.0.

The timing is such that the main branch must exist and this PR be applied at the same time.

## Status:
* The Browser Build for Trio will not run using this PR until the main branch actually exists

## Test
* In order to test, I [modified](https://github.com/loopdocs-tester/Trio/commit/507d42908989760f4b24748c1da907f8451bfa3f) build_trio.yml in the test fork to use the dev branch for both alive-main and alive-dev
* The [build action](https://github.com/loopdocs-tester/Trio/actions/runs/10237530672) was successful with this modification.

## Code Review:
I compared the *.yml files found in LoopWorkspace (main, aka 3.4.1)
These 3 are identical:
* validate_secrets.yml
* add_identifiers.yml
* create_certs.yml

These files differ only in references to LoopWorkspace vs Trio and LoopWorkspace embeds a section on customizations that it not found in Trio:
* build_loop.yml vs build_trio.yml

### DIFF
--- a/.github/workflows/add_identifiers.yml
+++ b/.github/workflows/add_identifiers.yml
@@ -8,27 +8,39 @@ jobs:
     name: Validate
     uses: ./.github/workflows/validate_secrets.yml
     secrets: inherit
-
+  
   identifiers:
+    name: Add Identifiers
     needs: validate
-    runs-on: macos-13
+    runs-on: macos-14
     steps:
-      # Uncomment to manually select Xcode version if needed
-      #- name: Select Xcode version
-      #  run: "sudo xcode-select --switch /Applications/Xcode_15.0.1.app/Contents/Developer"
+      # Uncomment to manually select latest Xcode if needed
+      #- name: Select Latest Xcode
+      #  run: "sudo xcode-select --switch /Applications/Xcode_13.0.app/Contents/Developer"
       
       # Checks-out the repo
       - name: Checkout Repo
         uses: actions/checkout@v4
-        
+      
       # Patch Fastlane Match to not print tables
       - name: Patch Match Tables
-        run: find /usr/local/lib/ruby/gems -name table_printer.rb | xargs sed -i "" "/puts(Terminal::Table.new(params))/d"
-        
+        run: |
+          TABLE_PRINTER_PATH=$(ruby -e 'puts Gem::Specification.find_by_name("fastlane").gem_dir')/match/lib/match/table_printer.rb
+          if [ -f "$TABLE_PRINTER_PATH" ]; then
+            sed -i "" "/puts(Terminal::Table.new(params))/d" "$TABLE_PRINTER_PATH"
+          else
+            echo "table_printer.rb not found"
+            exit 1
+          fi
+
       # Install project dependencies
       - name: Install Project Dependencies
         run: bundle install
 
+      # Sync the GitHub runner clock with the Windows time server (workaround as suggested in https://github.com/actions/runner/issues/2996)
+      - name: Sync clock
+        run: sudo sntp -sS time.windows.com
+      
       # Create or update identifiers for app
       - name: Fastlane Provision
         run: bundle exec fastlane identifiers
--- a/.github/workflows/build_trio.yml
+++ b/.github/workflows/build_trio.yml
@@ -2,20 +2,20 @@ name: 4. Build Trio
 run-name: Build Trio (${{ github.ref_name }})
 on:
   workflow_dispatch:
-  
+
   ## Remove the "#" sign from the beginning of the line below to get automated builds on push (code changes in your repository)
   #push:
-  
+
   schedule:
-    #- cron: '30 04 1 * *' # Runs at 04:30 UTC on the 1st every month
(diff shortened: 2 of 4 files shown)


marionbarker:
In addition to the files listed above, comparison was also made between LoopWorkspace 3.4.1 for these files:
* Gemfile
* Gemfile.lock
* fastlane/Fastfile

The first two files are identical. The Fastfile has the suitable changes between Loop and Trio along with some whitespace changes. The Trio version configures and uses the BUNDLE_ID which simplifies some lines.

.github/workflows/build_trio.yml
```diff
@@ -33,151 +33,172 @@ jobs:
...
+          (vars.SCHEDULED_BUILD != 'false' || vars.SCHEDULED_SYNC != 'false')
+        uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.GH_PAT }}
+          ref: ${{ steps.check_branch.outputs.ALIVE_BRANCH }}
+
+      - name: Sync upstream changes
+        if: | # do not run the upstream sync action on the upstream repository
+          needs.check_alive_and_permissions.outputs.WORKFLOW_PERMISSION == 'true' &&
+          vars.SCHEDULED_SYNC != 'false' && github.repository_owner != 'LoopKit' && steps.check_branch.outputs.ABORT_SYNC == 'false'
```
bjornoleh, line 138: Looks like some LoopKit details followed by accident. It should read nightscout here, perhaps similar changes are present too.

.github/workflows/add_identifiers.yml
```diff
@@ -8,27 +8,39 @@ jobs:
...
+          else
+            echo "table_printer.rb not found"
+            exit 1
+          fi
+
       # Install project dependencies
       - name: Install Project Dependencies
         run: bundle install
 
+      # Sync the GitHub runner clock with the Windows time server (workaround as suggested in https://github.com/actions/runner/issues/2996)
```
bjornoleh, line 40: This part is no longer needed, although it does no harm (it only takes 1 seconds to run this step). It was a bandaid while things were broken on GitHub servers.

marionbarker: I put it in because it is still in the LoopWorkspace version.
I have no strong opinions either way. Let me know if you want it removed.

bjornoleh: I think this can be up to whomever volunteers for the maintenance of these files across the projects. If it is easier to keep them as identical as possible, please feel free to add this.

dnzxy (approved):

marionbarker:
## Status
Success
Caveat - might still have some users with alive branch failures - if so, they just need to delete any branch starting with `alive` and then run again.

## Testing
This can be tested now that there is a `main` branch at nightscout/Trio
* Before - it would fail because SHA_MAIN was null

### Test Overview:
* Use loopdocs-tester GitHub username.
* Need to have the branch name be called dev for the test.
* Do the merge of patch from this PR into dev for just loopdocs-tester.

### Test dev branch with PR 374 applied
* Success: [build Trio (dev)](https://github.com/loopdocs-tester/Trio/actions/runs/10389901533) with this patch (so one commit ahead of nightscout:dev)
   * `alive-dev` and `alive-main` created

### Test main branch
* add new branch using nightscout:main
* this is identical to `dev` branch but without the PR applied
   * there is no alive branch
* Success: [build Trio (main)](https://github.com/loopdocs-tester/Trio/actions/runs/10390492685)
   * `alive` branch created

MikePlante1 (approved):
Pulled this PR in and successfully built, which correctly created two new branches: `alive-main` and `alive-dev`.

.github/workflows/add_identifiers.yml
```diff
@@ -8,27 +8,39 @@ jobs:
     name: Validate
     uses: ./.github/workflows/validate_secrets.yml
     secrets: inherit
-
+
```
MikePlante1, line 11: Whitespace added.

marionbarker: Sorry about that. I made it identical to LoopWorkspace (which has that extra whitespace.) (validate_secrets, add_identifiers and create_certs match exactly. build_trio/build_loop have the expected differences.)

I propose we leave it for now because we already have 2 approvals and whitespace doesn't matter. If there are future changes to this file, I expect the whitespace get fixed at that time.

If @dnzxy disagrees - I'll change it.

MikePlante1: In that case just leave it as is.

Issue #267, opened 2024-06-03
[bug] - Rounding of mmol/L manual BG values


bjornoleh:
## Describe the bug

The manual BG values 8,0 and 8,1 are both displayed as 8,0 in the list of BG values. 

Test of manual BG values (mmol/L) :
8,2  (stored as 147 mg/dL)
8,11  (stored as 146 mg/dL)
8,1  (stored as 145 mg/dL)
8  (stored as 144 mg/dL)

Displayed as follows in the app:

<img src="https://github.com/nightscout/Trio/assets/63544115/566ca73a-3e2b-4e56-8a81-c60511f9f32f" width="300">

Below is the json of the above manual BG values:

```

  {
    "type" : "Manual",
    "date" : 1717050971595.7067776,
    "dateString" : "2024-05-30T06:36:11.596Z",
    "glucose" : 147,
    "_id" : "5621BF52-49D5-4C17-B86D-A9826F33FA77"
  },
  {
    "glucose" : 146,
    "type" : "Manual",
    "_id" : "99655471-7A3F-41BB-BBBB-9694A28CFB32",
    "date" : 1717050965423.120384,
    "dateString" : "2024-05-30T06:36:05.423Z"
  },
  {
    "type" : "Manual",
    "date" : 1717050955886.6382848,
    "_id" : "15FBE50B-EBFA-4F9B-BE0A-C48E2B71C158",
    "glucose" : 145,
    "dateString" : "2024-05-30T06:35:55.887Z"
  },
  {
    "glucose" : 144,
    "date" : 1717050950799.5430912,
    "_id" : "67E4B93F-F25A-42BA-BF1D-9EC4555C8D0D",
    "dateString" : "2024-05-30T06:35:50.800Z",
    "type" : "Manual"
  },
```

### Trio Version:
* Repo: *nightscout/trio*, dev branch 
* Git Reference: 14c7643275e3f351448d7623432f0d1778b65af4

## Additional context

The BG values are not saved as mmol, but are converted to mgdL first.

It turns out 8,1 is stored as 145 mg/dL, although the correct conversion is 8,1 / 0,055 = 145,9459

Then the displayed value after converting back to mmol is no longer the same as the entered value of 8,1 mmol:
145 x 0,055 = 8,0475

The reason seems to be that the initial conversion to mgdL does not round the result correctly, and truncate the result to the integer value. 

The relevant code is here:

[...]

PR #377, opened 2024-08-10
Improve rounding when converting between mg/dL and mmol/L

MikePlante1:
Addresses Issue #267 

Currently, Trio just rounds down when logging a manual glucose entry in mmol/L. This means that 7.6 gets logged as 7.5 because 7.6 / 0.0555 ≈ 136.9 gets rounded to 136 when `Int(glucose)` is performed which gets displayed as 7.5 because 136 * 0.0555 ≈ 7.5.

This PR rounds `asMgdL` to 0 decimal places and `asMmolL` to 1 decimal place which allows for more accurate rounding when converting units and back.

Here's a spreadsheet that highlights the mmol/L readings that caused mismatches: https://docs.google.com/spreadsheets/d/10ZHV7ZDd0_fjrV6k0FH1ouuqy2TSsDSBpBULE4yAFyg/edit?usp=sharing

### DIFF
--- a/FreeAPS.xcodeproj/project.pbxproj
+++ b/FreeAPS.xcodeproj/project.pbxproj
@@ -287,6 +287,7 @@
 		BD2B464E0745FBE7B79913F4 /* NightscoutConfigProvider.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3BF768BD6264FF7D71D66767 /* NightscoutConfigProvider.swift */; };
 		BDF530D82B40F8AC002CAF43 /* LockScreenView.swift in Sources */ = {isa = PBXBuildFile; fileRef = BDF530D72B40F8AC002CAF43 /* LockScreenView.swift */; };
 		BF1667ADE69E4B5B111CECAE /* ManualTempBasalProvider.swift in Sources */ = {isa = PBXBuildFile; fileRef = 680C4420C9A345D46D90D06C /* ManualTempBasalProvider.swift */; };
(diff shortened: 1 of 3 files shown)


MikePlante1:
I've tested this with manual glucose entries, temp targets, profile overrides, and target glucose. Setting it to a known problem number (like 7.6 or 8.1) in mmol/L and then switching to mg/dL and then switching back to mmol/L.

marionbarker:
## Status:
Success

### Test with dev:
* started with Trio dev, commit c04d33e0
* test a few of the mmol/L values in @MikePlante1 spreadsheet
* use the log glucose screen:
   * enter 6.1, see 6.0 (109)
   * enter 11.1, see 11.1 (200)
   * enter 10.6, see 10.5 (190)

### Test with PR 377
* applied patch for PR
   * requires a Product Clean Folder before building
* test same entries in the log glucose screen:
   * enter 10.6, see 10.6 (191)
   * enter 11.1, see 11.1 (200)
   * enter 6.1, see 6.1 (110)

Modify to mg/dL, add to list above the mg/dL value in parentheses.

marionbarker (approved):

MikePlante1:
Note that this PR does not solve rounding issues pointed out in [this comment](https://github.com/nightscout/Trio/issues/267#issuecomment-2179981903) where settings can shift up to 1 mg/dL when switching units from mg/dL to mmol/L and then back to mg/dL.

I think the best way to solve that issue would be to update Trio so all mmol/L inputs are converted to mg/dL upon entry and only converted back to mmol/L when used for display. (Work has already begun on this in a private fork)

dnzxy (approved):

AndreasStokholm (approved):

PR #378, opened 2024-08-10
Fix capture-build-details.sh to work with detached state or tags

dnzxy:
Fixes build so that Trio can be built from a detached state (single commit checked out) or tags.

Thank you @bjorkert 🙏

### DIFF
--- a/scripts/capture-build-details.sh
+++ b/scripts/capture-build-details.sh
@@ -1,27 +1,22 @@
 #!/bin/sh -e
-
 #  capture-build-details.sh
 #  Trio
 #
 #  Created by Jonas Björkert on 2024-05-08.
-
 # Enable debugging if needed
 #set -x
-
 info_plist_path="${BUILT_PRODUCTS_DIR}/${CONTENTS_FOLDER_PATH}/BuildDetails.plist"
-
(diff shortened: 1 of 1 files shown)


Sjoerd-Bo3 (approved):
Code looks solid. Doesn’t break previous behaviour. Just adds an option for detached.

marionbarker:
## Status
Success

## Tested building a detached HEAD

### Background

Previously when checking out an earlier commit of a given branch and leaving it as a detached HEAD, the capture-build-details.sh failed when trying to build because there was no branch.

### Test

Update local clone to 
* `8811afe0 (dnzxy/fix-capture-build) Fix capture-build-details.sh to work with detached state or tags`
* test that build succeeds
* add empty commit to advance the SHA for the branch (in local clone)

```
git commit --allow-empty -m "empty commit"
```

* checkout a detached HEAD pointing to `8811afe0`

```
git checkout 8811afe0
xed .
```

Build was successful
* Trio, Settings indicates: `branch: detached 8811afe`

marionbarker (approved):

bjornoleh:
Merging with two approvals.

PR #383, opened 2024-08-17
Disable saving autotune to pump

MikePlante1:
0 U/hr basal rates seem to be causing a problem with this functionality.

This just comments out the button for now to prevent users from getting hit with this bug until the old bug is squashed.


AndreasStokholm (approved):

Sjoerd-Bo3 (approved):
Looks good to me

Issue #396, opened 2024-08-24
Rounding error in Log Glucose entry (and possibly elsewhere)


bjornoleh:
## Describe the bug
When entering BG values in the Log Glucose screen, there is rounding applied at the time of entering values. This gives the wrong result sometimes. Confirmed for when entering the value of 9,2 mmol/L, which is immediately registered as 9,1 mmol/L.

## To Reproduce
*Steps to reproduce the behavior:*
1. Set BG units to mmol/L
2. *Go to History / Glucose tab
3. *Click on `Log Glucose +`
4. *Enter `9,2`
5. *See the `New Glucose` value being entered as `9,1`

## Expected behavior
Duh :-)

### Trio Version:
* Version Number: *[e.g. 0.2.0]*
* Repo: *nightscout/trio*
* Git Reference: [6433d94](https://github.com/nightscout/Trio/commit/6433d947ac6abffb4e381c32b59662314b65aa38)

## Technical Details
*Built to phone, but should be easily reproducible in any setup

I'd expect there to be more instances of this rounding error, but did not look for them yet.

dnzxy:
Fixed with #398 .

PR #398, opened 2024-08-25
glucose rounding fix

MikePlante1:
This is to address Issue https://github.com/nightscout/Trio/issues/396.

Since rounding is taken care of in BloodGlucose now, the down rounding in the numberFormatter can be changed to round halfUp.

Tested from 8 to 11 in increments of 0.1 mmol/L. Here's a snipped:
Screenshot 2024-08-25 at 3 53 29 PM


bjornoleh (approved):
Approving based on testing on a real phone, this works as it should when testing 9,1 - 9,3 mmol/L.

Thanks!

dnzxy (approved):
Tested in sim. Small and sane change.
May become obsolete in v1.0.0 with the rework being done to unit parsing and handling. Proceeding with merge, as this LGTM 🚀

PR #404, opened 2024-09-09
[DanaKit] feat: Allow users to disable bolus syncing

bastiaanv:
Some users are reporting issue regarding double bolus entries. This PR allows a user to disable the bolus syncing of DanaKit to prevent the double entries. Normally, a time sync would be enough, but there are cases where this didn't solve the issue.

This PR does bring the risk of users entering a bolus outside Loop/iAPS/Trio and causing the algorithm to make mistakes. But the chance of happening is very low, the risk is medium to high. The user should use this feature with caution!

A visual indicator will be made later on and clear docs are also planned for later. This is a quick PR to solve a urgent issue

EDIT: Also saw that a small time format change sneaked into the PR. It changes the day format from `3.5days` to `3d 12h`. Nice minor change


Sjoerd-Bo3 (approved):
Only touches DanaKit itself and the code looks good.

Let's get this to testers ASAP

PR #413, opened 2024-09-20
update submodule LibreTransmitter: support libre 2 plus european sensors

marionbarker:
Support the new European version of Libre 2 plus


dnzxy (approved):

Sjoerd-Bo3 (approved):
LGTM implements the Libre 2 Plus improvement

PR #421, opened 2024-09-28
Fix possible incorrect total basal units in profile editor

robertdejong1:
Fixes possible incorrect total basal units in profile editor

**Steps to reproduce**
* Edit simple basal profile eg. 00:00 1 u/hr, 12:00 0.5 u/hr
    * Total shows 18 u/day, which is correct.
* Add a new entry and let it start at 10:00 with 0.75 u/hr
    * Total units now shows 21.5 u/day which is incorrect. It should be 17.5 u/day.

**Cause**\
This is caused by the fact that the total is calculated before the entries are sorted.

**Effects**\
It seems to be a visual problem only, saving on pump seems to save the profile correctly. And after reopening the profile editor, the total is calulated correctly aswell.

**Priority**\
Even though this might not be a problem when 1.0 comes with reworked settings, this bug got me scared when editing my basal profile. So I figured this might be worth fixing.

**Deduplication of items**\
Furthermore, upon inspecing the validate() function I noticed that entries will get deduplicated based on the start time. As this method is called .onAppear of the RootView it's possible that while editing an entry it will get depulicated if one selects an already used start time. after that happens, changing the time or rate of this entry will cause the application to crash

I included a fix for that particular situation by making it impossible to select an already used start time, which I feel like is a nice addition from a users perspective aswell.
But this is more of an edge case and this problem also exists in other similar views like the ISF profile editor. So if we think it's worth fixing, I could also fix that in a separate PR and then include the other views aswell.

This is my first PR to this project, so feedback is welcome :)

### DIFF
--- a/FreeAPS/Sources/Modules/BasalProfileEditor/BasalProfileEditorStateModel.swift
+++ b/FreeAPS/Sources/Modules/BasalProfileEditor/BasalProfileEditorStateModel.swift
@@ -82,6 +82,7 @@ extension BasalProfileEditor {
                 let sorted = uniq.sorted { $0.timeIndex < $1.timeIndex }
                 sorted.first?.timeIndex = 0
                 self.items = sorted
+                self.calcTotal()
             }
         }
     }
--- a/FreeAPS/Sources/Modules/BasalProfileEditor/BasalProfileEditorStateModel.swift
+++ b/FreeAPS/Sources/Modules/BasalProfileEditor/BasalProfileEditorStateModel.swift
@@ -86,9 +86,9 @@ extension BasalProfileEditor {
             }
         }
 
-        func availableTimeIndices(forItemIndex: Int) -> [Int]
+        func availableTimeIndices(_ itemIndex: Int) -> [Int]
         {
-            let usedIndicesByOtherItems = items.filter { $0 != items[forItemIndex] }.map(\.timeIndex)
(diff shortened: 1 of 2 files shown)


dnzxy:
Hi @robertdejong1 
thank you very much for your contribution. 
Good find, and a good solution. 

We are not really adding anything on top of 0.2.0 other than critical fixes, because with the (in the works) Trio 1.0.0 the entirety of settings are reworked. 

Let me verify that we have addressed this in 1.0-dev ☺️ if not, I‘ll adapt your work to 1.0 with appropriate contribution credit by you 🙏

Sjoerd-Bo3:
@robertdejong1 are you already on discord? That way it's easier to contribute. We would love fellow dev's / enthusiasts to help.

robertdejong1:
Hi @Sjoerd-Bo3
Yes I'm robert0648 on discord. Would love to help out where I can.

dnzxy (approved):
If you change the function signature to `func availableTimeIndices(_ itemIndex: Int) -> [Int]` it may become a little more readable and you can omit the parameter name when the function is called.

That's about the only thing I could add here. Great contribution. 

Looking at the other settings (glucose target, isf, carb ratio), settings there are essentially copy paste, so this is probably also applicable there; doesn't necessarily fall into the scope of this PR though.

LGTM.

dnzxy:
Fixed in 1.0

PR #459, opened 2025-01-21
[DanaKit] update submodule

bastiaanv:
(Sync with TCD)

Changelog to DanaKit:
- Allow connection to be reused
  - CoreBluetooth doesn't allow us to directly disconnect BLE devices. If another app uses the same device, it will keep the connection alive. Therefore, the disconnect function doesn't work every time.
  - This fix will allow DanaKit to reuse old connection as long as no commands are using the BLE socket
- keep connection alive while bolus
  - An iAPS user reported this issue, he is using the pump at it's lowest bolus speed 1min per 1U
  - When he enacts a bolus of 12U, the pump will disconnects it self
  - In trio is not a big of a deal, since the entry is already stored, but for iAPS this is a big problem
  - But it is always nice to keep the connection open during bolus, to check for occlusions or empty reservoirs
- Fix negative timezone crash
- correctly reporting pump time


Sjoerd-Bo3 (approved):
LGTM

Sjoerd-Bo3:
@bastiaanv we have merge conflicts

PR #467, opened 2025-02-02
[DanaKit] update submodule 

bastiaanv:
Changelog to DanaKit:
- Allow connection to be reused
  - CoreBluetooth doesn't allow us to directly disconnect BLE devices. If another app uses the same device, it will keep the connection alive. Therefore, the disconnect function doesn't work every time.
  - This fix will allow DanaKit to reuse old connection as long as no commands are using the BLE socket
- keep connection alive while bolus
  - An iAPS user reported this issue, he is using the pump at it's lowest bolus speed 1min per 1U
  - When he enacts a bolus of 12U, the pump will disconnects it self
  - In trio is not a big of a deal, since the entry is already stored, but for iAPS this is a big problem
  - But it is always nice to keep the connection open during bolus, to check for occlusions or empty reservoirs
- Fix negative timezone crash
- correctly reporting pump time
- Prevent app crash using DanaRS & pump setting sync
- Update translations


dnzxy (approved):

bastiaanv:
@Sjoerd-Bo3

marionbarker:
## Status
LGTM

## Testing

Built on iPhone 8 running iOS 16.
Connected to rPi Dana simulator.
* modified time zone on phone, wait one cycle, no notification on main screen
   - I know this is fixed in closed beta, so don't expect any changes in 0.2.3
   - The Dana screen shows the time zone change, update pump and timezone is updated
* restore time on phone and make sure it gets fixed going the other way too
  - going in the other direction, Trio timed out on making connection to the pump when looping 
  - tried to sync the pump time on the Dana screen and Trio crashed (screenshot of Xcode saved).
  - rebuilt Trio and this time the sync pump time was successful
  - this is probably good enough (a restart fixed the issue and it could have been the simulator)
* exercise all the other options on the Dana screen
   * stop temp basal (another crash, fixed by rebuild) - no way of knowing if real or simulator issue, but works fine after restart
   * suspend insulin delivery (OK)
   * resume insulin delivery (OK)
   * manual bolus (OK)

## Browser Build

Push bastiaanv's feat/dana version to loopdocs-tester/Trio (branch name bast_feat_dana) and test the build using Browser Build method
* Build was successful
* Install the TestFlight build onto the iPhone 8 tested above
* keep phone open on the app and let it keep looping (no heartbeat so have to keep it in foreground and open)
* let it run for 15 minutes
* maintained green loop the whole time

marionbarker (approved):

PR #477, opened 2025-03-17
update G7SensorKit: scan-fix modification

marionbarker:
Use @motinis proposed scan-fix branch for G7SensorKit
* This improves performance in which G7 sensor readings have not been appearing in the main Trio screen for use in looping
* There can still be issues, but this reduces the number

Because this PR used the current `trio` branch for the loopandlearn G7SensorKit repository, it also brings in modifications to the `pxbproj` file that will later be required to modify the build order for the app. That commit has no effect in how the code performs.


dnzxy (approved):

t1dude (approved):

motinis:
Note that Pete brought a different fix - I think it is working better, although you must start scanning explicitly for a new sensor yourself. There is still work-in-progress there, but for your consideration. https://github.com/LoopKit/G7SensorKit/pull/34

marionbarker:
Additional changes and logging were added to that proposed change. I would let the testing on Loop continue before suggesting changes on Trio.

dnzxy:
For as long as Pete cannot properly explain what the changed condition is attempting to do or fix, we'll go with this attempted fix solution.

PR #478, opened 2025-03-19
Remove Close button from Enact Bolus after Add Carbs

aidanlane:
For the Enact Bolus screen, only show close nav bar button and allow drag to dismiss if not following the Add Carbs screen, otherwise force use of the "Continue without bolus" button to close.

Without this patch in place, some people, especially new users and occasional caretakers (e.g. school staff) may tap the "Close" button in the Enact Bolus screen thinking that it will dismiss the carb entry too, such as in the case where they want to double check something before entering them (again). This can lead to double carb entries.

Note that the Swift UI native way to prevent drag to dismiss is broken in iOS 18.1: https://developer.apple.com/forums/thread/767475
We could introduce a UIKit workaround for that, however that would make the patch much larger and provides questionable value, especially with the Trio 1.0 public beta on its way.


aidanlane:
Note that there is no expectation of this being merged. PR raised purely for awareness for that others can benefit from it prior to Trio 1.0 being released, which will resolve this by merging the Carb and Bolus screens.

For reference: https://discord.com/channels/1020905149037813862/1239982177190678549/1349331896860938241

aidanlane:
### When Adding Carbs (close button removed from waiting and bolus screens):


### When directly Enacting a Bolus (close button still available):

dnzxy:
Closing as this was addressed with a major UI and UX redesign in upcoming public beta 0.5.0.
Thank you for your contribution and wanting to fix this. Please feel free to jump on any glaring issues or bugs, once 0.5.0 goes public, dear @aidanlane 🙏

PR #508, opened 2025-05-17
Ensure proper user intent tracking for BLE denial #503

### LINKED ISSUE #503: Bluetooth not able to be enabled after not allowed during onboarding
## Describe the bug
If Bluetooth is not enabled during onboarding, it can seemingly only be enabled by deleing and reinstalling the app to go through onboarding again (and this time enabling it)

## To Reproduce
1) Install fresh 0.5.0 app to a real phone (not a simulator)
2) In onboarding, select "Don't Allow" when you get to Bluetooth permissions
3) Finish onboarding to see the "Bluetooth Required" error
4) Tap to go to iOS settings and see that Bluetooth is not an option to enable.
5) Uninstall, Reinstall, agree to Allow BT this time

## Expected behavior
Should be able to toggle BT on in settings after finishing onboarding without agreeing to enable it

### Trio Version:
* Version Number: 0.5.0
* Repo: *nightscout/trio*
* Git Reference: [f445d78](https://github.com/nightscout/Trio/commit/f445d78ac30e449bfa2dada68deadbf43526d455)

dnzxy:
## Summary

This PR updates the Bluetooth permission onboarding step to ensure proper user intent tracking when the system Bluetooth permission prompt is denied by the user. Fixes #503.

### Changes
* Calls `authorizeBluetooth()` even in the `"Don’t Allow"` path to:
  * Trigger `CBCentralManager` initialization
  * Ensure iOS recognizes Trio as a Bluetooth-using app
  * Enable manual reactivation via **Settings > Privacy & Security > Bluetooth**

#### Additional Context

Calling `authorizeBluetooth()` regardless of user choice ensures the app is properly registered in iOS privacy settings. This prevents the Bluetooth toggle from being missing in iOS Settings when a user initially denies access.

#### Note:
We are bound to the bluetooth authorization state here; the app requires BLE to function, period. So we unfortunately cannot offer a single pump selection (simulator). I don't think this is an issue. Just allow BLE 🤓

### DIFF
--- a/Trio/Sources/Modules/Onboarding/View/OnboardingSteps/BluetoothPermissionStepView.swift
+++ b/Trio/Sources/Modules/Onboarding/View/OnboardingSteps/BluetoothPermissionStepView.swift
@@ -69,6 +69,15 @@ struct BluetoothPermissionStepView: View {
                 allowTitle: String(localized: "Allow"),
                 denyTitle: String(localized: "Don’t Allow"),
                 onAllow: {
+                    /// Requests Bluetooth permission and updates onboarding state based on the system’s response.
+                    /// It calls `authorizeBluetooth`, which initializes `CBCentralManager` and triggers the
+                    /// native system permission prompt (if not previously shown).
+                    ///
+                    /// The resulting authorization is checked — if the user grants permission (`.authorized`),
+                    /// `hasBluetoothGranted` is set to `true`, allowing the app to proceed with Bluetooth operations.
+                    /// Otherwise, it remains `false`, and the user can be guided to manually enable Bluetooth later.
+                    ///
+                    /// This ensures the app only treats Bluetooth as granted when the system confirms it.
                     bluetoothManager.authorizeBluetooth { auth in
                         DispatchQueue.main.async {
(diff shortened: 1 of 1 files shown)


MikePlante1 (approved):
I reproduced this error on a real in-vitro iPhone SE (2nd gen) running iOS 18.2.1. I then built this PR to that same phone and the error was not reproduced.

https://github.com/user-attachments/assets/3ff88fdf-0a2d-4858-84c9-e1b200c927da

I also ran through onboarding 3 more times and verified the bluetooth toggle is now always available in iOS settings no matter what selections were made during onboarding.

| first alert | second alert | Bluetooth result |
| --- | --- | --- |
| Don't Allow | Don't Allow | Disabled |
| Don't Allow | Allow | Enabled |
| Allow | Allow | Enabled |
| Allow | Don't Allow | Disabled |


LGTM

kingst (approved):
It's a little strange to prompt twice (tested on an old iPad running iOS 17), but this is a rare case. I confirmed that bluetooth does show up in settings and prompting twice is better than having people get stuck, so LGTM.

Issue #506, opened 2025-05-17
[Polish][Translation] Invalid translation for warning message in utilities settings


anywheresebzaj:
## Describe the bug
The key "The settings in this section typically do not require ANY modifications. Do not alter them without a solid understanding of what you are changing and the full impact it will have on the algorithm." contains an invalid Polish translation. Currently it says "ANAL DATA", so somebody made a mistake in translation

## Preconditions
Application language is set to Polish

## To Reproduce
*Steps to reproduce the behavior:*
1. Go to settings
2. Click "Algorithms" settings
3. Click "Utilities"/"Additional"
4. Observe the "Warnings" 

## Expected behavior
The valid translation is "Ustawienia w tej sekcji zazwyczaj nie wymagają ŻADNYCH modyfikacji. Nie zmieniaj ich bez solidnego zrozumienia tego, co zmieniasz i pełnego wpływu, jaki będzie ona miała na algorytm."

### Smartphone:
* Hardware: *iPhone 13 Pro Max*
* OS Version: *18.0.1*

### Trio Version:
* Version Number: *0.5.0*
* Repo: *nightscout/trio*

MikePlante1:
Thanks for flagging this!

Fyi, if you’re interested in helping with translations, you can join the fun at crowdin.triodocs.org (@t1dude might have to set you up, not sure how that works)

grostecki:
Poprawione. Wkradła się literówka ANY-> ŻADNYCH a wyszło analnych :-D

t1dude:
Polish translation was updated and approved an hour ago. This was pulled in with PR 509 https://github.com/nightscout/Trio/pull/509

<img width="1131" alt="Image" src="https://github.com/user-attachments/assets/6c96c015-087d-4aea-a42b-98c3f34433e2" />

Issue #505, opened 2025-05-17
App Crash on Step "Algorithm Settings > High Temp Target" on iOS 17.6.1


dnzxy:
## Describe the bug

User RedneckNinja reported on Discord, that when proceeding through onboarding after upgrading from `v0.2.5` to `v0.5.0`, their app crashed consistently at the "Algorithm Settings > High Temp Target" step when pressing "Next". This occurred despite default settings and various toggles being tested. Even after deleting and reinstalling the app with no prior configuration, the crash persists at the same step.

Eventually, the crash could be bypassed by updating iOS from **17.6.1 to 18.5** and **switching the system language from Swedish to English**.

## Attach a Log

* Attached crash log:  [Trio-2025-05-17-065458.ips.zip](https://github.com/user-attachments/files/20263993/Trio-2025-05-17-065458.ips.zip)

* Attached logfile: [Trio-2025-05-17-LOG.txt](https://github.com/user-attachments/files/20263996/Trio-2025-05-17-LOG.txt)

## To Reproduce

1. Install Trio v0.5.0 via TestFlight
2. Launch app and begin onboarding (fresh install or upgrade from v0.2.5)
3. Proceed to "Algorithm Settings" step
4. On "High Temp Target" page, press "Next"
5. App crashes

## Expected behavior

The app should transition to the next step in onboarding without crashing.

## Screenshots

N/A

### Smartphone:

* Hardware: iPhone SE
* OS Version: iOS 17.6.1 (issue), iOS 18.5 (worked after update)

## Trio Version:
* Version Number: 0.5.0 (8)
* Repo: `nightscout/trio`
* Git Reference: `91a9a92` (dev)

## Additional context

* Crash occurred repeatedly at same point, regardless of prior Nightscout setup or glucose unit selection
* User attempted up to 20 retries before switching language and updating iOS
* System language may affect localization or persistence logic during onboarding

MikePlante1:
I was able to reproduce this with iOS 17.5 on an iPhone SE (2nd gen) simulator when the phone was set to Swedish.

Not reproduced when:
* Using real phone or simulator phone with Swedish or english with iOS 18.4 or 18.5
* using a simulator with iOS 17.5 set to english

t1dude:
Working on a fix now.

t1dude:
Fixed in PR 509: https://github.com/nightscout/Trio/pull/509

1. Fixed translation in Crowdin
2. Pulled String Catalog from Crowdin
[...]

PR #509, opened 2025-05-17
Fixing issue #505 and issue #506 and updated translations from Crowdin as of 2025-05-17.

t1dude:
Address an issue with one string in the Swedish translation which ended in a `%` character, like `…some text here 100%%`. It should have ended with `%%`, as the first `%` serves as delimiter. 

Fixed in Crowdin translation by adding a `%`. But, when I pulled the string catalog, it also adds all other translations that have been made since yesterday.

Additionally ensured no other reoccurrences of that particular faulty string. I could not find any others who have chosen to translate %% into %.


kingst:
Is it possible to test all of our string / language combos for this bug just to make sure?

t1dude:
This PR also pulls other translations that were updated since yesterday. Seems to address https://github.com/nightscout/Trio/issues/506 as well.

MikePlante1 (approved):
I reproduced the error in an iOS 17.5 iPhone SE (2nd gen) simulator with language set to Swedish. I then built this PR to that same sim and the error was not reproduced.

LGTM

MikePlante1:
But maybe update the PR description to include that other translations are updated in this commit as well.

t1dude:
> But maybe update the PR description to include that other translations are updated in this commit as well.

Done.

kingst (approved):
Having a unit test would be a nice, but optional, addition just to protect against translation issues in the future. LGTM

dnzxy (approved):

Issue #511, opened 2025-05-17
The system is not delivering bolus increments of 0.05


Larissa-Alvarenga:
[preferences.json](https://github.com/user-attachments/files/20266047/preferences.json)

## Describe the bug
The system is not delivering bolus increments of 0.05, for example I want to give a bolus of 0.15 the system delivers 0.1.

## Attach a Log
*Tap the Trio settings icon at the bottom of the screen, then tap 'Share logs' on the bottom of the list and attach it to this ticket.*

## To Reproduce
*Steps to reproduce the behavior:*
1. *Go to '...'*
2. *Click on '....'*
3. *Scroll down to '....'*
4. *See error*

## Expected behavior
*A clear and concise description of what you expected to happen.*

## Screenshots
*If applicable, add screenshots to help explain your problem.*

### Smartphone:
* Hardware: iPhone 13 Pro
* OS Version: iOS 18.4

### Pump:
* Manufacturer: Insulet
* Model:  Omnipod Dash

### CGM:
* Device: Dexcom G7
* Manager app: Dexcom App

### Trio Version:
* Version Number: V05.0
* Repo: 

[log_prev.txt](https://github.com/user-attachments/files/20265825/log_prev.txt)
[log.txt](https://github.com/user-attachments/files/20265826/log.txt)

nightscout/trio
* Git Reference: *[e.g. commit hash]*

## Technical Details
*If applicable, provide any technical details that might help in diagnosing the problem. This could include logs, error messages, or relevant configuration details.*

## Additional context
*Add any other context about the problem here.*

kingst:
I was able to reproduce this on the simulator using my standard setup, closed loop on, and configuring the simulated pump to emulate "pod" settings.

MikePlante1:
Something you could try the next time you change your pod, if this hasn’t been addressed by then: remove the dash as your pump of choice, so it says “add pump” on the main screen. Then select to add dash again.

NightScoutForSeb:
Where do I find that .json file so I can confirm before a reboot?


Hanna SandovalSent from Yahoo Mail for iPhone


On Sunday, May 18, 2025, 12:37 AM, davidh5480 ***@***.***> wrote:

davidh5480 left a comment (nightscout/Trio#511)
[...]

PR #518, opened 2025-05-18
Fix FPU Processing for Possible out of range Values

### LINKED ISSUE #520: Guardrails not honored with migration
## Describe the bug

User was able to increase her spread interval to 90 min initially after onboarding. It was set to 90 min in Trio 0.2.5. The UI reset it to 60 minutes during onboarding. The issue is she was _able_ to set it to 90 min. When she went back to confirm the ability to set above the guardrail, it only allowed her to go up to 60 min and the guardrail was honored. It wasn’t honored the first time.

If this is set outside of the guardrails, FPU calculations result in a very large number of 1g carb entries over an extended period of time.

## To Reproduce
1. Build Trio 0.2.5
2. Set FPU spread interval to 90
3. Update to Trio 0.5.0
4. After onboarding, spread interval will say 60, but will allow you to increase to 90.
5. If you set it outside of the guardrails, it will cause the mentioned issue with FPU calculations.

## Expected behavior
Guardrails to be honored and migrated settings outside of those limits to be either reset to default or set to the closest in-bounds limit.

@MikePlante1 was able to replicate

### Trio Version:
* Version Number: Trio 0.5.0
* Repo: nightscout/trio (dev)

## Technical Details
@MikePlante1 can you update this issue with your additional context, if needed?

dnzxy:
## Summary
This PR adds a guarding logic for possible out-of-range values for the FPU settings following a migration from `v0.2.5` to `v0.5.0`. Addresses and fixes #520. 

### Changes
* Introduced upper bounds for key FPU-related settings (`timeCap`, `minuteInterval`, `delay`, `individualAdjustmentFactor`) to prevent invalid configurations post-upgrade.
  * ~~Takes the smaller of the two values at hand: the read input value and the maximum allowed value as of `v0.5.0`~~
  * Clamps the value and verifies it lies within allowed Trio `v.0.5.0` bounds.
  * Validation is applied when generating carb equivalents and durations in CarbsStorage.

* Changed types for `timeCap`, `minuteInterval`, and `delay` in TrioSettings from `Int` to `Decimal` to improve precision and alignment with user inputs.

### Additional Context
Due to a small regression during Onboarding, it is possible, that users who had very out of range settings in `v0.2.5` now still retain them in `v0.5.0` post Onboarding, where these ratings are now out of range. This leads to a faulty behavior of carb equivalent spread creation. This was reported on Facebook.


kingst (commented):
The code LGTM. I tested it in the simulator but wasn't able to set out of bounds ranges with 0.2.5. Do I need to modify the JSON directly?

One question: If we're checking for wild max values, should we also check for min values? In other words, do we need to use a clamp instead of a min check?

dnzxy:
> The code LGTM. I tested it in the simulator but wasn't able to set out of bounds ranges with 0.2.5. Do I need to modify the JSON directly?

@MikePlante1 has tested this and the spread was no longer broken and 1g entries for hours exceeding `timeCap`. 

Either modify the JSON before building over, just enter it in 0.2.5, as they are text inputs you should be able to. 

> One question: If we're checking for wild max values, should we also check for min values? In other words, do we need to use a clamp instead of a min check?

Good question. I hadn’t considered it but it may make sense. In our specific case here, with the calculation we do, it becomes an issue the moment values exceeding 60 min as that breaks the carb equivalent size. Generally speaking may be a sensible change. I could adopt the clamp over min check in #519 if you want, as I am min-guarding all number based Trio settings there.

dnzxy:
@kingst I have changed this to adopt clamping instead of the previous max-value check.

kingst (approved):
Looks good!

MikePlante1 (changes requested):
Either the name and description of "Spread Interval" needs to be changed, or the code in the PR needs to be changed so it works like it did before (at least for values that aren't outside the guardrails)

I used these settings:
Screenshot 2025-05-18 at 8 55 12 PM

Entered this meal:

Screenshot 2025-05-18 at 8 55 40 PM

And got this result, with entries only spaced out every minute instead of every 60m

Screenshot 2025-05-18 at 8 57 00 PM

dnzxy:
@MikePlante1 could you give this another test-run and then get this in, as #519 builds on this one and we need to get #519 in to solve #511 which is rather high priority.

dnzxy:
Did a few tests in-vivo and discussed this further with @kingst via DM. This should be fine. Since this is a prerequisite for #519, will move ahead and merge this. Thanks you both.

marionbarker:
## Test

Start with 0.2.5 build on a test phone (SE 2nd gen running iOS 18.5).
* the guardrails for 0.2.5 are same as for 0.5.0 for these quantities:
   * `Fat and Protein Delay` max 120 min (typed 140)
   * `Maximum Duration` max 12 hours (typed 14)
   * `Spread interval` max 60 min (typed 90)
   * `Fat and Protein Percentage` 1.2 (120%) (typed 1.4)
* I typed in larger numbers for each, but they returned to the maximum values when I looked back at meal entry screen
* When I entered 14 g Carb, 44 g Protein, 88 g Fat, I got multiple screens of 1 g FPU entries (captured screen shots of the first and last screens)
* See graphic below - left and middle panels are from 0.2.5

Build commit `457b3e575`, branch `fix-fpu-processing`
* onboarded - accepting defaults
* leave the 4 quantities at maximum values
* enter same meal as above
* got reasonable distribution of FPU items
* See graphic below - right panel is from this build of Trio 0.5.0

trio-pr518

marionbarker (approved):
Approve from code review and test.

Issue #510, opened 2025-05-17
Treatment View Determination Failure with no Auth Method Configured


dnzxy:
### Describe the bug

Reported on Facebook. When attempting to manually bolus in the Treatments screen, the app shows the following error:

“Determination Failed — Failed to update COB/IOB: The operation couldn’t be completed. (Trio.UnlockError error 1.)”

This happens even though all settings and inputs seem valid. **The underlying cause appears to be the absence of a device passcode or biometric authentication setup (Face ID / Touch ID).**

### Attach a Log

No log was provided for this report, but error screenshot attached.

### To Reproduce

1. Install and configure Trio on an iPhone without a passcode or biometric lock
2. Open the Treatments screen
3. Enter carbs or bolus and tap to apply
4. Error appears: `Trio.UnlockError error 1`

### Expected behavior

Trio should either:

- Prompt the user to set up authentication before attempting secure operations, or
- Handle the missing device security gracefully with a clear message

Screenshots
![Image](https://github.com/user-attachments/assets/2702f6ca-fb2f-4fa6-bce1-e43704b2cba2)

### Setup Information

Smartphone:

- Hardware: iPhone (exact model unknown)
- OS Version: Unknown
- No passcode / Face ID / Touch ID enabled
- Pump: Unknown (not specified)
- CGM: Unknown (not specified)

### Trio Version:
- Version Number: 0.5.0
- Repo: nightscout/trio

### Technical Details

Trio uses LocalAuthentication for confirming bolus and secure operations. If the phone has no secure authentication mechanism enabled, this can lead to UnlockError error 1.

### Additional context

After enabling a passcode (and likely Face/Touch ID), the issue was resolved. The app should consider warning the user during onboarding or treatment setup if secure authentication is required.

PR #524, opened 2025-05-18
Improve Authentication Failure Feedback in Treatments View

dnzxy:
## Summary
This PR addresses #510 by improving the user-facing experience when biometric or device passcode authentication fails during treatment operations (e.g., manual bolus or external insulin entry).

### Changes

**1. Descriptive Error for Missing Passcode**

* When the unlock operation fails due to no passcode being set on the device, the user is now shown a helpful message:

  > “Authentication failed. Please ensure you have configured a Passcode for your iPhone under iOS Settings > Face ID & Code.”
* Replaces generic system error text with a localized string to clarify the failure and guide users toward resolution.

**2. Updated Alert Messaging**

* Renamed the alert title from “Determination Failed” to “Error while processing Treatment” for more accurate and user-friendly language.
* The alert message now directly displays the custom `determinationFailureMessage`.

**3. Replaced `print` Statements with Structured Debug Logging**

| OLD | NEW |
|--------|--------|
| Screenshot 2025-05-18 at 22 00 38 | Screenshot 2025-05-18 at 22 03 03 |

### Chores
* Migrated all `print(...)` statements in various modules (`APSManager`, `DeviceDataManager`, `GlucoseStorage`, `Script`) to use the `debug(...)` logging system and non-localized errors.


Trio/Sources/Modules/Treatments/TreatmentsStateModel.swift
```diff
@@ -474,14 +475,19 @@ extension Treatments {
                     }
                     await apsManager.enactBolus(amount: maxAmount, isSMB: false, callback: nil)
                 } else {
-                    print("authentication failed")
+                    debug(.bolusState, "Authentication failed")
```
kingst, line 478: From looking at the UnlockManager, it looks like the `unlock` function can never return `false` so this branch will never execute.

Trio/Sources/Modules/Treatments/TreatmentsStateModel.swift
```diff
@@ -510,14 +516,19 @@ extension Treatments {
                     // perform determine basal sync
                     try await apsManager.determineBasalSync()
                 } else {
-                    print("authentication failed")
+                    debug(.bolusState, "Authentication failed")
+                    throw UnlockError(error: LAError(.authenticationFailed))
```
kingst, line 520: same here, this branch can't execute

Trio/Sources/Modules/Treatments/TreatmentsStateModel.swift
```diff
@@ -474,14 +475,19 @@ extension Treatments {
...
-                print("authentication error for pump bolus: \(error.localizedDescription)")
+                debug(.bolusState, "Authentication error for pump bolus: \(error)")
                 await MainActor.run {
                     self.isAwaitingDeterminationResult = false
                     self.showDeterminationFailureAlert = true
-                    self.determinationFailureMessage = error.localizedDescription
+                    self
+                        .determinationFailureMessage =
+                        String(
+                            localized: "Authentication failed. Please ensure you have configured a Passcode for your iPhone under iOS Settings > Face ID & Code."
```
kingst, line 489: Do we need to have separate strings for authentication failure, passcode not set, and maybe a generic error? The exception should have the details:

https://developer.apple.com/documentation/localauthentication/laerror-swift.struct/code

Trio/Sources/Modules/Treatments/TreatmentsStateModel.swift
```diff
@@ -510,14 +516,19 @@ extension Treatments {
...
-                print("authentication error for external insulin: \(error.localizedDescription)")
+                debug(.bolusState, "authentication error for external insulin: \(error)")
                 await MainActor.run {
                     self.isAwaitingDeterminationResult = false
                     self.showDeterminationFailureAlert = true
-                    self.determinationFailureMessage = error.localizedDescription
+                    self
+                        .determinationFailureMessage =
+                        String(
+                            localized: "Authentication failed. Please ensure you have configured a Passcode for your iPhone under iOS Settings > Face ID & Code."
```
kingst, line 530: Same question here about different text for different error types

kingst (commented):
I wasn't able to test this due to me forgetting my Apple ID password for my test device, thus I couldn't disable my passcode and touch ID.

My big question is around different text for different types of failures, the comments about code that doesn't execute are minor and can/should be ignored.

Trio/Sources/Modules/Treatments/TreatmentsStateModel.swift
dnzxy: I‘ll lock into adding a switch case here and adjust the error message response.

marionbarker:
## Test

What a pain - not easy to turn off passcode - but I did it.

With Trio 0.5.0, trying to bolus without a passcode or fingerprint fails.

Build using this branch, I get a nice message saying Authentication is required

The message assumes Face ID & Passcode. However, on phones with a home button, it should say Touch ID & Passcode.

The instructions on line say:
```
If your iPhone has Face ID, tap "Face ID & Passcode". 
If it has Touch ID, tap "Touch ID & Passcode". 
```
trio-pr524

Trio/Sources/Modules/Treatments/TreatmentsStateModel.swift
```diff
@@ -457,6 +458,91 @@ extension Treatments {
...
+                )
+
+            case .notInteractive:
+                return String(
+                    localized: "Authentication UI cannot be displayed. Try restarting the app."
+                )
+
+            case .passcodeNotSet:
+                return String(
+                    localized: "Authentication requires a device passcode. Please set one in iOS Settings > Face ID & Passcode."
```
marionbarker, line 514: Suggest modification:

Please set one in iOS Settings > Face ID & Passcode or Touch ID & Passcode.

marionbarker (changes requested):
Approve based on test but suggested change to message to cover different types of iPhones.

kingst (approved):
The PR LGTM. I tested it on the simulator with both the success and error case (user cancels) and it worked in both cases

dnzxy:
@marionbarker We won’t be able to distinguish that during that situation, so I won’t bother adjusting, sorry. Of authentication fails, we cannot access the biometry authentication type from the LAContext, as it failed and threw an error. It also only applies to 2 phones, SE 2nd and 3rd gen. I‘m counting on users to be able to do some abstraction and realize they are after passcode.

PR #525, opened 2025-05-18
fix[DanaKit]: small fixes

bastiaanv:
List with small fixes:

- Prevent DanaKit from double recording resume & suspend events
- Prevent app crash, during slow BLE connection. Switched from CheckedContinuation to AsyncStreaming
  - Using the CheckedContinuation twice or more cashes app crashes. This cannot be catch'ed or anything, thus switching to the more safe AsyncStreaming
- Seperated the Stop temp basal button from the Suspend/Resume delivery button


dnzxy (approved):
Approving this based on code review.

dnzxy:
Merging based on in-vivo testing by Dana users. 
Thanks for the fix, @bastiaanv 🤝

Issue #533, opened 2025-05-19
Elapsed time calculated unevenly


g5WS:
Here's an issue that was present in previous versions that I've never understood.  There may be a good reason for this behavior, but it's not something I currently see.

I'll report rounded seconds to make things easier to see.  The cycle starts when xDrip receives a new BG.  We'll call that time 0.

0 s/0 s: xDrip receives new BG.
10 s/10 s: Trio sees new BG, and reports BG received "<1 min" ago.
50 s/60 s (1 minute): Trio reports BG received "1 min" ago.  Trio is using xDrip's timestamp.
30 s/90 s (1.5 minutes): Trio reports BG received "2 min" ago.  This is the weird one, as it's only been 30 seconds since Trio reported "1 min".  Not even xDrip had the BG 2 minutes ago.
60 s/150 s (2.5 minutes): Trio reports BG received "3 min" ago.
60 s/210 s (3.5 minutes): Trio reports BG received "4 min" ago.
60 s/270 s (4.5 minutes): Trio reports BG received "5 min" ago.
40 s/310 s (5 minutes, 10 seconds): Trio receives new BG, reports "<1 min" ago, and the cycle begins again.

Time since last loop is calculated and displayed similarly.

It may be worth reconsidering how time is calculated and displayed.  It's concerning to see a BG reading that is 5+ minutes old; it raises concerns that xDrip has dropped a reading.  The expected ages would be "<1 min", "1 min", "2 min", "3 min", "4 min" for most people.  Perhaps consider using xDrip's elapsed time, and only showing "5 min" for the 10 seconds it takes Trio to pick up xDrip's new reading?

![Image](https://github.com/user-attachments/assets/42afc9bc-d103-4058-ab23-eeb0cac78f3e)

dnzxy:
xDrip4iOS writes its readings to a "shared container" container from which Trio reads it from. If Trio makes it to 6min ago for the BG value, it really actually means that the reading was not received at the time and xDrip dropped a reading or there was latency in the "feeding-into" of the reading. 
The timestamp used by Trio for that reading is the creation timestamp in the database I believe, that is what drives the minutes ago string. I can look this up in more detail and lay out the logic, but that will have to wait until later, I'm afraid.

g5WS:
Not to argue, but to clarify: why is the gap between “1 min” and “2 min” only 30 seconds?

dnzxy:
> @g5WS :
[...]

PR #548, opened 2025-05-21
Fix Elapsed Time Display

dnzxy:
## Summary

Fixes inconsistent elapsed time display (e.g. glucose age jumping to "2 min" after only 30 seconds) by introducing a shared helper for consistent time formatting between `CurrentGlucoseView` and `LoopView`.

## Changes

* Added `TimeAgoFormatter` with:

  * `minutesAgo(from:)` → formatted string like `"< 1 m"` or `"2 m"`
  * `minutesAgoValue(from:)` → floored integer minutes for logic use
* Updated `CurrentGlucoseView` to use `TimeAgoFormatter` for glucose age
* Updated `LoopView` to use the same helper for loop status time

## Additional Context

This addresses #533, where elapsed time was shown unevenly due to floating-point behavior between our "Glucose Bobble" and the Loop view.


kingst (approved):
Thanks for fixing this, the code LGTM

Issue #544, opened 2025-05-20
Trio Should Only Register Carbs Written by Trio (to Prevent Dangerous Cross-App Carb Entry Duplication)


brmccollum:
## Describe the bug
*When switching from IAPS to Trio with Nightscout integration, I observed that IAPS was writing erratic, repeated future carb entries to Nightscout. Trio, in turn, used these entries and treated them as valid, resulting in repeated, automatic carb entries and subsequent insulin dosing.*

## Expected behavior
*Trio should not use carbs entered by on nightscout from another closed loop system*

## Screenshots

![Image](https://github.com/user-attachments/assets/4530813e-c8d9-47b4-9d3f-42097bb330da)

![Image](https://github.com/user-attachments/assets/9e068370-58ab-4795-b2f9-892bfe5cc08c)

<img width="1340" alt="Image" src="https://github.com/user-attachments/assets/0019dc12-c1e0-401e-a8c5-2106cb552507" />

### Smartphone:
* Hardware: *Iphone 15 Pro Max*
* OS Version: *iOS 18.2*

### Pump:

* Model: *Omnipod Dash*

### CGM:
* Device: *Dexcom G7*
* Manager app: *Dexcom App*

### Trio Version:
* Version Number: *Trio v0.5.0.3*
* Repo: *nightscout/trio*

## Technical Details
*If applicable, provide any technical details that might help in diagnosing the problem. This could include logs, error messages, or relevant configuration details.*

## Additional context
*Not sure why iAPS was writing the carbs in the first place but turning off its access to Nightscout and deleting all future entries from nightscout (there were over 260) solved the issue.  It was suggested to me on discord to write the issue here to have Trio future proof itself from these type of issue as an added level of safety.*

dnzxy:
The main issue with this is that if you use the NS Care Portal to remotely log `carb correction`s, you are not required to provide an `enteredBy` key. So we could guard the fetch so, that it does not fetch carbs that have been authored by `iAPS`, but we cannot just guard against *any* other or non existent `enteredBy` key, as that would break Care Portal functionality with Trio.

bjorkert:
Currently working on this.

PR #550, opened 2025-05-21
Filter out carbs and temp targets for non manual entries

bjorkert:
## Summary

This PR addresses [#544](https://github.com/nightscout/Trio/issues/544) by ensuring that Trio ignores Nightscout treatment entries (carbs and temp targets) that originate from other closed loop systems such as iAPS, AndroidAPS, or Loop.

Previously, Trio could inadvertently read and act on entries written by other apps, possibly leading to unsafe automatic insulin dosing.

## Changes

- Introduced a `makeNeQueryItems()` helper in `NightscoutAPI.swift` that builds a safe Nightscout `$and` + `$ne` query to filter out unwanted `enteredBy` sources.
- Replaced ad-hoc filtering in `FetchTreatmentsManager` with filtering done directly at the Nightscout API request level.
- Removed now-unnecessary filtering of carbs and temp targets after fetching.
- Ensured values like `loop://iPhone` are correctly filtered without causing 500 errors from Nightscout (which occurs when such strings are passed in `$nin` arrays).
- Preserved entries with custom or no `enteredBy`, to maintain compatibility with Care Portal usage.

## Filtered `enteredBy` sources

These are excluded from remote treatment imports:
- `NightscoutTreatment.local` (Trio itself)
- `AndroidAPS`
- `openaps://AndroidAPS`
- `iAPS`
- `loop://iPhone`

## Why `$ne` instead of `$nin`?

Nightscout's query validator incorrectly fails when `$nin` contains values like `loop://iPhone`, due to how it sanitizes array elements.

## Testing

- ✅ Verified that `loop://iPhone` and similar values are excluded as expected.
- ✅ Confirmed that no `500` errors occur when the Nightscout query includes complex `enteredBy` strings.
- ✅ Confirmed that entries where enteredBy is empty or contain custom values (from Care Portal) are preserved and stored.

### DIFF
--- a/Trio/Sources/APS/FetchTreatmentsManager.swift
+++ b/Trio/Sources/APS/FetchTreatmentsManager.swift
@@ -35,17 +35,17 @@ final class BaseFetchTreatmentsManager: FetchTreatmentsManager, Injectable {
                         async let carbs = self.nightscoutManager.fetchCarbs()
                         async let tempTargets = self.nightscoutManager.fetchTempTargets()
 
-                        // Filter and store if not from "Trio"
-                        let filteredCarbs = await carbs.filter { $0.enteredBy != CarbsEntry.local }
-                        if filteredCarbs.isNotEmpty {
-                            try await self.carbsStorage.storeCarbs(filteredCarbs, areFetchedFromRemote: true)
+                        // Store carbs directly (no filtering, as it's done in fetchCarbs)
+                        let fetchedCarbs = await carbs
+                        if fetchedCarbs.isNotEmpty {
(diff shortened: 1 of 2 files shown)


dnzxy (approved):
LGTM.

kingst (approved):
The code LGTM. I'm not that familiar with the NS API specs, but conceptually I understand what this is doing and it looks right.

One question I had though: is it possible to only include Trio entries vs excluding specific other systems? I'm not sure if there are other places we'd want to pull in NS values from so this is a low context question.

PR #553, opened 2025-05-21
Localize info log message notifications

kingst:
When we switched our logging system to use non localized errors, we inadvertently caused the notifications sent to also not be localized. This PR exposes a new log.info interface that allows callers to set the notificationText explicitly.

This issue was reported on discord by @kskandis 

As a side note: in my testing there were some cases of double notifications from the log.info call invoking a notification, but that is a separate issue. My goal is this PR is to restore the previous functionality, we can take on the question of whether our log.info message should send notifications later.

### DIFF
--- a/Trio/Sources/Logger/Logger.swift
+++ b/Trio/Sources/Logger/Logger.swift
@@ -40,6 +40,29 @@ func info(
     }.perform()
 }
 
+func info(
+    _ category: Logger.Category,
+    _ message: String,
+    notificationText: String,
+    type: MessageType = .info,
+    file: String = #file,
+    function: String = #function,
+    line: UInt = #line
+) {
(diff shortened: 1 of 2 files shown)


marionbarker:
## Test

Configuration:
* iPhone SE 2nd gen running iOS 18.5
* Running Trio 0.5.0.x with rPi DASH connected
* After triggering a DASH error (pod not connected) using ^C on the rPi, build using commit `02f10043c`, branch `fix-notification-localization`

### Stop rPi pod simulator to trigger an error

The graphic on the left is before, the graphic on the right is after building with the fix provided in this PR.

trio-pr553

marionbarker (approved):
Approve from test

dnzxy:
Merging with 2 approvals and successful test.
Thank you both.

PR #561, opened 2025-05-22
Fix for checking the presence of alive branches

### LINKED ISSUE #554: Fix alive branch logic in build_trio.yml
## Describe the bug
The logic in build_trio.yml skips creating an alive branch if either alive-dev or alive-main exists. So you can have a failure if building with `main` or `dev` without an associated `alive-xxx` branch.

| Build Trio with branch | does alive-main exist? | does alive-dev exist? | Result |
|:--|:--|:--|:--|
| `dev` | yes | no | [Failed](https://github.com/docs-test/Trio/actions/runs/15165437380) |
| `dev` | no | no | [Success](https://github.com/docs-test/Trio/actions/runs/15165547065) |
| `main` | no | yes | [Failed](https://github.com/docs-test/Trio/actions/runs/15166415084) |

## Attach a Log
N/A - see table above

## To Reproduce
See table above

## Expected behavior
*If building with `main` branch, check to see if `alive-main` exists*
*If building with `dev` branch, check to see if `alive-dev` exists*

OR

Modify the entire behavior now that the external repository gautamkrishnar/keepalive-workflow is not available.

## Screenshots
N/A

## Work Around

Edited to add: the work around is to delete any alive branch and build again.

bjornoleh:
build_trio.yml: check for the presence of alive-main and alive-dev independently, and create missing branches.
    

- Explicitly check for each branch independently.
- Ensure both conditions must be met (&&) for the environment variable to be set to true.
- Create only the missing ones.
- Avoid relying on a single flag (ALIVE_BRANCH_EXISTS), because that doesn't tell you which branch is missing.

Fixes #554 .

### DIFF
--- a/.github/workflows/build_trio.yml
+++ b/.github/workflows/build_trio.yml
@@ -55,29 +55,36 @@ jobs:
 
       - name: Check for alive branches
         if: steps.workflow-permission.outputs.has_permission == 'true'
+        id: check-alive
         env:
           GITHUB_TOKEN: ${{ secrets.GH_PAT }}
         run: |
-          if [[ $(gh api -H "Accept: application/vnd.github+json" /repos/${{ github.repository_owner }}/Trio/branches | jq --raw-output '[.[] | select(.name == "alive-main" or .name == "alive-dev")] | length > 0') == "true" ]]; then
-            echo "Branches 'alive-main' or 'alive-dev' exist."
-            echo "ALIVE_BRANCH_EXISTS=true" >> $GITHUB_ENV
+          branch_list=$(gh api -H "Accept: application/vnd.github+json" /repos/${{ github.repository_owner }}/Trio/branches | jq -r '.[].name')
+      
+          if echo "$branch_list" | grep -q '^alive-main$'; then
+            echo "alive-main exists"
+            echo "ALIVE_MAIN_EXISTS=true" >> $GITHUB_ENV
           else
-            echo "Branches 'alive-main' and 'alive-dev' do not exist."
-            echo "ALIVE_BRANCH_EXISTS=false" >> $GITHUB_ENV
+            echo "alive-main missing"
+            echo "ALIVE_MAIN_EXISTS=false" >> $GITHUB_ENV
+          fi
+      
+          if echo "$branch_list" | grep -q '^alive-dev$'; then
+            echo "alive-dev exists"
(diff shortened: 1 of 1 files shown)


bjornoleh:
Tested and verified to work (on a testing branch using a different bundle ID) here: https://github.com/nightscout/Trio/issues/554#issuecomment-2901879944

dnzxy (approved):
I would have probably preferred the removal of the entire `alive-xxx` workflow portions, as we no longer rely on it since removing the `keepalive` tooling, but this PR is surgical in its changes, tested, and the changes look more than good to me. 
Thanks, approving.

marionbarker:
## Test

### Configuration

* upload this branch to my `docs-test` *GitHub* organization
* rearrange branches there for this test
   * make `fix_check_alive_branches` the default branch
   * delete the `main` and `dev` branches and then make new versions of `main` and `dev` from this branch
   * delete the `alive-main` and `alive-dev` branches
   * configure `dev` (rename of `bjornoleh:fix_check_alive_branches`) as default
* test the build with the truth table below
   * the `alive-main` and `alive-dev` columns have a "y" if the branch exists before starting the build
   * the "y" in the success column means the build succeeded and both an `alive-main` and `alive-dev` branch are available after building starts
   * the link is for each successful build
* The fact that all builds below succeeded indicates this PR is a valid fix
* Eventually we might want to get rid of the alive branches, but at this point in time, do this minimal fix

### Test Results

| alive-main? | alive-dev? | build branch | success? | link |
|:--:|:--:|:--:|:--:|:--:|
| n | n | `main` | y | [Success](https://github.com/docs-test/Trio/actions/runs/15310152453) | 
| n | y | `main` | y | [Success](https://github.com/docs-test/Trio/actions/runs/15310470877) |
| y | n | `dev` | y | [Success](https://github.com/docs-test/Trio/actions/runs/15310794125)  |
| y | y | `dev` | y | [Success](https://github.com/docs-test/Trio/actions/runs/15311139162)  |

### After the test completes
* delete the fake `main` branch and add `main` from nightscout/trio
* set the real `main` branch as default branch
* delete the fake `dev` branch and add `dev` from nightscout/trio

### Final Test

Find out if I need to delete the `alive-dev` branch by running build with `dev` branch.
* Answer - no worries, the new build will be successful

marionbarker (approved):
Approve from code review and test.

PR #562, opened 2025-05-22
Add locking to our Swinject resolver

kingst:
In our Crashlytics logs we can see crashes from Swinject service resolution happening on multiple threads. This PR adds a simple locking mechanism to our resolver so that it is thread safe. The locking mechanism simply enforces the invariant that no two threads can access the resolver concurrently. By using a recursive lock, we can allow nested resolution on the same thread without impacting the safety of the data structures.


dnzxy (approved):
As discussed in detail via DM, this is good to go and was only waiting for extended testing to finish. Let’s go 🚀

marionbarker:
In case this is useful, here are the log files zipped together for this test phone. I will free the test phone for a different purpose soon.

[Logs_SE_3rd_gen_PR_562_2_days.zip](https://github.com/user-attachments/files/20432937/Logs_SE_3rd_gen_PR_562_2_days.zip)

PR #565, opened 2025-05-23
Swift File Headers

bjorkert:
This PR restore headers in swift files, before headers were stripped away.

Example of a header:
```
// Trio
// OverviewStepView.swift
// Created by Deniz Cengiz on 2025-04-21.
```

Headers are generated by https://github.com/nicklockwood/SwiftFormat. SwiftFormat was introduced in FreeAPS X and has been used since then. 

This PR also excludes DanaKit from SwiftFormat processing and bumps the SwiftFormat version. The new version does some more formatting compared to the old version, visible in the diff.


bjorkert:
100% agree, it is not something SwiftFormat has built in. Some of it might be possible to do with scripts, not sure how much it would affect build time.

MikePlante1:
Imho, headers are unnecessary as GitHub tracks the authors of all files and commits already anyway.

marionbarker:
I think blame is more useful to figure out who changed a line. We don’t need credit added to every file. Could we use: created by Trio authors?

dnzxy:
> I think blame is more useful to figure out who changed a line. We don’t need credit added to every file. Could we use: created by Trio authors?

I‘m fairly certain some of the contributors feel we do, so let’s not invalidate their take. May I suggest going with @marv-out ‘s proposal and add a line with 

>  © The Trio Authors

bjorkert:
Thanks everyone for the feedback. I understand there are differing views on file headers. My intention with this PR was to restore what was previously removed, align with existing SwiftFormat behavior, and include some minimal info about who initially added each file and when. I’m not planning to maintain custom formatting rules, so my suggestion is to either accept the standard headers as they are now – or skip headers entirely, as it was before. If the preference is to leave headers out, I can submit a separate PR just to bump the SwiftFormat version and exclude the DanaKit submodule from formatting.

dnzxy (approved):
Had a look into custom swift formatter rules and, as @bjorkert points out, it involves forming the formatter repo 🫠


Subsequently, approving as is, as this follows best practices and cleans up our current stairs who with some files that have headers and some files thar don’t.

I believe dedicated credit or "blame" (who did what when in which mine) will and is always available via git.

AndreasStokholm (approved):

Issue #527, opened 2025-05-18
Rename “Fatty Meal Option”


g5WS:
Some users may perceive the use of the word “fatty” as stigmatizing, judgmental, or inaccurate as it does not mention that protein can also slow down absorption.

Consider renaming this to “Slow-Absorbing Meal Option” or “Delayed Absorption Meal Option”. 

![Image](https://github.com/user-attachments/assets/a2423493-4157-4043-9f45-c8ce9075127a)

dnzxy:
Hi there, thanks for the suggestion.
I've gone ahead and relabeled this to an enhancement, as it's not really a bug. 

Do you happen to have any idea for this that is not ~3 words and 6-8 syllables? 
We need it short and sweet, if we are considering to change this :)

g5WS:
Slow Meal (Option)?

Slow Delivery (Option)?

FPU Option?

kristinkand:
Pizza option 


(here they call the slow digestion having a "pizza tail")

kingst:
I'm opposed to "Slow meal". It's ambiguous (e.g., does it mean that you plan to eat slowly?) and requires an extra mental step. I've seen "fatty" in the literature (e.g., fatty liver disease) but I can understand wanting to change it. How about:

High fat meal
Fat rich meal

g5WS:
> High fat meal 
> Fat rich meal

Hmm. Protein slows down absorption, too. And it would be nice if the label additionally worked for people with gastroparesis as well.

kingst:
Ok, fair point. Generalizing further, when I pre-bolus I'd like to bolus less upfront as well. UX is not really my area of expertise, but maybe we want something more direct:
- Bolus less
- Reduced bolus

bjornoleh:
> [@kingst](https://github.com/kingst) :
> 
> > High fat meal
> > Fat rich meal
> 
> Hmm. Protein slows down absorption, too. And it would be nice if the label additionally worked for people with gastroparesis as well.

Regarding gastroparesis, if this affects _every meal_, then the bolus % setting can be used instead.

bjornoleh:
> Ok, fair point. Generalizing further, when I pre-bolus I'd like to bolus less upfront as well. UX is not really my area of expertise, but maybe we want something more direct:
> 
> * Bolus less
> * Reduced bolus

I was initially in favour of something like «Slow absorbing meal» (although too long). 

[...]

PR #566, opened 2025-05-23
Rename Fatty Meal to Slow Meal

Lamercho:
Renames "Enable Fatty Meal Option" to "Enable Slow Meal Option" and "Fatty Meal Bolus Percentage" to "Slow Meal Bolus Percentage".

Closes https://github.com/nightscout/Trio/pull/566

### DIFF
--- a/Trio/Sources/Modules/BolusCalculatorConfig/View/BolusCalculatorConfigRootView.swift
+++ b/Trio/Sources/Modules/BolusCalculatorConfig/View/BolusCalculatorConfigRootView.swift
@@ -98,8 +98,8 @@ extension BolusCalculatorConfig {
                     ),
                     units: state.units,
                     type: .conditionalDecimal("fattyMealFactor"),
-                    label: String(localized: "Enable Fatty Meal Option"),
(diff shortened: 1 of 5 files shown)


Lamercho:
Because of the https://github.com/nightscout/Trio/issues/527

Sjoerd-Bo3:
Closes #527

Trio/Sources/Modules/BolusCalculatorConfig/View/BolusCalculatorConfigRootView.swift
Sjoerd-Bo3, line 1: I think you forgot this one 2 times in this file :)
"When \"Fatty Meal\" is selected

Trio/Sources/Services/BolusCalculator/BolusCalculationManager.swift
Sjoerd-Bo3, line 1: If changing this, Maybe we should also rename it in the variables? Or at least a comment at this function, that it before was named Fattymeal and now Slow meal. But better to change the name of the feature in all. Not only display.

kingst:
I don't think we should use "Slow meal" I left a comment on the issue with a few other suggestions.

PR #568, opened 2025-05-23
rename fatty meal to reduced bolus

Lamercho:
Fatty Meal changed to Reduced Bolus. 

Closes https://github.com/nightscout/Trio/issues/527#issuecomment-2905475012


Trio.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -1,5 +1,5 @@
 {
-  "originHash" : "89074a88ed67a58ecd7534519854c5a0928a4046d7c8a6123a7d70f27bf8b44d",
+  "originHash" : "19d4832d449c850d10dc8b15597db0825e41ce3f2a3b0d4b8471c7f66a7517dc",
```
kingst, line 2: Can you remove the changes to this file? Xcode will automatically change it, but in this change it looks like it removed tidepoolKit.

Trio/Sources/Localizations/Main/Localizable.xcstrings
```diff
@@ -86472,7 +86472,7 @@
         }
       }
     },
-    "Enable Fatty Meal Option" : {
+    "Enable Reduced Bolus Option" : {
```
kingst, line 86475: @t1dude can you help review the Localization changes? We should have these strings re-translated since we're updating the text. I think these are right, but I'm not 100% sure how Crowdin works.

kingst (changes requested):
Overall looking good. The code looks great and I tested it in the simulator -- it's working well. I agree with the decision to only update the user-visible strings and not variable names.

Two requests:
- Let's get someone, like @t1dude , to take a look to make sure that these strings will get translated
- There was a change to the Package.resolved file, which we should remove in this PR

tmhastings:
Once this is merged, can someone submit an issue to the trio-docs repo to update the text from "Fatty Meal" to "Reduced Bolus" in the docs?

kingst (changes requested):
With the latest changes your code removes the `Trio.xcworkspace/xcshareddata/swiftpm/Package.resolved` file. What I would do is something like this:

```
(1) Create a diff for the Package.resolved file
# git diff HEAD HEAD~1 -- Trio.xcworkspace/xcshareddata/swiftpm/Package.resolved > tmp.diff

(2) Apply the diff
# patch -p1 < tmp.diff

(3) Add the file back to git
# git add Trio.xcworkspace/xcshareddata/swiftpm/Package.resolved

(4) Commit the changes and push the updates to your branch
# git commit -a -m "Back out changes from Package.resolved"
# git push
```

You'll know that it worked if this file doesn't show up in the "Files changed" part of your PR.

Lamercho:
sorry, wrong button :-)

kingst (commented):
The updated text and extra modified file all look good. At this point we just need feedback on the strings / translation setup and then this one should be ready

t1dude:
Building now, updating string catalog and uploading to Crowdin, making sure the strings get translated.

t1dude:
New string catalog has been built, and uploaded to Crowdin. All translators have been notified. This PR looks good to me.

t1dude (approved):
- Code looks good
- Builds fine
- String catalog updates fine
- New and changed strings have been added to Crowdin and are ready for translation

dnzxy (approved):

PR #572, opened 2025-05-24
Custom file headers

marv-out:
As already mentioned in PR No. [565](https://github.com/nightscout/Trio/pull/565), I believe it would be fairer to generate a custom header for the files. I have done this with the following PR. I created a Bash script and added it to the build phases in Xcode. This script initially generated appropriate headers for all files in the following format:

```
//
// Trio
// HomeStateModel.swift
// Created by Deniz Cengiz on 2025-01-01.
// Last edited by Marvin Polscheit on 2025-05-24.
// Most contributions by Marvin Polscheit and Deniz Cengiz.
//
// Documentation available under: https://triodocs.org/
```

After that, I added a function that ensures only modified files get updated. This results in no loss of build time and keeps the headers always up to date—after building and before committing.

You can see how it updates when you edit a file and build:
goileHeaders

Additionally, I added a plist file located in the `xcshareddata` folder, which should ensure that every newly created file uses the same header format from now on. Once other people start working on a newly created file, it will then be updated via the update-header script...

(There is also some logic to replace my (several) different names with my actual one. This could be easily added in the script for other names as well...)


marionbarker:
## Test

I confirm the headers from this PR are not modified by building with Xcode on my Mac.

The 2 files that have my name associated with them as being last edited by me or list me as second most prolific contributor are correct.

### Modify a file test

* not a success

I then modified a file with a dummy comment and did not commit the change. This modification was made to the header when I built:

```
-// Last edited by Deniz Cengiz on 2025-03-27.
-// Most contributions by Deniz Cengiz.
+// Last edited by Marvin Polscheit on 2025-05-24.
+// Most contributions by Deniz Cengiz and Marvin Polscheit.
```

I'm not sure if I need to modify something in my local git config or if it picked @marv-out because of branch ownership, or if the custom script needs another modification.

marionbarker:
With this PR, we go from not having headers to having headers with statistics included. I do not think headers are the right place for tracking who is modifying what file.

marv-out:
> I'm not sure if I need to modify something in my local git config or if it picked @marv-out because of branch ownership, or if the custom script needs another modification.

Hmm I will have another look today

marv-out:
> With this PR, we go from not having headers to having headers with statistics included. I do not think headers are the right place for tracking who is modifying what file.
> 
> 

But I think just mentioning the original authors of the files is unfortunately not enough here, given the numerous refactorings that have taken place. You might as well leave out the headers entirely then, since they don't provide any added value, imo 🤷🏻‍♂️

Sjoerd-Bo3:
> My understanding of @marv-out ‘s take on this is that he does not want to take undue credit for new files that were created based on refactoring of existing ones.
> 
> Even for refactored files, the git blame seems to be complete, like this file that was created in January 1. 2025:
> 
> https://github.com/nightscout/Trio/blame/dev/Trio/Sources/Modules/Home/HomeStateModel.swift
> 
> And in the commit history, there’s a link to the file history before the renaming at the bottom. https://github.com/nightscout/Trio/commits/dev/Trio/Sources/Modules/Home/HomeStateModel.swift?after=68163f267e0cd6ce814452e00592bbcd7339f145+34
> 
> Question: does this hold also if parts of a file was copied to a new one?
> 
> One concern with the header scripting is that PR authors absolutely must build and commit before pushing changes to PRs, otherwise there will be uncommitted changes. But that happens regularly with Xcode linting too, and build before push isn’t the worst rule. But it does complicate making simple edits from GitHub.
> 
> I don’t know if authors must be mentioned in files, but I am looking forward to getting the file names in the headings, this will make it much easier to search for files in Xcode ( at least for me, the Xcode search does not return anything for file names without headers, maybe I missed something).

Thinking out loud here, if we would make this into a GH Action that runs on every PR? Would that solves this?

Sjoerd-Bo3:
@bjornoleh Claude thinks it can:
https://claude.ai/public/artifacts/3a790f42-8ecb-4b02-8022-10b0bc04f4b3
You're right to think about this carefully. Here's the breakdown of when headers should be updated and how to handle merges:

Here's what this version does to handle your concerns:

## When it runs:
- **After PR merges** to main/master/develop branches
- **Not on every PR** (to avoid noise and inaccurate data)
- **Not on feature branches** (only on your default branches)

## How it handles merges:
1. **Excludes merge commits** using `--no-merges` flag in git log commands
2. **Filters out bot authors** (GitHub Actions, dependabot)
3. **Uses git blame with copy detection** (`-M -C` flags) to track code movement accurately
4. **Ignores whitespace changes** (`-w` flag) in blame

## Key improvements:

1. **Infinite loop prevention**: 
   - Uses `[skip ci]` in commit message
   - Checks if the commit message contains "Update Swift file headers"

2. **More accurate attribution**:
   - Merge commits are excluded from author detection
   - Bot commits are filtered out
   - Blame uses copy/move detection to track code properly

3. **Better blame analysis**:
   ```bash
   git blame -w -M -C --line-porcelain
   ```
   - `-w`: Ignore whitespace changes
   - `-M`: Detect moved lines within file
   - `-C`: Detect copied lines from other files

## Workflow:
1. Developer creates PR
2. PR gets reviewed and merged to main
3. Action runs on the merge commit to main
4. Headers are updated with accurate attribution
5. Changes are committed with `[skip ci]` to avoid retriggering

This approach ensures:
- Headers reflect the actual merged state
- No pollution from unmerged PR data
- Accurate contributor attribution
- No infinite loops
- Merge commits don't affect attribution

marv-out:
Of course, there are many ways to solve this problem. But I think the current approach is actually quite good because only the changed files are updated. So it doesn’t increase the build time at all. Previously, I had the issue that my script always updated all the header files, which noticeably increased the build time.
Sure, it’s a bit inconvenient that you have to build before committing. But actually, as Bjorn already mentioned, you should always do that anyway, because otherwise changes might be introduced due to linting. So I don’t really see any issues with it.

Sjoerd-Bo3:
> Of course, there are many ways to solve this problem. But I think the current approach is actually quite good because only the changed files are updated. So it doesn’t increase the build time at all. Previously, I had the issue that my script always updated all the header files, which noticeably increased the build time. Sure, it’s a bit inconvenient that you have to build before committing. But actually, as Bjorn already mentioned, you should always do that anyway, because otherwise changes might be introduced due to linting. So I don’t really see any issues with it.

Only that GH Edits are not possible that way. But also that's minor ofc. And the efficiency is true, but that's ofc easy to fix (only address the files since last time running this action e.g last header commit). 

It's more that it is not 100% foul proof, but now you are able to see who didn't build their change prior to committing :P

PR #573, opened 2025-05-24
Adjustable interval for glucose simulation in simulator

marv-out:
This PR makes the interval at which new glucose values are generated adjustable for debugging purposes when the simulator is chosen. This way, you don’t have to wait 5 minutes or repeatedly make the same code changes—it can simply be adjusted directly within the simulator and it automatically updates the loop interval so that it can loop as well.

grafik
grafik

I've also added a debug flag, set the default to release mode and made sure that you need to build in debug mode to see the glucose + pump simulator. Building in relase mode doesn't let you choose the sims anymore:

grafik


marionbarker:
## Status

Quick look and test - this seems to operate as designed.

## Review

I did a quick review and the code changes looked reasonable.

## Test

I built this onto a phone previously using the CGM simulator and connected to an rPi DASH pod
* The main screen showed add CGM as expected.
* I then deactivated the pod, switched insulin delivery and observed that the pump simulator is not an option
* I observed that the CGM simulator is not an option

I then edited the Config file and modified HIDE_SIMULATOR to be 0 and rebuilt.
* I can now add Pump Simulator, and observe that it is on the bottom row
* I can now add CGM Simulator, and observe that it is on the bottom row

## Missing Element

I thought I posted this earlier but I don't see it in the comments. I think we need to be able to cover the full range of glucose values that can come from CGM and glucose meters.

For example - I would not be able to use this simulator to test the case where CGM values are 400.

## Consideration

Does this really need to be a build flag? I'm thinking of parents who might want to have a version of Trio on their own phone using the pump simulator. If this was handled by a switch on the Feature page, then they would not have to build 2 different versions - one for themselves and one for their kid.

marionbarker:
## Update

Setting a data interval less than 3 minutes prevents Trio from looping.

### Test Data Interval

Phone unlocked and app in foreground so Trio should loop even without a heartbeat. Leave phone operating at each setting for at least 15 minutes.

**Data Interval of 1 min - red loop**
* I have the phone configured for 1 minute CGM updates and the phone is not looping. Pressing the loop icon does not help.

**Data Interval of 5 min - green loop**
* Modify the CGM to 5 minute intervals. The app now loops. 

**Data Interval of 3 min - green loop**
* Modify to 3 minutes for CGM update interval. The app continues to loop.

**Data Interval of 2 min - red loop**
* Modify to 2 minutes for CGM update interval.

mountrcg:
@marv-out the bug @marionbarker mentioned above most probably has to do with the LoopIntervall, wouldn't it make more sense to make the 10s dependent on the pumpSim being used instead of the glucoseSim? I for instance never use the glucoseSim in simulator, but real values from NS.

Issue #528, opened 2025-05-18
Add toggle for external insulin to shortcut action


g5WS:
![Image](https://github.com/user-attachments/assets/27c09220-ef9a-41de-81e1-ce82f50f8f14)

It would be very helpful if this shortcut action could be extended to include a toggle to indicate that the insulin was delivered externally.

dnzxy:
Just adding `long-term` here, so the issue does not go stale and gets auto-closed by our bot.

idoodler:
Can't you just add Insulin via the Apple Health App?

MikePlante1:
> Can't you just add Insulin via the Apple Health App?
> 
> Edit: I just tested it. It appears as if Trio is not reading the Data from Apple Health. So my workaround is not possible.

Trio (by design) does not read anything from Apple Health.

MikePlante1:
This was added in #607 

https://github.com/user-attachments/assets/81c31b14-fb79-41eb-9ac6-aec8992f642b

PR #607, opened 2025-05-29
external insulin to the bolus shortcuts.

avouspierre:
Add toggle for external insulin to shortcut action #528.

### Behavior
If toggle, the Insulin will be added to IOB but it will not be delivered.
Not dependant of the settings configuration about the bolus available in shortcuts but maxBolus is checked. 
Translation is available 

|  English | French |
 | - | - |
 | image| image |

### DIFF
--- a/Config.xcconfig
+++ b/Config.xcconfig
@@ -19,7 +19,7 @@ TRIO_APP_GROUP_ID = group.org.nightscout.$(DEVELOPMENT_TEAM).trio.trio-app-group
 
 // The developers set the version numbers, please leave them alone
 APP_VERSION = 0.5.1
-APP_DEV_VERSION = 0.5.1.2
+APP_DEV_VERSION = 0.5.1.3
 APP_BUILD_NUMBER = 1
 COPYRIGHT_NOTICE =
 
--- a/Trio/Sources/Services/LiveActivity/LiveActivityManager.swift
+++ b/Trio/Sources/Services/LiveActivity/LiveActivityManager.swift
@@ -219,6 +219,11 @@ final class LiveActivityManager: Injectable, ObservableObject, SettingsObserver 
     ///
     /// - Parameter state: The new content state to push to the live activity.
     @MainActor private func pushUpdate(_ state: LiveActivityAttributes.ContentState) async {
+        if !settings.useLiveActivity || !systemEnabled {
+            await endActivity()
+            return
+        }
+
         if currentActivity == nil {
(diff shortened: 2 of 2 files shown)


dnzxy (changes requested):
The code generally looks good to me and it works (tested in-vitro).
I have two change requests:
1. External insulin when added in-app via treatment view is guarded by `maxBolus x 3`. Here we would allow any number. Please apply the same max bound check and an appropriate error, if a user attempts to enter say 5x max bolus insulin.
2. I would change the wording from "immediately applying" to something that conveys the nature of "you log this, but the pump does nothing, you have to inject / take the dose" in a more intuitive way. In the app we call this "Log External Insulin". Maybe using that phrase (once as a question "How much external insulin do you want to log?", once as a statement "Logging xxx U of external insulin.") may be clearer for users.

dnzxy:
Hi @avouspierre. Are you able to take a look at this PR anytime soon? If not, does someone else maybe want to take over and collaborate with Pierre? Maybe @dsnallfot or @bjorkert ?

avouspierre:
HI @dnzxy , 

Lately but a update of this PR with : 
- add guardrail about max bolus if external insulin as Trio (3 x max bolus)
- change the label "apply" ➡️ "log" when external insulin.

Trio/Sources/Shortcuts/Bolus/BolusIntentRequest.swift
```diff
@@ -27,4 +27,25 @@ import Foundation
...
+            )
+        } else {
+            bolusQuantity = apsManager.roundBolus(amount: Decimal(bolusAmount))
+            await pumpHistoryStorage.storeExternalInsulinEvent(amount: bolusQuantity, timestamp: Date())
+            // perform determine basal sync
+            try await apsManager.determineBasalSync()
+
+            return String(
+                localized:
+                "A external bolus of \(bolusQuantity.formatted()) U of insulin was recorded."
```
marv-out, line 47: AN external....

marv-out (changes requested):
Please solve the merge conflicts, add the missing "n" (see comment in BolusIntentRequest) and add a condition to the confirmation message that clearly distinguishes between the logging of external insulin and bolusing actual insulin. Currently if you enable the confirmation message for logging external insulin it also says "Are you sure to bolus x U of insulin"

So in fact two minor changes, then this is good to go 🚀

marv-out (approved):

PR #614, opened 2025-05-31
sync[DanaKit]: Add logs missing pumpDelegate & DanaRS-v1 disabled

bastiaanv:
Changelog:
- Changing the DanaRS-v1 to disabled state, since the encryption does not work
- Small rework for `UnfinalizedDose` -> Made the usage of it easier
- Small fix for `Stop temp basal` button -> Hitting the button did not show the spinner
- Adding extra logs when pumpDelegate is `nil`
  - One or two users are reporting that SMB's are missing in their Trio v0.5.0.14
  - DanaKit logs show that the bolus is enacted and successfully completed
  - Trio logs show `enactBolus(units:automatic:) - 1195 - DEV: Bolus succeeded: 0.25`
  - But no bolus is shown in the history -> I suspect that the `pumpDelegate` is empty, but added these log lines in order to proof this in the future :)


dnzxy:
> @bastiaanv : 
>   * One or two users are reporting that SMB's are missing in their Trio v0.5.0.14
>   * DanaKit logs show that the bolus is enacted and successfully completed
>   * Trio logs show `enactBolus(units:automatic:) - 1195 - DEV: Bolus succeeded: 0.25`
>   * But no bolus is shown in the history -> I suspect that the `pumpDelegate` is empty, but added these log lines in order to proof this in the future :)

We added logging to the completion handler that feeds our storage with nightscout/Trio-dev#498 — have you checked logs for that?

dnzxy:
Thanks for the added context and some more background info. 

I‘m okay with getting more logging in to pinpoint this issue, but would just like to note that if the pump manager were nil, the dose could not have been enacted previously (via APSManager when acting upon a determination) and the entire UI would not be showing a pump. There are many layers building upon a present device data manager pump manager instance being available. I‘m more so suspecting a race condition somewhere. 

I must admit that we keep mixing separate things / fixes for separate issues with single PRs. Maybe we could try and keep them small and single scopes going forward.

kingst (approved):
Adding more logging makes sense, but I agree with @dnzxy that there is likely something else going on. There are a number of async calls on the delegateQueue, perhaps one of these isn't returning?

Issue #610, opened 2025-05-30
Add Confirmation Dialogs for Stopping Overrides and Temp Targets in Adjustments View


dnzxy:
## Is your feature request related to a problem? Please describe.

The "Stop Temp Target" and "Stop Override" buttons in the Adjustments tab can be triggered accidentally due to their large tap areas and fixed position just above the tab bar. 

![Image](https://github.com/user-attachments/assets/ed5059a4-b403-4b37-a35d-809c7a3f4df7)

![Image](https://github.com/user-attachments/assets/4b4c564a-aeb5-4b8b-ae5e-1c9d44a97e32)

This makes them prone to unintended taps, especially when navigating between views. The Home view already includes a confirmation dialog for stopping these modes, but this safeguard is missing in Adjustments.

## Describe the solution you'd like

Add a confirmation dialog when tapping "Stop Temp Target" or "Stop Override" in the Adjustments tab, similar to the behavior in the Home view.

Example:
> Stop the Override <name>?  
> [Cancel] [Stop]

This would make the experience safer and more consistent.

## Describe alternatives you've considered

- Relying on user precision — not reliable on small screens.

## Additional context

These stop buttons are sticky just above the tab bar, increasing the chance of mistaken cancellation during navigation.

Issue was recently also brought up and discussed on Discord.

## Technical Details

- Reuse existing confirmation logic from the Home view.
- Wrap button actions with a `ConfirmationDialog`.
- Low complexity, good beginner issue.

## User Impact

Improves safety by preventing unintended cancellations. Enhances UI consistency across views, especially helpful for users on small screens or with accessibility needs.

Stampertina:
Thank you for this feature improvement!

PR #617, opened 2025-06-01
Add Confirmation Dialogs for Stopping Overrides and Temp Targets in Adjustments View

marv-out:
**Summary**
This PR addresses issue [#610](https://github.com/nightscout/Trio/issues/610) by adding confirmation dialogs before stopping Overrides or Temporary Targets in the Adjustments view, enhancing user experience and preventing accidental stops.

**Problem**
Previously, clicking the "Stop Override" or "Stop Temp Target" buttons in the Adjustments view would immediately stop the active override or temporary target without any confirmation. This could lead to accidental cancellations, which might disrupt the user's diabetes management.

**Solution**
The solution implements confirmation dialogs similar to those already present in the Home view:

**Benefits**

- Prevents accidental cancellation of overrides and temporary targets
- Provides clear feedback about which adjustment will be stopped
- Maintains consistency with the confirmation behavior already present in the Home view

### DIFF
--- a/Config.xcconfig
+++ b/Config.xcconfig
@@ -19,7 +19,7 @@ TRIO_APP_GROUP_ID = group.org.nightscout.$(DEVELOPMENT_TEAM).trio.trio-app-group
 
 // The developers set the version numbers, please leave them alone
 APP_VERSION = 0.5.0
-APP_DEV_VERSION = 0.5.0.28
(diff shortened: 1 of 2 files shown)


dnzxy (approved):
Tested with simulator and in-vitro. LGTM 🚀

PR #619, opened 2025-06-01
Replace logic for check for backdated carbs

marv-out:
This PR replaces the logic for the check for backdated carbs in `PopupView.swift` and `ForecastChart.swift` with the updated logic I've already used in line 395 in `calculateInsulin()` in `TreatmentsStateModel.swift`. I just forgot to use it there as well and the PR was merged before I got the chance to change this.

### DIFF
--- a/Trio/Sources/Modules/Treatments/TreatmentsStateModel.swift
+++ b/Trio/Sources/Modules/Treatments/TreatmentsStateModel.swift
@@ -391,8 +391,8 @@ extension Treatments {
                 simulatedCOB = min(maxCobInt16, cobInt16)
             }
 
-            // Check if this is a backdated entry by comparing with the default date
-            let isBackdated = date != defaultDate
+            // Check if this is a backdated entry by comparing with the default date using a tolerance
+            let isBackdated = abs(date.timeIntervalSince(defaultDate)) > 1.0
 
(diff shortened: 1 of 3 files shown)


MikePlante1:
I got this linting when building: Screenshot 2025-06-02 at 1 03 35 PM

For this test, my COB was 44g:



This PR seems to break the calculator:

| Treatment View | Details |
| --- | --- |
| IMG_7486 | IMG_7487 |
| IMG_7488 | IMG_7489 |

Here's 0.5.0.29:

| Treatment View | Details |
| --- | --- |
| 0g added | 60g added |
| IMG_7491 | IMG_7492 |

marv-out:
They don't break the calculator, but apparently the condition indeed failed at some point. I've debugged that and it seems that this was due to some minor difference between the two timestamps (because they are not initialized 100% at the same time). I've added a tolerance to the check and that seems to work....pls re-test @MikePlante1

MikePlante1 (approved):
Retested with the latest commit and it was successful. No linting now either.

| test | image |
| --- | --- |
| 3g COB + 0g added |  |
| 3g COB + 60g added to now |  |
| 3g COB + 60g backdated |  |

MikePlante1:
> Looking at the file changes, how can the proposed changes break the calculator?

I meant the calculator *<details sheet>, sorry.

Issue #529, opened 2025-05-18
Run Unit Tests as Part of Trio's CI/CD Setup for Pull Requests


dnzxy:
## Is your feature request related to a problem? Please describe.

Currently, unit tests for Trio are not automatically executed during the CI/CD process when changes are pushed to a pull request. This increases the risk of regressions being merged into the codebase and delays feedback on broken tests, especially if contributors miss or forget to run unit tests before merging pull requests.

## Describe the solution you'd like

Automated unit test execution integrated into CI/CD for Trio. Specifically, unit tests should run automatically whenever changes are pushed to a pull request. This could be implemented using GitHub Actions or an external tool such as CircleCI. Test results should be visible in the PR checks to inform reviewers of test pass/fail status before merging.

## Describe alternatives you've considered

* Manually running tests before pushing changes (not scalable or enforceable across all contributors).
* Relying on local CI scripts or documentation to guide developers (still prone to human error and oversight).
  The proposed solution ensures consistency, improves code quality, and reduces the burden on maintainers.

## Additional context

This request aligns with best practices for open source development and would help ensure continued reliability and maintainability of the Trio codebase. It also supports more robust collaboration by providing immediate feedback to contributors through CI, when they create a PR.

## Technical Details

* The test runner should support all existing unit test targets in the Trio Xcode project.
* If using GitHub Actions, a workflow file should be configured to install dependencies, build the project, and run tests on macOS runners.
* If using CircleCI or another external tool, secure environment configuration and macOS support must be considered.
* Cache dependencies where possible to reduce build time.
* Fail the CI job if any test fails.

## User Impact

Automated testing ensures a higher level of quality and reliability for all users. It helps prevent bugs from reaching production and reduces testing overhead for contributors, ultimately improving confidence in new releases.

dnzxy:
@aidanlane, I believe @bjornoleh is already working on this one. I thought he had "taken" and assigned the ticket, I just did that. Apologies. If you guys could get in touch and see if anything else is needed ✊

aidanlane:
Thanks, done ✅

PR #625, opened 2025-06-03
Automated unit tests on pull and PR as a GitHub action

bjornoleh:
Create unit_tests.yml: Automated unit tests on push and PR to main and dev

- Adds automated unit tests on push and PR to `main` and `dev`.
- Runs Trio Tests, but currently not Watch App tests or LoopKit tests
- Annotates the result and fails on failing tests.
- Before merging it will be changed to only run for repository owner `nightscout` (Todo: revert 97b1c44a25dcd064cf0ad3be6998301bd1e35273)

Attempts to solve https://github.com/nightscout/Trio/issues/529

Tested:

- [x]  Runs on push and PR
- [x] Fails on failing tests
- [x] Emits annotations that indicate that tests failed:

TODO:

- [ ] Revert 97b1c44a25dcd064cf0ad3be6998301bd1e35273 before merging so that the workflow only runs for repository owner `nightscout`
- [ ] Add conditionals for `Flaky tests`? One or more tests currently typically fail (mostly `BolusCalculatorTests.testHandleBolusCalculationFunction()`)
- [ ] Add  Watch App tests or LoopKit tests?

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -1,0 +1,67 @@
+name: zzz [DO NOT RUN] Automated unit tests
+
+on:
+  workflow_dispatch: # TODO: remove this after testing
+  pull_request:
+    branches:
+      - dev
+      - main
+  push:
+    branches:
+      - dev
+      - main
+
+jobs:
+  test:
+    name: Run Unit Tests
+    runs-on: macos-15
+    if: github.repository_owner == 'nightscout'
+
+    steps:
+      - name: Select Xcode version
+        run: sudo xcode-select -s /Applications/Xcode_16.3.app/Contents/Developer
+
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+          submodules: recursive
+
+      - name: Build for testing
+        run: |
+          set -o pipefail && \
+          time xcodebuild build-for-testing \
+            -workspace Trio.xcworkspace \
+            -scheme "Trio Tests" \
+            -destination 'platform=iOS Simulator,name=iPhone 16,OS=18.4' \
+      - name: Run tests
+        run: |
+          set -o pipefail
+          time xcodebuild test-without-building \
+            -workspace Trio.xcworkspace \
+            -scheme "Trio Tests" \
+            -destination 'platform=iOS Simulator,name=iPhone 16,OS=18.4' \
+          2>&1 | tee xcodebuild.log
+
+      - name: Annotate test results
+        if: always()
+        run: |
+          if [ -f xcodebuild.log ]; then
+            if grep -q "Failing tests:" xcodebuild.log; then
+              echo "::error title=Unit Tests Failed::Some tests failed"
+              echo "## ❌ Some tests failed:" >> $GITHUB_STEP_SUMMARY
+              grep -A 20 "Failing tests:" xcodebuild.log | \
+                grep -E '^\s+[A-Za-z0-9]+\..+\(\)' | \
+                sed 's/^/  - /' >> $GITHUB_STEP_SUMMARY
+              echo "::group::Failed Test List"
+              grep -A 20 "Failing tests:" xcodebuild.log | \
+                grep -E '^\s+[A-Za-z0-9]+\..+\(\)' | \
+                sed 's/^/  - /'
+              echo "::endgroup::"
+            else
+              echo "::notice title=Unit Tests Passed::✅ All tests passed"
+              echo "✅ All tests passed" >> $GITHUB_STEP_SUMMARY
+            fi
+          else
+            echo "::warning::Test log (xcodebuild.log) not found"
+          fi
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -15,7 +15,7 @@ jobs:
   test:
     name: Run Unit Tests
     runs-on: macos-15
-    if: github.repository_owner == 'nightscout'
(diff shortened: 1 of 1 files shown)


.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,67 @@
...
+jobs:
+  test:
+    name: Run Unit Tests
+    runs-on: macos-15
+    #if: github.repository_owner == 'nightscout' # TODO: remove comment after testing
+
+    steps:
+      - name: Select Xcode version
+        run: sudo xcode-select -s /Applications/Xcode_16.3.app/Contents/Developer
+
```
dnzxy, line 23: I would suggest to introduce caching here, to improve the runtime of the overall workflow.
Example:
```yaml
- name: Cache DerivedData and .build
        uses: actions/cache@v4
        with:
          path: |
            ~/Library/Developer/Xcode/DerivedData
            .build
          key: ${{ runner.os }}-trio-${{ hashFiles('**/*.swift') }}
          restore-keys: |
            ${{ runner.os }}-trio-
```

.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,67 @@
...
+            -destination 'platform=iOS Simulator,name=iPhone 16,OS=18.4' \
+      - name: Run tests
+        run: |
+          set -o pipefail
+          time xcodebuild test-without-building \
+            -workspace Trio.xcworkspace \
+            -scheme "Trio Tests" \
+            -destination 'platform=iOS Simulator,name=iPhone 16,OS=18.4' \
+          2>&1 | tee xcodebuild.log
+
```
dnzxy, line 45: I would suggest to upload the build artifact on failure. 
Example
```yaml
- name: Upload xcodebuild.log on failure
        if: failure()
        uses: actions/upload-artifact@v4
        with:
          name: xcodebuild-log
          path: xcodebuild.log
```

.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,67 @@
...
+  workflow_dispatch: # TODO: remove this after testing
+  pull_request:
+    branches:
+      - dev
+      - main
+  push:
+    branches:
+      - dev
+      - main
+
```
dnzxy, line 13: I would suggest to add a couple `path-ignore` patters here, e.g.
```yaml
paths-ignore:
      - '**.md'
      - '**/README'
      - '**.yml'
      - '**.txt'
```
for which we do not need to run swift-based unit tests.

dnzxy (commented):
Looking great! I have some suggestions, feel free to address.

`xcodebuild` also seems to offer `test -parallel-testing-enabled YES`. Have you had any success with that or tried that? It should be enabled by default (I think), just wondering, if we should maybe enhance # of workers, or if it's actually running with that.

We could also select a smaller-footprint simulator, like `'destination=platform=iOS Simulator,name=iPhone SE (3rd generation),OS=18.4'`, as this one has a smaller memory footprint and boots quicker than an `iPhone 16`.

I am a bit worried about the runtime of these unit tests at about 9-11min per entire suite run.

.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,67 @@
+name: zzz [DO NOT RUN] Automated unit tests
+
+on:
```
dnzxy, line 3: Question for these event hooks: will this not get triggered by the same commit push when the PR is open and you then push to it, as we are not limiting the `pull_request` events? Should we limit this? Not sure.

Something like below would limit runs trigger by the `pull_request` event 🤔 
```yaml
pull_request:
    types: [opened, synchronize]
 ```

bjornoleh:
Most of the suggestions have been adressd in the last commit:

> Add caching and GitHub Actions optimizations to unit test workflow
Added cache restore and save steps for Xcode DerivedData and .build to improve CI performance.
> 
> Limited workflow triggers to relevant file changes using paths-ignore for PRs and pushes.
> 
> Narrowed PR events to opened and synchronize only.
> 
> Added steps to display and check cache contents before and after the build.
> 
> Included a check for uncommitted changes with GitHub warnings for visibility.
> 
> Enabled optional parallel test execution using ENABLE_PARALLEL_TESTING flag.

Changing simulator device made no definite difference to the time of the test (2-3 minutes), so I did not include this here.

Caching made a meaningful difference to the build time.

- First run after merging of dev, relevant cache not available (build time 9:58): https://github.com/bjornoleh/Trio/actions/runs/15452289321/job/43497045706
- Second run with cache available (Cache Size: ~1205 MB):  https://github.com/bjornoleh/Trio/actions/runs/15452570278

I am unsure about whether or not the parallel testing makes a difference or not (or if it is set up correctly). It currently requires adding the repo variable ENABLE_PARALLEL_TESTING = `true` to utilise this, as it might add to the flakiness of tests. I did not see any evidence of that yet.

There's a check for uncommitted changes included. I have set it to not fail on uncommittet (linting) changes, but should emit a warning annotation (not tested)

.github/workflows/unit_tests.yml
dnzxy: @bjornoleh Are you looking to add this upload or leave it as-is? Just asking to know when you’re done with changing up things to take a final look at it ☺️

dnzxy:
I'm going ahead and merging this.
Let's look at adjusting and possibly introducing a flaky test conditional flag in a subsequent, related PR.
Thank you for looking into this @bjornoleh 🙏

dnzxy (approved):

PR #632, opened 2025-06-04
Release v0.2.6 — Browser Build Fix + Code Owners Update [to main] 

dnzxy:


bjornoleh (approved):
Built to confirm all is good. Nothing in here that should suggest otherwise though :-)

PR #653, opened 2025-06-11
Fix override banner in Live activity

marv-out:
I've noticed that the LA banner disappears in the following situation: 

- start an Override preset -> banner is there as it is supposed to be
- now edit the running override (not the preset) and save -> banner is gone

Debugging this showed me that we are 

1. missing the ID when we are performing our copy override logic and
2. (the "actual fix) are setting the same date as the copied override. This leads to the fetch only showing the first (now) disabled override that we've copied previously. And because of the fact that its disabled, it isn't shown in the UI. Also updating the date to `Date()` fixes this as this override will now be the latest override.

I really hope that this PR doesn't break anything related to overrides, as this logic is pretty well tested and works. I don't think that this small change will cause any issues, but nevertheless, it needs to be tested thoroughly.

### DIFF
--- a/Trio/Sources/APS/Storage/OverrideStorage.swift
+++ b/Trio/Sources/APS/Storage/OverrideStorage.swift
@@ -172,13 +172,14 @@ final class BaseOverrideStorage: @preconcurrency OverrideStorage, Injectable {
     /// otherwise we would edit the Preset
     @MainActor func copyRunningOverride(_ override: OverrideStored) async -> NSManagedObjectID {
         let newOverride = OverrideStored(context: viewContext)
+        newOverride.id = override.id
         newOverride.duration = override.duration
         newOverride.indefinite = override.indefinite
         newOverride.percentage = override.percentage
         newOverride.smbIsOff = override.smbIsOff
         newOverride.name = override.name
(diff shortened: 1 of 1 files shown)


dnzxy:
I understand why you are added an actual override date when the new override gets stored (based on your description of the issue and the intended fix). Why are you removing the duration helper function? No longer needed or mistakenly removed, @marv-out ?

MikePlante1 (dismissed):
Tested this in a simulator. I confirmed it fixes the problem of the OR banner in LA being hidden when a running OR is edited. I did not notice any adverse effects.

I also started a preset 10 min OR, edited the target after 6 minutes, and it expired 4 minutes later as expected.

I'll note that the Adjustments History page only shows the "Custom Override" for the full duration of the OR and does not show the initial preset, but that's the same as it is in current `dev`, so not anything introduced by this PR.

marv-out:
Ok I think this makes sense from looking at my changes. Hmmm, will have another look later how to avoid this behaviour

kingst (approved):
LGTM based on code review

dnzxy:
@MikePlante1 have you had the chance to test this again based on the comment by Marvin from July 2nd?

PR #659, opened 2025-06-13
hotfix/dev fastlane 2.228.0

marionbarker:
## Action of this PR

The Gemfile.lock file is updated to use fastlane 2.228.0

## Reason for this PR

There were changes at Apple that broke some of the code (fastlane) we use with GitHub Actions to enable Loop to build from a browser.

People who already had valid certificates with properly configured Identifiers could build without issue. Those who needed to renew certificates were blocked from building.

People who wanted to configure to build for the first time could not run Add Identifiers.

## Partial Solution

This update fixes these situations
* people who already use GitHub Actions (aka Browser Build) and need to have their certificates updated can continue to build
* people who switch to a different GitHub username (or organization) but use the same Apple Developer ID can build from the new username

The portion of the GitHub action used for new builders that enable them to Add Identifiers with the required capabilities select is still not working. However, if those Identifiers are manually created and configured, then the user can build Loop.


MikePlante1 (approved):
It's been a while since I've used Browser Build in my mikelnl-org account, so I had to update the following:
* Create a Fastlane Key and add it to my org secrets
* Add Nuke Certs to my org variables
* Agree to the new PLA

My identifiers were already good from building Trio 0.5.0 to TestFlight from Xcode, though.

Approving this PR, as this build succeeded: https://github.com/mikelnl-org/Trio/actions/runs/15639794064/job/44064258838

PR #662, opened 2025-06-15
Fix for Data race and build adaptations for iOS 26

marv-out:
This pull request addresses a data race issue in the `fetchForecastHierarchy()` function within the BaseDeterminationStorage class. Additionally, it swaps the SwiftUICore library with SwiftUI in some files. This popped up as build errors when I tried to build on a iOS26 simulator and this shouldn't destroy anything. So I changed it as well. Building on a iOS 18 device still works. The other changes were automatic changes made by xcode.

### DIFF
--- a/Trio/Resources/InfoPlist.xcstrings
+++ b/Trio/Resources/InfoPlist.xcstrings
@@ -457,6 +457,18 @@
         }
       }
     },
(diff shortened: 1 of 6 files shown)


kingst (approved):
I don't have an iOS 26 device but the code looks good

Issue #547, opened 2025-05-21
Add snooze function to "no pod" alert


g5WS:
Knocked a pod off, so I stopped the current pod so Trio wouldn't think it was delivering insulin.  Received this message every 5 minutes until the pod was replaced.

Suggestion: add ability to snooze this alert.

![Image](https://github.com/user-attachments/assets/7d0d8884-c570-4eb2-b65f-b58299f591ae)

dnzxy:
We cannot snooze this alert, as it’s thrown in a different layer (submodule) when the app attempts to loop. We may be able to try and surprise looping, but this may have a lot of negative side effects. 

For some added context: Was the last recorded pump event when this happened a pump suspend?

g5WS:
@dnzxy 

> For some added context: Was the last recorded pump event when this happened a pump suspend?

It was a Trio-initiated 0.0 u/h, not a user-initiated suspend, yes.

dnzxy:
So there was a temp basal running while you had the pod knocked off, but you did not hit "Replace pod" just yet in the pump manager UI, so the pod had not been discarded from Trio yet and the pump suspend had not been issued yet? 

Sorry for asking with such detail, just trying to think of something to address this. Knowing if there was a pump suspend present as the last recorded pump event would help.

g5WS:
- Trio had set a 0.0 u/hr temp rate because I was exercising and trending low
- Pod cannula wiggled loose
- I did hit “Replace pod” to prevent Trio from bolusing and recording insulin that would not be delivered subcutaneously
- Shortly after, the notifications began.

dnzxy:
Okay. You got these notifications because the app attempts to loop, and the pump driver tells the app "I cannot loop, there is no pod" (let’s ignore the missing localization for now, different issue). 

So you did tap Replace Pod and discarded the bumped pod via this UI. 

![Image](https://github.com/user-attachments/assets/f6012217-d664-4769-acc2-67094a111bf0)

This should have triggered a pump suspend event (happens automatically, not your manual doing),   so the last recorded entry at that time, i.e. "latest" pump event, should have been a `PumpSuspend`. 

Do you happen to use Nightscout to be able to check this in the Treatment Reports? Or do you still happen to have your logs on device from
when that happened?

g5WS:
@dnzxy 

![Image](https://github.com/user-attachments/assets/1fd4a650-0d9b-4622-9c5c-c082be292f95)

g5WS:
[...]

PR #663, opened 2025-06-15
Fix "No pod paired" notification when Omnipod is empty

marv-out:
**Problem**
When the Omnipod pod is empty, a misleading "No pod paired" notification is displayed as described in #547. But we don't want to show a message in those cases. 

**Root Cause**

- When the reservoir is empty, the pod sets a fault state (faultEventCode = 0x18)
- This causes podState.isActive to return false, as isActive only returns true when no fault is present
- When isActive is false, hasActivePod also becomes false
- This leads to operations failing with OmniBLEPumpManagerError.noPodPaired
- The current implementation in enactDetermination() only checks for suspension, not for faults or empty reservoir

**Solution**

- Added import of OmniBLE to access pod-specific types
- Implemented direct checking for pod faults in enactDetermination()
- Added specific handling for empty reservoir fault (code 0x18)
- When an empty reservoir is detected, the function logs appropriately and returns without error
- This prevents the misleading "No pod paired" notification while still correctly handling the empty reservoir condition

### DIFF
--- a/Trio/Sources/APS/APSManager.swift
+++ b/Trio/Sources/APS/APSManager.swift
@@ -3,6 +3,7 @@ import CoreData
 import Foundation
 import LoopKit
 import LoopKitUI
+import OmniBLE
 import SwiftDate
 import Swinject
 
@@ -666,6 +667,20 @@ final class BaseAPSManager: APSManager, Injectable {
         if pump.status.pumpStatus.suspended {
             info(.apsManager, "Skipping enactDetermination because pump is suspended")
             return // return without throwing an error
+        }
+
+        // Directly check if the pod is faulted
+        if let omnipod = pump as? OmniBLEPumpManager, let podState = omnipod.state.podState {
+            if podState.isFaulted, let fault = podState.fault {
+                // Check if this is an empty reservoir fault (0x18)
+                if fault.faultEventCode.rawValue == 0x18 { // reservoirEmpty
+                    info(
+                        .apsManager,
+                        "Skipping enactDetermination because pod reservoir is empty"
+                    )
+                    return // return without throwing an error - empty reservoir is expected
+                }
+            }
         }
 
         // Unable to do temp basal during manual temp basal 😁


dnzxy:
Let's add a `FIXME` comment to this change please. It looks okay, it works (tested in-vitro), but this goes so deep into a specific pump manager's notification behavior, this seriously should be handled in a different way. We will have to revisit pump manager notifications and interfacing on a larger scale for that though.

Trio/Sources/APS/APSManager.swift
```diff
@@ -668,6 +669,20 @@ final class BaseAPSManager: APSManager, Injectable {
             return // return without throwing an error
         }
 
+        // Directly check if the pod is faulted
```
dnzxy, line 672: Let's add a `FIXME` comment to this change please. It looks okay, it works (tested in-vitro), but this goes so deep into a specific pump manager's notification behavior, this seriously should be handled in a different way. We will have to revisit pump manager notifications and interfacing on a larger scale for that though.

Trio/Sources/APS/APSManager.swift
```diff
@@ -668,6 +669,20 @@ final class BaseAPSManager: APSManager, Injectable {
...
+        if let omnipod = pump as? OmniBLEPumpManager, let podState = omnipod.state.podState {
+            if podState.isFaulted, let fault = podState.fault {
+                // Check if this is an empty reservoir fault (0x18)
+                if fault.faultEventCode.rawValue == 0x18 { // reservoirEmpty
+                    info(
```
kingst, line 677: I believe that this will result in a notification each time we try to loop and the pump is empty. Is this the expected behavior?

kingst (commented):
Code LGTM, I don't have a good way to test this though. I left one question...

Trio/Sources/APS/APSManager.swift
dnzxy: The opposite. It skips the notification when the pod is empty and we attempt to enact the determination so you don’t get the frequent no pod alerts. 

Thinking about this some more I am not exactly sure if this PR will work in general. When there is no pod paired, there is no reservoir fault either, or is there @marv-out ?

kingst: `info` logs show an alert: https://github.com/nightscout/Trio/blob/23fe974bc9d1d1fa2882be3f39000e5982924c28/Trio/Sources/Logger/Logger.swift#L287

marv-out: Ok I thought this will produce such a yellow log in the console. I will change it then

dnzxy: But it doesn’t, according to this comment [here](https://github.com/nightscout/Trio/issues/547#issuecomment-2973323909).

The issue may not be rooted here in enacting determine basal, but where notification is thrown/created.

itsmojo:
Adding another import of OmniBLE and digging around in Pump Manager internals like this is not a good thing to do for a # of reasons and the proposed code has various issues of concern anyways (e.g., we’ll need an additional similar hack to handle Eros pods and eventually yet another [probably different] hack for Omnipod 5, we shouldn’t be using a hardcoded 0x18 value, etc). I’d really like to work towards removing references to PodState outside of the Omni PM’s and not add yet another case of Trio looking at PM internals like this. FYI, PodState is only declared as public because it was needed for OmniKit and OmniKitUI and now unfortunately its incorrect use has been expanding into iAPS/Trio [but not in Loop which supports plugin Pump and CGM managers and never looks at PodState or any other internal PM data like this which is needed to have plugin PM support] for things like this instead of using & possibly improving the existing PM interfaces which is the better approach. I was able remove the public PodState OmniXXX declarations in Loop and would like to be able to eventually do this in Trio as well.

The Loop app uses the PM independent buildPumpHighlight() to handle pod fault matters smoothly at the HUD. For OmniXXX, this func returns an appropriate localized “No Insulin”, "Pod Expired", "Pod Occlusion" string for these particular [expected] pod fault types and “Pod Error” for all other [unexpected] pod faults while a localized “No Pod” string is returned only when appropriate at the higher level (i.e., there is no PodState independent of things like a pod being partially activated, faulted, etc). I’m not familiar with the details of APS/Trio for this stuff, but at a high level it seems that APS/Trio should move towards a model where it to uses pump independent PM state/funcs (e.g., by using the status.basalDeliveryState variable &/or buildPumpHighlight()) to determine if a pump is available and if it is not, just don’t run the algorithm that will just fail trying to make any insulin delivery changes anyways. A potential bonus in taking this approach is that Trio wouldn’t generate annoying no pod/pump error messages during pod activation or pump reservoir replacement. Additionally, buildPumpHighLight() can be used to return a generic PM independent, localized, user friendly, and more appropriate string to display whenever the pump is not available.

Unfortunately I won’t be available for consulting for this particular issue for a while as I’ll be away traveling until July and I’ll have some more pressing matters to deal with on my return. If no one can see how to update the higher level Trio code to improve the Trio handling on an empty reservoir pod fault w/o adding more Omni imports and digging into the PodState details, I’d suggest tabling this issue until later. Let’s not add another hack just to handle this particular case since an appropriate (albeit not the ideal) notification is already being displayed.

dnzxy:
That’s exactly the issue: the notification being displayed over and over and over again, @itsmojo . You’re missing the point with your comment. We *are* simply forwarding the highlight message. The issue is that we keep calling the pod in a state of pod suspend when a pod was deactivated, which results in constant notifications of no pod paired. Essentially the app keeps attempting to loop with no paired pump, hence notification spam. 
I agree with this ideally not being a hacky thing, as I wrote I really dislike digging deep into the pump manager, but it’s an apparent issue and it needs solving, rather now and not in August.

dnzxy:
I'll close this one for now. Let's revisit this after we've moved in the latest OmniXXX updates and with a fresh mind @itsmojo @marv-out .

Issue #621, opened 2025-06-02
Overrides end spontaneously after approx. 2 days


christophkrall19:
## Describe the bug
A permanently set override Gets ended automatically without a meaningful reason. 

## Attach a Log
[log.txt](https://github.com/user-attachments/files/20550897/log.txt)
[log_prev.txt](https://github.com/user-attachments/files/20550896/log_prev.txt)

## To Reproduce
*Steps to reproduce the behavior:*
1. *Go to '...'*
2. *Click on '....'*
3. *Scroll down to '....'*
4. *See error*

## Expected behavior
*A clear and concise description of what you expected to happen.*

if override is activated permanently, it should remain on forever, no?

### Smartphone:
* Hardware: iPhone 14
* OS Version: iOS 18.4

### Pump:
* Manufacturer: Insulet
* Model: Omnipod Dash 

### CGM:
* Device: Dexcom G7
* Manager app: Dexcom App 

### Trio Version:
* Version Number: 0.5
* Repo: *nightscout/trio*
* Git Reference: *[e.g. commit hash]*

christophkrall19:
[log.txt](https://github.com/user-attachments/files/20550897/log.txt)
[log_prev.txt](https://github.com/user-attachments/files/20550896/log_prev.txt)

dnzxy:
Hi @christophkrall19 , thanks for filing this bug report.

> if override is activated permanently, it should remain on forever, no?

Technically, an "indefinite" override runs a full month. 

Could you please also provide:
- detailed steps to reproduce, i.e. exactly the input / entries you do for this to happen
- the specific version of Trio you have built (looking for the 0.5.0**.X**, the `.X` part) or the commit hash (that is the alpha numerical string next to the branch name when you tap Trio settings)

Thanks :)

christophkrall19:
I realized however, that there is no correlation with a month ending or so - it just quits the override after approx. 2 days without an obvious reason. 

Here the further info you requested: 

- detailed steps to reproduce, i.e. exactly the input / entries you do for this to happen: I normally enable the override by adding an override - setting the Basal to e. g. 130% and do not change any other setting in the dialog. So there is only a change in basal rate and the activation time is infinitely activated. 
- Version is V0.5.0.18 (10) where the branch is dev 9221da1

LiroyvH:
I can confirm my override just spontaneously ended after +/- 2 days as well. Because of sickness I've had an 125% override active with the following configuration:

Name: Insensitive
[...]

PR #664, opened 2025-06-15
Fix Overrides being ended after 2 days

marv-out:
Currently, it seems that indefinite Overrides are ended after >24h as described in #621. This PR aims at fixing this behaviour by editing the NSPredicate that we are using to fetch th latest active Override to exclude the date check for indefinite overrides and instead check for the `indefinite` property.

### DIFF
--- a/Model/Helper/OverrideStored+helper.swift
+++ b/Model/Helper/OverrideStored+helper.swift
@@ -7,10 +7,12 @@ extension NSPredicate {
     }
 
     static var lastActiveOverride: NSPredicate {
-        let date = Date.oneDayAgo
-        return NSPredicate(
-            format: "date >= %@ AND enabled == %@",
-            date as NSDate,
+        // For non-indefinite overrides, we still want to filter by date
+        // For indefinite overrides, we want them regardless of date
+        NSPredicate(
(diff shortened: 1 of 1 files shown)


MikePlante1 (approved):
Installed this branch, started an indefinite override, and after 48 hours it was still running. It still displayed on the home chart, the home banner, and the last determination HUD showed the adjusted target and ISF from the override.

Profile target is 92 mg/dL, Profile ISF is 60 mg/dL/U

| 3/15 | 3/15 |
| --- | --- |
| IMG_7618 | IMG_7619 |
| 3/17 | 3/17 |
| IMG_7635 | IMG_7636 |

MikePlante1:
Some remaining shortcomings, that could be fixed in a future PR:

* Once cancelled, the OR tape disappears from the chart
* Nightscout only shows the OR tape until midnight of the day it was enacted (before, as well as after, cancelling the OR)


Screenshot 2025-06-17 at 15 11 06

MikePlante1:
> > * Once cancelled, the OR tape disappears from the
> 
> Quite certain we had that issue before and it was fixed in closed beta, which would point at this PR for possibly reintroducing it.

This was an issue immediately before this PR as well. I’m just talking about infinite overrides after 24 hours, though.

PR #673, opened 2025-06-21
fix: Fix Swift 6 error in startBackgroundTask method

marv-out:
This PR fixes the Swift 6 error `Mutation of captured var 'taskID' in concurrently-executing code; this is an error in the Swift 6 language mode` in the startBackgroundTask method


kingst (approved):
I'm not sure how to test this, but the change LGTM.

PR #674, opened 2025-06-21
Optimize watch managers to skip processing when no watch is connected

marv-out:
# Watch Manager Optimizations

This PR improves app performance by preventing unnecessary processing when no watch is connected. It addresses two key issues:

### Problem
The app was previously creating tasks and performing expensive calculations for watch data even when no watch was connected or the watch app wasn't installed.

### Solution
1. Added early-return guards in both watch managers:
   - `BaseWatchManager.setupWatchState` now checks if a watch is paired, the session is reachable and the app is installed
   - `BaseGarminManager.setupGarminWatchState` now checks if any Garmin devices are connected

2. Optimized CoreData publishers to prevent task creation:
   - Added watch connectivity checks in all publisher callbacks
   - Tasks are only created when watches are actually connected
   - Added debug logs to indicate when processing is skipped

## Performance Impact
These changes can potentially reduce CPU and memory usage when no watch is connected, as the app no longer:
- Creates unnecessary background tasks
- Performs expensive CoreData fetches
- Builds watch state objects that would never be sent

## TODO
- [ ] Verifying that watch functionality works normally when watches are connected
- [x] Confirmed that unnecessary processing is skipped when no watch is connected
- [ ] Testing both Apple Watch and Garmin devices


dnzxy:
@t1dude (Apple watch), @mountrcg (Garmin), you both are avid watch users. Could you please give this a test, especially you @t1dude and see that this does no longer do any calls and updates to the watch manager, while keeping functionality (and especially sync after staleness) intact.

t1dude:
...testing in vivo for a few days.

marv-out:
Ok. So we still need to confirm that apple watch works properly when connected without any dropouts. This was @dnzxy worry from looking at the guards

mountrcg:
@marv-out I have that PR in my alpha branch and @JohnKitching is running a derivative of that, he never complained. @JohnKitching you are running with AppleWatch, right?

JohnKitching:
> @JohnKitching you are running with AppleWatch, right?

I was until very recently. Once I realised I had to remove the watch app and reinstall after a TAI upgrade it's been fine.

PR #675, opened 2025-06-22
Fix missing live activity determination

### LINKED ISSUE #630: Live Activities stopped updating
## Describe the bug
I have a shortcut for restarting LiveActivity in Trio that runs at 6am. At 12:36 I noticed my LiveActivity had disappeared from my LockScreen and DynamicIsland. Opening Trio again immediately restarted LiveActivity. Looking through the logs, it looks like they had stopped updating at 12:16.

## Attach a Log

[log 24.txt](https://github.com/user-attachments/files/20595567/log.24.txt)

## To Reproduce
Look at your phone... occasionally LiveActivity from Trio won't be there when it should be.

## Expected behavior
If you setup Shortcuts to restart LiveActivity in Trio, and you're closed looping successfully, and LiveActivity is enabled in Trio, LiveActivity should always be displayed.

### Smartphone:
* Hardware: iPhone 14 Pro Max
* OS Version: iOS 18.5

### Pump:
* Manufacturer: Medtronic
* Model: 722

### CGM:
* Device: Anubis (Dexcom G6)
* Manager app: official Dexcom G6 app

### Trio Version:
* Version Number: v0.5.0.29
* Repo: mikeplante1/Trio
* Git Reference: [af72be6](https://github.com/MikePlante1/Trio/commit/af72be62c8dd6d7c6ff064e91ed98511524d896c)
* Note: No changes from this branch should be affecting LiveActivity at all.

## Technical Details

Here's the line where it looks like LiveActivity stopped updating:

```
2025-06-04T12:16:06-0400 [Default] LiveActivityManager.swift - pushUpdate(_:) - 372 - DEV: /Users/michaelplante/Documents/GitHub/Trio-mp/Trio/Sources/Services/LiveActivity/LiveActivityManager.swift: Error creating new activity: visibility
```

10nas:
Fixes #630. Explanation of this fix:

> The culprit seems to be that no `determination` is set when `LiveActivityManager` is created when the app starts. Without a determination, [no live activity will be created](https://github.com/nightscout/Trio/blob/3dac85256016a93708ceba175932367c0e0c4109/Trio/Sources/Services/LiveActivity/LiveActivityManager.swift#L478). Trio _tries_ to create the LA while the app is still active, but due to the missing determination, it doesn't have all the information available to be able to actually create a live activity (see previous link). A `determination` only eventually becomes available in `LiveActivityManager` when the first loop happens ([`cobOrIobDidUpdate`](https://github.com/nightscout/Trio/blob/3dac85256016a93708ceba175932367c0e0c4109/Trio/Sources/Services/LiveActivity/LiveActivityManager.swift#L160)). If the app is already in the background, it's too late to create a live activity.
>
> Fix for that is to load an initial value for determination to make sure that a LA can be created when the app is still in active.

The fix is fully contained in the first commit. The second commit just does various cleanups and is strictly seen not necessary to fix #630. But I figured I do these cleanups while I was going through the Live Activity code.

### DIFF
--- a/Trio/Sources/Services/LiveActivity/LiveActivityManager.swift
+++ b/Trio/Sources/Services/LiveActivity/LiveActivityManager.swift
@@ -89,6 +89,7 @@ final class LiveActivityManager: Injectable, ObservableObject, SettingsObserver 
         registerHandler()
         monitorForLiveActivityAuthorizationChanges()
         setupGlucoseArray()
+        setupDetermination()
         broadcaster.register(SettingsObserver.self, observer: self)
     }
 
@@ -257,6 +258,12 @@ final class LiveActivityManager: Injectable, ObservableObject, SettingsObserver 
             } catch {
                 debug(.default, "\(DebuggingIdentifiers.failed) failed to fetch glucose with error: \(error)")
             }
+        }
+    }
+
+    private func setupDetermination() {
+        Task { @MainActor in
+            self.determination = try await fetchAndMapDetermination()
         }
     }
 
--- a/Trio/Sources/Services/LiveActivity/LiveActivityManager.swift
+++ b/Trio/Sources/Services/LiveActivity/LiveActivityManager.swift
@@ -286,9 +286,7 @@ final class LiveActivityManager: Injectable, ObservableObject, SettingsObserver 
     /// Otherwise, it ends the current live activity.
     @MainActor private func forceActivityUpdate() {
         if settings.useLiveActivity {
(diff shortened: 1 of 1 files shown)


marv-out (approved):

dnzxy (approved):
Fixes the intermittent "I cannot start my LA" issue, not the intermittent crashes (which are out of scope here).

Looking good, merging.

dnzxy:
Does not fully address #630, so I will leave that issue open for now @10nas .

PR #676, opened 2025-06-22
Shorten remote command notes for NS view

aug0211:
- Removed "Remote meal command" text and replaced with shortened version
- Less cluttered Nightscout screen
- Selected 📡📲 (satellite to phone) for a very clear, succinct visual explanation

### DIFF
--- a/Trio/Sources/Services/RemoteControl/TrioRemoteControl+Meal.swift
+++ b/Trio/Sources/Services/RemoteControl/TrioRemoteControl+Meal.swift
@@ -77,7 +77,7 @@ extension TrioRemoteControl {
             carbs: carbsDecimal ?? 0,
             fat: fatDecimal,
             protein: proteinDecimal,
-            note: "Remote meal command",
+            note: "📡📲",
             enteredBy: CarbsEntry.local,
             isFPU: false,
             fpuID: fatDecimal ?? 0 > 0 || proteinDecimal ?? 0 > 0 ? UUID().uuidString : nil


aug0211:
Oh, and tested in vivo and validated in the NS view as well as on the looping device.

Looping device:
Screenshot 2025-06-22 at 11 54 54 AM

Nightscout view:
IMG_8846

MikePlante1 (approved):
LGTM.

I think 📡 would have been sufficient, but I can see why some would think 📡📲 to be a better option, so I'll approve it as is.

t1dude (changes requested):
As stated in my comment, I think this should be handled by a setting or a build customization script.

MikePlante1:
To add my 2¢, I can get behind the "through a setting" approach.

I think adding a textbox in ⚙️ > `Features` > `Remote Control` would be good. Maybe default it to "via Remote", but then anyone can just change it to whatever they want, including an emoji

dnzxy:
> As a community, we should really consider more iterative development and merge changes in even if they are small step change improvements

You received 2 kinds of feedback. 
1. Giving context, agreeing with the issue you raised, but not removing a label that was intentionally put in place. 
2. A change request to facilitate this idea / proposal through means of a setting or customization script. 

There is a difference between posting a PR and getting feedback on it, and not going ahead and merging something. No one here said it won’t be merged — it just won’t immediately be merged the way you proposed it. 

You got 2 suggestions here how this can be improved upon, which as you are asking, is iterative. Your PR is the idea and you provided code with, you got feedback, let’s build on feedback. 

> and then build upon them further (eg, if we want to add toggles, allow users to customize these notes, etc).

It’s an improvement from your point of view and the single person asking for it on FB, it’s a regression for how it was in the app. 
FWIW: the label should definitely be shortened, which was my initial suggestion here, as well. So agreeing with you ☺️

I’d ask folks to not immediately close PRs once the feedback is not "LGTM, let’s merge". That‘s also part of collaborative and iterative development.

PR #678, opened 2025-06-22
Don't reset CAGE for non-fixed prime events

MikePlante1:
This changes the event trigger for resetting CAGE from `.prime` to `.replaceComponent(componentType: X)` where `X` can be `.infusionSet` (used by MinimedKit) or `.pump` (used by OmniBLE and OmniKit).

Marked as Draft, because DanaKit does not yet implemented `.replaceComponent(componentType: .infusionSet)` events

### DIFF
--- a/Trio/Sources/APS/Storage/PumpHistoryStorage.swift
+++ b/Trio/Sources/APS/Storage/PumpHistoryStorage.swift
@@ -176,7 +176,8 @@ final class BasePumpHistoryStorage: PumpHistoryStorage, Injectable {
                     newPumpEvent.isUploadedToHealth = false
                     newPumpEvent.isUploadedToTidepool = false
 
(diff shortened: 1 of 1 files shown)


dnzxy:
I know this is a draft PR, but why would we mutate pump manager (PM) emitted events, when storing them in our database? That's mutating information that should not be mutated, is it not? 

By adding this change, we're essentially ignoring true `.prime` events, and are only really storing a prime Trio-sided, when the component `.infusionSet` or `.pump` (depending on underlying PM) gets replaces, thus omitting true primes from our storage.

Is this *really* what you are after here? If so, I strongly suggest that we look for a better alternative. The PM is our single source of truth, and imho we should not mutate (any) information – or here: event (type) – it emits.

PR #680, opened 2025-06-22
Shorten Watch command notes for NS view

aug0211:
- Removed "Via Watch" text and replaced with shortened version
- Less cluttered Nightscout screen
- Selected ⌚️ (watch emoji) for a very clear, succinct visual explanation


aug0211:
Bummer to see the above. I'll close this PR because I don't see anyone realistically running forward with an approach that pleases the more modern crowd who understand emojis and the more conventional group who likes text strings.

PR #686, opened 2025-06-24
Hotfix/add identifiers: Bring in updates to restore Add Identifiers to dev branch

marionbarker:
See PR #685


Sjoerd-Bo3 (commented):
Into DEV it should not have this app version right?

Config.xcconfig
Sjoerd-Bo3, line 1: This is what I mean.

dnzxy: image

MikePlante1 (approved):

PR #688, opened 2025-06-26
Fix Disappearing Live Activity, Cleanup

### LINKED ISSUE #630: Live Activities stopped updating

10nas:
Fixes #630 

What I believe happened there and I was able to repro: Trio is sometimes terminated in the background (or simply crashes for unknown reasons not directly related to Live Activity). The app eventually starts again while in the background (I assume due to a bluetooth event from the CGM?). At that point the live activity would disappear.

Live activities are independent of and persist beyond the app lifecycle. When the app is terminated in the background and restarted the same live activity still exists. However, the current implementation simply ends that live activity and tries to start a new one. If this happens while the app is not even active, creating the new live activity fails because a live activity can only be created while the app is active. The effect is what happened in #630: The live activity disappears.

This PR ensures that an existing live activity is reused, instead of ending any existing live activities and starting a new one upon app launch. I've confirmed that this works. It resolves #630.

This PR also removes a bunch of unnecessary and confusing code paths for updating a live activity. It's now a simple pipeline: Data that is needed to update live activities is in `LiveActivityData`. Whenever `LiveActivityData` changes (via `ObservableObject`), the live activity is updated. All the rest of `LiveActivityManager` needs to do is apply changes to `LiveActivityData` when necessary. A bug that is fixed as part of this is that `widgetItems` was previously uninitialized.

### DIFF
--- a/Trio/Sources/Services/LiveActivity/LiveActivityAttributes+Helper.swift
+++ b/Trio/Sources/Services/LiveActivity/LiveActivityAttributes+Helper.swift
@@ -55,7 +55,7 @@ extension LiveActivityAttributes.ContentState {
         return formatter.string(from: deltaAsDecimal as NSNumber) ?? "--"
     }
 
-    init?(
+    init(
         new bg: GlucoseData,
         prev _: GlucoseData?,
         units: GlucoseUnits,
--- a/Trio/Sources/Services/LiveActivity/LiveActivityManager.swift
+++ b/Trio/Sources/Services/LiveActivity/LiveActivityManager.swift
@@ -239,7 +239,7 @@ final class LiveActivityManager: Injectable, ObservableObject, SettingsObserver 
         }
 
         if let currentActivity {
-            if currentActivity.needsRecreation(), UIApplication.shared.applicationState == .active {
+            if currentActivity.needsRecreation() && UIApplication.shared.applicationState == .active {
                 debug(.default, "[LiveActivityManager] Ending current activity for recreation: \(currentActivity.activity.id)")
                 await endActivity()
                 // After endActivity(), currentActivity is guaranteed to be nil
@@ -294,7 +294,7 @@ final class LiveActivityManager: Injectable, ObservableObject, SettingsObserver 
                     staleDate: Date.now.addingTimeInterval(5 * 60)
                 )
                 await activity.update(updateContent)
-                debug(.default, "[LiveActivityManager] Updated new activity with actual data")
+                debug(.default, "[LiveActivityManager] Set initial content for new activity: \(activity.id)")
             } catch {
                 debug(
                     .default,
--- a/Trio/Sources/Services/LiveActivity/LiveActivityManager.swift
+++ b/Trio/Sources/Services/LiveActivity/LiveActivityManager.swift
@@ -7,7 +7,6 @@ import UIKit
 
 @available(iOS 16.2, *) private struct ActiveActivity {
     let activity: Activity<LiveActivityAttributes>
-    let startDate: Date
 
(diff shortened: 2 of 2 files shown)


MikePlante1:
Built this in-vivo ~90 mins ago and my LA keeps not appearing. Restarted my phone just now and LA appeared once again. I'll keep an eye on it.

I didn't look through my log at all yet, but I'll include it here just in case: 
[log 26.txt](https://github.com/user-attachments/files/20931763/log.26.txt)

marv-out:
I think the properties in the `LiveActivityData`class should be updated on the main thread since this is of type ObservableObject and the properties are published. As you can see here they are currently updated on the background thread which the fetch is happening on: 
 
grafik
grafik

10nas:
> I think the properties in the `LiveActivityData`class should be updated on the main thread since this is of type ObservableObject and the properties are published. As you can see here they are currently updated on a background thread. I think this is what @dnzxy meant.
> 
> grafik

This is fine. An `ObservableObject` can in principle be updated from any thread. What matters is that UI actions resulting from `ObservableObject` observations are executed on the main thread. And this is done by dispatching to the main thread [here](https://github.com/nightscout/Trio/blob/8790033ca335eef14a5c78631a1e69d80fa7970e/Trio/Sources/Services/LiveActivity/LiveActivityManager.swift#L94) in the only subscriber to `data` (the `ObservableObject`). This is also neat: instead of having to make sure that every call site that makes any changes to `data` is on the main thread, you don't have to care. Instead you dispatch to the main thread in the sink. This ends up guaranteeing that the UI changes are done on the main thread.

marv-out:
> This is fine. An `ObservableObject` can in principle be updated from any thread. What matters is that UI actions resulting from `ObservableObject` observations are executed on the main thread.

OK. I think this is good to go then. I also didn’t notice a warning in the terminal saying that we were updating on the wrong thread, as in previous cases. And I can confirm there are no more sudden LA crashes.

marv-out (approved):

g5WS:
I think there may be a bug here. #693  

(Sorry!)

Issue #627, opened 2025-06-04
Pod deactivation during meal bolus caused negative insulin entry in the app


dankvistedal:
## Describe the bug
Yesterday afternoon 16:37 CET I was running low on insulin in the pump, had like 1-2u left. I entered a meal and input a bigger bolus than what I had insulin for, 12u. 
Trio started delivering insulin, and had given me someting like 5-6u before the pod went empty and started beeping. So far all is as expected.
I entered the treatments to enter external insulin as I was prepared to give 6 units with a pen. I thought that the recommendation looked weird, so I checked the home screen and noticed that i had a few units negative IOB. Then looking at the history screen I could only see a negative insulin entry of -3,15.
In order to have more correct data I removed the negative entry from the history and entered 12u external insulin while I administered 6u with a pen.

I made a log after deactivating my pod. 
Later in the evening I paired a new pod (had a few manual insulin injections in between, all entered in trio) and made a new log.

## Attach a Log

[log after pairing new pod.txt](https://github.com/user-attachments/files/20586858/log.after.pairing.new.pod.txt)
[log after pod deactivation.txt](https://github.com/user-attachments/files/20586859/log.after.pod.deactivation.txt)

## To Reproduce
*Steps to reproduce the behavior:*
1. Give bolus bigger than remaining insulin in pod.
2. Deactivate the pod when screaming.
3. Check history screen

## Expected behavior
I have given similar insulin doses before and Trio has always been able to tell me how much of the requested dose was given.

## Screenshots

![Image](https://github.com/user-attachments/assets/d499a520-0ce9-412c-9331-ed408a05abe4)
![Image](https://github.com/user-attachments/assets/59215691-2f0b-4128-9396-a7ff2b34237c)
![Image](https://github.com/user-attachments/assets/2b11469d-def7-4ee6-953b-d642cd2c5767)

### Smartphone:
* Hardware: Iphone 14
* OS Version: iOS 17.5.1

### Pump:
* Manufacturer: Insulet
* Model: Omnipod Dash

### CGM:
* Device: Libre2 plus
* Manager app: xDrip4iOS

### Trio Version:
* Version Number: v0.5.0.3 (14)
* Repo: nightscout/trio
* Git Reference: 0309f6a

## Technical Details
*If applicable, provide any technical details that might help in diagnosing the problem. This could include logs, error messages, or relevant configuration details.*

## Additional context
*Add any other context about the problem here.*

itsmojo:
[...]

PR #690, opened 2025-06-30
update OmniXXX: handle interrupted bolus during pod fault correctly

marionbarker:
## Purpose:

Fix the case where a pod fault is treated as an unacknowledged command during an interrupted bolus and the undelivered insulin is subtracted twice from the original bolus amount.

This was reported for both Trio and Loop (the two apps use a common set of OmniBLE and OmniKit modules):
* [Trio in Issue 627](https://github.com/nightscout/Trio/issues/627)
* [Loop in Issue 2232](https://github.com/LoopKit/Loop/issues/2322)

## Method

The OmniXXX repositories were updated and the code modification tested as reported in these comments:
* https://github.com/LoopKit/OmniBLE/pull/144#pullrequestreview-2969436633
* https://github.com/LoopKit/OmniBLE/pull/144#issuecomment-3017275154

See the two PR for OmniXXX:
* https://github.com/LoopKit/OmniBLE/pull/144
* https://github.com/LoopKit/OmniKit/pull/52


Sjoerd-Bo3 (approved):
LGTM

Issue #624, opened 2025-06-03
Onboarding: Uploading to Nightscout is not enabled when users establish connection


dnzxy:
## Describe the bug

When users enter a Nightscout URL and API secret **during Onboarding**, *uploading* to Nightscout is **not** enabled by default. This leads to confusion, as users typically expect that entering credentials will allow data to be written to Nightscout automatically.

While this is currently _by design_ to keep uploading optional, the result is that treatment data and setting changes made during onboarding are **not** backed up unless the user explicitly enables upload _after completion_ of Onboarding. This behavior can result in loss of data if the app is deleted or the phone is reset right after / during Onboarding, which is not intuitive for users expecting Nightscout to serve as a persistent sync/backup.

## Attach a Log

N/A (user report)

## To Reproduce

1. Launch Trio and begin onboarding
2. Enter Nightscout URL and API secret
3. Complete onboarding and make setting adjustments
4. Delete the app or reset the device
5. Reinstall and reconnect Nightscout — previous changes are not preserved

## Expected behavior

If a user enters a valid Nightscout URL and API secret during onboarding, **upload to Nightscout should be enabled by default**, ensuring data is written and recoverable. Users should still have the option to disable uploading afterward if desired.

## Screenshots

N/A

### Smartphone:

* Hardware: Any iPhone
* OS Version: Any supported iOS version

### Pump:

* N/A

### CGM:

* N/A

### Trio Version:

* Version Number: 0.5.0+
* Repo: `nightscout/trio`

## Technical Details

* Current behavior: Upload is not auto-enabled after Nightscout entry
* Fetching is also disabled unless explicitly enabled
* Settings imported from Nightscout are not backed up to it unless upload is on
* No upload is triggered by default even after adding credentials

## Additional context

This issue particularly affects users (e.g., parents) who assume Nightscout serves as a full sync backend. They're surprised when no data appears after onboarding or when data is lost after app reinstall. Defaulting upload to *on* upon Nightscout connection would reduce support confusion and improve user trust.

**Proposed change**:
Enable Nightscout upload by default if a URL and API secret are entered. Still allow user override.

MikePlante1:
This should have auto-closed when #735 was merged

PR #696, opened 2025-07-04
Enable upload by default in settings and Nightscout config

MikePlante1:
To address Issue #624

### DIFF
--- a/Trio/Resources/json/defaults/freeaps/freeaps_settings.json
+++ b/Trio/Resources/json/defaults/freeaps/freeaps_settings.json
@@ -1,7 +1,7 @@
 {
   "units" : "mg/dL",
   "closedLoop" : false,
-  "isUploadEnabled" : false,
+  "isUploadEnabled" : true,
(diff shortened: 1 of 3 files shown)


kingst (approved):
Tested on the simulator and the code LGTM. I'm not 100% sure about the settings flow but this looks right.

dnzxy (changes requested):
This PR is not doing what it‘s supposed to be introducing. The setting is not supposed to default to true (so enabled), but when a user enters their NS credentials, and has valid credentials, it needs to enabled as a default "follow-up" action.

This becomes especially important during Onboarding, so that if users enter settings when completing the wizard, we can then upload their newly entered settings to NS upon completion. 

This means:
- add logic to make this behavior possible in NightscoutSettingsStateModel and OnboardingStateModel+Nightscout
- Do not default to true generally
- Ensure the NS upload for the profile gets triggered upon completing Onboarding (in the finalize methods)

PR #701, opened 2025-07-05
feat: Add settings export functionality

### LINKED ISSUE #596: Provide way to see all settings in a shareable format
*Provide a clear and concise description of the problem. Explain how this issue affects your experience with the Trio app and any specific scenarios where it occurs.*

As a caregiver I frequently need to make adjustments to settings for my son, but want to review ALL settings at the same time to verify that nothing else may need tweaking as well. Eg maybe the basal and ISF are ok but I truly need to tweak SMB basal minutes or SMB ratio etc. Since all settings are in different screens it takes forever to clock and scroll through them all. I need a way to see everything in read-only format in 1 place, ideally in a format that my teenage son can export from his phone. Then I can tell him exactly what to change from/to what value. This would also help to get assistance with settings - no need to take many screenshots. 

*Detail the desired change or feature you'd like to see implemented in the Trio app. Be specific about how this solution would improve your experience and address the problem described above.*
Click a button "settings log" (or similar naming) that exports a text file, eg TrioSettings_currentdatetime.log that the user can then either open and read, or save and share with someone. 

*List and describe any alternative solutions or features you've considered that could also address the problem. Explain why you believe the proposed solution is the best option.*
The proposed solution doesn't require taking any screenshots and can be easily shared. 

## Additional context
*Include any other context, screenshots, or relevant information that might help in understanding the issue or the proposed solution. If applicable, describe any previous discussions or decisions that relate to this feature request.*

## Technical Details
*If applicable, provide any technical details or considerations that might impact the implementation of this feature. This could include dependencies, potential risks, or required changes to existing functionalities.*

## User Impact
*(Optional) Describe the impact of this issue on your use of the Trio app. Include any specific examples or data that demonstrate how widespread or severe the problem is.*

Sjoerd-Bo3:
## Summary
- Implement complete CSV export for all Trio settings and presets
- Separate preset categories for granular selection control
- Add loading spinner with proper timing during export operations

### Export Categories
Users can selectively export any combination of:
- **Export Info**: Date, app version, build information
- **Devices**: CGM and pump configuration  
- **Therapy**: Basal profiles, ISF, carb ratios, targets
- **Algorithm**: SMB, autosens, dynamic settings
- **Features**: UI preferences, meal settings
- **Notifications**: Alert and notification settings
- **Services**: Nightscout, Apple Health integration
- **Temp Target Presets**: Exercise, eating soon, and custom temp targets
- **Override Presets**: Sensitivity adjustments and insulin factor overrides  
- **Meal Presets**: Saved meal configurations with carbs, fat, and protein

#### CSV Format
- Clear section headers with organized data hierarchy
- Individual preset types grouped in separate categories
- Blank lines between presets for visual clarity
- Proper escaping for special characters
- Unit labels included for all measurements

## Screenshots

| Dark Mode | Light Mode |
|--------|--------|
| image | image | 

Resolves #596


mountrcg:
"Choose which categories to export to a CSV file for backup or sharing with your healthcare provider." text doesn't match the functionality of csv/json export. May be just skip the "to a CSV file".

mountrcg:
I think maxBolus and maxBasal from pumpSettings are missing? They are not in preferences - also DIA is not - but in

```
                // Get pump settings from storage
                if let pumpSettings = storage.retrieve(OpenAPS.Settings.settings, as: PumpSettings.self) {
                    addSetting(
                        category: devicesCategory,
                        subcategory: additionalsSubcategory,
                        name: String(localized: "Duration of Insulin Action (DIA)"),
                        value: String(describing: pumpSettings.insulinActionCurve),
                        unit: String(localized: "hours")
                    )
                    addSetting(
                        category: devicesCategory,
                        subcategory: ...,
                        name: String(localized: "Max Bolus"),
                        value: String(describing: pumpSettings.maxBolus),
                        unit: "U"
                    )
                    addSetting(
                        category: devicesCategory,
                        subcategory: ...,
                        name: String(localized: "Max Basal"),
                        value: String(describing: pumpSettings.maxBasal),
                        unit: "U/hr"
                    )
                }
 ```

**Comment by Sjoerd-Bo3** [comment]
Still a thing on my mind:

Can we make this smarter so adding the settings to the export integrates in the settings themselves? That way it is less likely that changing, removing or adding a setting all together doesn't align with the export feature anymore. @mountrcg You have a view on this?.

I like that way for maintainability reasons.

**Comment by mountrcg** [comment]
re-tested, works now with TT and the remaining pumpSettings. Really useful feature not just for the odd migration but also to just a quick, one glance overview of all settings. Of course I did find a mistake in my settings.

**Comment by dnzxy** [comment]
I‘d very much suggest to break this into a multi step wizard instead of a mega scroll view for more ease of use.

**Comment by dnzxy** [comment]
The placement in Services is also rather misplaced, as it’s not a third party service, but an in-app export. 

This belongs in the root.

**Comment by dnzxy** [comment]
I’d also suggest to get HCP feedback for this feature. I‘m not sure if the current structure of the exported document is useful to someone reviewing settings. They will have insight on that.

**Comment by Sjoerd-Bo3** [comment]
> The placement in Services is also rather misplaced, as it’s not a third party service, but an in-app export.
> 
> This belongs in the root.

Yeah was struggling with that aswell, was in the rootview at first, will do that again!

**Comment by dnzxy** [comment]
> > Still a thing on my mind:
> > Can we make this smarter so adding the settings to the export integrates in the settings themselves? That way it is less likely that changing, removing or adding a setting all together doesn't align with the export feature anymore. @mountrcg You have a view on this?.
> > I like that way for maintainability reasons.
> 
> @dnzxy any good ideas / suggestions on this maybe?

For a very very large part and for good reason we moved away from a single big class or collection of settings that get dynamically rendered. iAPS and 0.2.x main has that.

There are DecimalPickerSettings, Preferences, TrioSettings and PumpSettings structs that contain all settings - so for as long as you build it like that, it probably is about as much as you can do. 

In Onboarding I went with some custom
structs that build on existing structs but after all it did / does lead to somewhat duplicated code. 

Settings don’t change often and as much - so while I generally agree it would be nicer for maintainability, we just don’t really often have to maintain these.

**Comment by Sjoerd-Bo3** [comment]
> > > Still a thing on my mind:
> > > Can we make this smarter so adding the settings to the export integrates in the settings themselves? That way it is less likely that changing, removing or adding a setting all together doesn't align with the export feature anymore. @mountrcg You have a view on this?.
> > > I like that way for maintainability reasons.
> > 
> > 
> > @dnzxy any good ideas / suggestions on this maybe?
> 
> For a very very large part and for good reason we moved away from a single big class or collection of settings that get dynamically rendered. iAPS and 0.2.x main has that.
> 
> There are DecimalPickerSettings, Preferences, TrioSettings and PumpSettings structs that contain all settings - so for as long as you build it like that, it probably is about as much as you can do.
> 
> In Onboarding I went with some custom structs that build on existing structs but after all it did / does lead to somewhat duplicated code.
> 
> Settings don’t change often and as much - so while I generally agree it would be nicer for maintainability, we just don’t really often have to maintain these.

Thnx! Thats the kind of elaborated response I hoped to get from you😬 Thanks for that💙

**Comment by dnzxy** [comment]
How's the status on this one?

**Comment by dnzxy** [comment]
3 months later, what's the status here?

**Comment by oddst** [comment]
Firstly I would like to praise this effort, as this is going to be a huge improvement for many newer Trio users that typically are less technically experienced and their support crews...! For people like me that currently supports multiple Trio Dev Open Beta users remotely, this will be very helpful. 

I would also claim that settings may be something that do not change very often for the very experienced Trio developers, but for new Trio users, young users with lot's of hormonal changes,  users migrating from Loop, users coming from MDI and generally just less experienced users - this is going to be a life saver...!

Helping people remotely migrating from Loop -> Trio, is no small task - especially when we are talking about young people...! Especially young people between 15-25 yo that would like to push Trio as far as they can with regard to not entering carbs or avoiding meal boluses - demands a fair amount of settings changes over at least 3 months time per user. 

So I am eternally greatful when this effort now will find it's way to Trio Dev Open Beta...! Thank you, thank you, thank you...!!! From me, and but also from the young people that won't have to send me around 30 screen shots for every round of changes...!

Since the examplification data (the CSV/JSON export files) in this PR might not be completely up-to-date, I will ask some  questions - and my apologies if things have already been handled/fixed...!

- Export formats:
   - JSON is fantastic for developers and technical people
   
   - CSV will be much more helpful for less technical people, but depends on the viewing device's functionality with regard to displaying CSV files. On my Mac this works very user friendly, on the iPhone a little bit less user friendly and on a PC more challenging. Override definitions will be just too difficult for ordinary users to view in the JSON exports.
   
   - PDF, a PDF export alternative might improve the user friendly access to these exported data (especially for users on mobile devices or PCs). 
   
   - SEND, a possibility to send the exported report (with the chosen format) via iMessage, e-mail or messenger type services would also be helpful.  

I would like to question whether the following information is now a part of the export or not (if not, including it would be helpful):

- Devices > Dexcom G6/G7 (the Upload Readings settings for both G6 and G7 would be helpful to have included)

- Devices > Omnipod Dash (i would like to be able to see the unique settings for Omnipods like:
   - Expiration Reminder
   - Low Reservoir
   - Confidence Reminder
   - Silence Pod
   - Insulin Type (is listed under Therapy settings which is not similar to Trio)

- Devices > Smart Watch (the underlying information here, would be useful)

- Algorithm > Additionals (the actual Custom Type number of minutes (PT) should be included, when this override is enabled)

- Features > Bolus Calculator (here are a number of settings that will be changed a lot the first 3-6 months at least)

- Notifications > Live Activity settings (would be helpful)

- Notifications > Calendar Events (helpful with setup of watches)

- Services > Nightscout (should Nightscout URL be included, or is that deemed too sensitive info?)

Thank you...!

**Comment by dnzxy** [comment]
> I would also claim that settings may be something that do not change very often for the very experienced Trio developers, but for new Trio users, young users with lot's of hormonal changes, users migrating from Loop, users coming from MDI and generally just less experienced users - this is going to be a life saver...!

Just a side note: The context above was settings do not change much in code, not the user interface end of a setting. @Sjoerd-Bo3 and I were discussing code maintainability. No one implied that there isn’t a user group with frequent setting changes :)

**Comment by Sjoerd-Bo3** [comment]
> 3 months later, what's the status here?

Friday is my pre-hackathon day (took the day off to get in the groove and make features) and will finish this and another feature I have been working on!

**Comment by dnzxy** [comment]
So, how's it looking now?

**Comment by dnzxy** [comment]
@Sjoerd-Bo3 are you going to finish this in the next 4 weeks? If not, I'll convert this to a draft and probably finish this myself.

**Comment by oddst** [comment]
> Firstly I would like to praise this effort, as this is going to be a huge improvement for many newer Trio users that typically are less technically experienced and their support crews...! For people like me that currently supports multiple Trio Dev Open Beta users remotely, this will be very helpful.
> 
> I would also claim that settings may be something that do not change very often for the very experienced Trio developers, but for new Trio users, young users with lot's of hormonal changes, users migrating from Loop, users coming from MDI and generally just less experienced users - this is going to be a life saver...!

@Sjoerd-Bo3, any chance that there will be time to finish this feature...? If possible, it would be very helpful. I do however understand that all DIY development takes place due to the kindness of developers devoting their (free) time to make this happen. Thanks...!

**Comment by dnzxy** [comment]
@oddst if Sjoerd cannot finish it then I'll look into it. 
This feature was discussed with various folks, including several HCPs, at the hackathon in Vancouver.

File: Trio/Sources/Modules/Export/ExportStateModel.swift [hunk 1]
```diff
@@ -0,0 +1,1514 @@
...
+            debug(.default, "🔄 EXPORT: Starting settings export...")
+
+            let categoriesToExport = categories ?? selectedCategories
+            let exportFormat = format ?? selectedFormat
+            debug(
+                .default,
+                "🔄 EXPORT: Exporting categories: \(categoriesToExport.map(\.rawValue).joined(separator: ", ")) in \(exportFormat.rawValue) format"
+            )
+
+            debug(
```
kingst, line 94: CoreData should be ready to go -- it won't load the app unless it is. Have you seen failures in testing with it?

Trio/Sources/Modules/Export/ExportStateModel.swift
```diff
@@ -0,0 +1,1514 @@
...
+                            break // Success, exit retry loop
+                        } catch {
+                            debug(.default, "❌ EXPORT: Attempt \(retryCount + 1) failed: \(error.localizedDescription)")
+                            debug(.default, "❌ EXPORT: Full error: \(error)")
+                            retryCount += 1
+                            if retryCount < 3 {
+                                // Wait progressively longer between retries
+                                debug(.default, "🔄 EXPORT: Waiting \(retryCount * 200)ms before retry...")
+                                do {
+                                    try await Task.sleep(nanoseconds: UInt64(retryCount * 200_000_000)) // 0.2s, 0.4s
```
kingst: The retry / sleep code shouldn't be needed. My concern is that this is masking another issue or there are problems in the way we're initializing CoreData and it isn't actually ready when we let the app launch.

Trio/Sources/Modules/Export/ExportStateModel.swift
```diff
@@ -0,0 +1,1514 @@
...
+            }
+
+            // Convert data to the selected format and write to file
+            do {
+                let content: String
+
+                switch exportFormat {
+                case .csv:
+                    // Helper function to escape CSV values
+                    func csvEscape(_ value: String) -> String {
```
kingst: It might be worth checking if there is a standard CSV encoder, although CSV is really simple this type of logic is quite error prone.

Trio/Sources/Modules/Export/ExportStateModel.swift
```diff
@@ -0,0 +1,1514 @@
...
+                            }
+                        }
+                    }
+
+                    debug(.default, "🔄 EXPORT: Found \(overridePresetIDs.count) override preset IDs")
+                    if !overridePresetIDs.isEmpty {
+                        let overrideSubcategory = String(localized: "Override Presets")
+                        // Convert NSManagedObjectIDs to actual OverrideStored objects and extract their data within the Core Data context
+                        // Add a small delay to ensure Core Data is ready, especially on first app launch
+                        try await Task.sleep(nanoseconds: 100_000_000) // 0.1 second
```
kingst: Same comment, I'm concerned about sleeps masking some other issue that we don't fully understand...

kingst (commented):
Overall the code looks like what I would expect it to look like. I wish there was a way to simplify this, but I couldn't think of one.

My one concern is over the use of `sleep` and comments about CoreData. Did you experience errors with CoreData when testing? Any time I see `sleep` in code it makes me very very nervous unless we have a rock solid understanding of the failure mode and reasoning to show that `sleep` is the right way to handle it. We make sure that CoreData is loaded and ready before launching the app, so this shouldn't be needed, but if it is there is something else going on.

I tested it in the simulator and it appeared to work well (I didn't see any mistakes)

Trio/Sources/Modules/Export/ExportStateModel.swift
dnzxy: Apparently there is no standard CSV decoder. I could find these
- https://github.com/swiftcsv/SwiftCSV
- https://github.com/swiftcsv/SwiftCSV
Both seem rather unmaintained and I am not super comfortable introducing a third party library for this. 
What do you think?

Trio/Sources/Modules/Export/ExportStateModel.swift
dnzxy: Removing this. Should not be needed.

Trio/Sources/Modules/Export/ExportStateModel.swift
dnzxy: Removing this, too.

dnzxy:
Thanks for the review and feedback. I have pushed the following adaptations:

* JSON export implementation
  * Before: JSON was manually assembled using nested `[String: Any]` dictionaries with forced casts.
  * After: Introduced dedicated `Codable` models (`ExportSetting`, `ExportSettingPayload`) and switched to `JSONEncoder` with `.prettyPrinted` and `.sortedKeys` for safer, deterministic output.

* Export model extraction
  * Before: `ExportSetting` was defined inline inside `exportSettings`, tightly coupled to the export logic.
  * After: Add `ExportSetting` + payload into a standalone model file (`Models/ExportSetting.swift`)

* Core Data access
  * Before:Core Data objects were partially mapped into tuples; removed retry and sleep logic.
  * After: All Core Data object resolution is done via `viewContext.perform { existingObject(with:) }`

* Export lifecycle handling
  * Before: `isExporting` state handling was implicit and not guaranteed to reset.
  * After: Explicitly set on start and reliably reset via `defer` + `MainActor`.

* CSV export robustness
  * Before: CSV escaping did not handle carriage returns; no BOM.
  * After: Escaping includes `\r`, and a UTF-8 BOM is added for Excel compatibility.

* Removed debug-only logging, retry loops, preset name heuristics, and redundant file-existence checks.
* Reduced overall code size and branching while keeping exported data unchanged.

dnzxy:
Final addition: 
I have realized that prior to `04fd879` we were exporting all keys as localized keys, even for JSON, and that is probably rather useless, as it would contain whitespaces &/or umlaut (for certain languages).
Since I could not think of any way a JSON export working like that would be useful for, and after some private discussions with @kingst, I have removed JSON from the UI (the format selection) as well as from the state model's processing logic and cleaned up accordingly. View a JSON as an everyday-Joe or an endocrinologist/HCP who wants an overview of settings is not really helpful. 

For now, PR should be good to go from my end. 

I'll record an additional issue (to #596) to put down some more work to further clean up the state model and potentially enhance this feature after getting more user feedback…

Trio/Sources/Modules/Export/ExportStateModel.swift
kingst: I'm surprised that Apple doesn't support it, yeah no need for a third party for this.

kingst (approved):
The changes all LGTM

MikePlante1:
Played around with this PR a bit in an iPhone Air simulator running iOS 26.1.

* Should `Export Information`/`File Name` be removed from the UI? It doesn't seem to serve any useful purpose as you can't click it to edit the file name, and the actual name appears as soon as you click `Export Settings` 

* Should `Select All` change to `Deselect All` when everything is selected? It also looks like `Select All` kind of flickers when it's pressed even though none of the rest of the text on the screen does. I base this question off of the History Treatment Filter menu.
  * Sidenote... I think the History Filter menu should get updated to to use boxes instead of circles... in a different PR, of course

* How do settings get organized? In `Algorithm > Additionals` it's strange to me that `Use Custom Peak Time` is separated from `Insulin Peak Time` by `Duration of Insulin Action`.

* I find it strange to have `Insulin Type` listed under `Therapy > Units and Limits`. Maybe it could be moved to `Devices > Insulin Type` right under `Devices > Pump Type`?

* It would be nice to include the 4th version number when it's available, so `0.6.0.34` instead of just `0.6.0`

* For the times for Therapy Settings, we could drop the seconds which will always be `:00` — so just `Basal Rate (18:30)` instead of `Basal Rate (18:30:00)`

dnzxy:
Thanks for looking into it and testing @MikePlante1 

> * Should `Export Information`/`File Name` be removed from the UI? It doesn't seem to serve any useful purpose as you can't click it to edit the file name, and the actual name appears as soon as you click `Export Settings`

Agreed, removed.

> * Should `Select All` change to `Deselect All` when everything is selected? It also looks like `Select All` kind of flickers when it's pressed even though none of the rest of the text on the screen does. I base this question off of the History Treatment Filter menu.

Agreed, adjusted this.

>   * Sidenote... I think the History Filter menu should get updated to to use boxes instead of circles... in a different PR, of course

Follow-up PR coming up.
**Edit:** https://github.com/nightscout/Trio/pull/913

> * How do settings get organized? In `Algorithm > Additionals` it's strange to me that `Use Custom Peak Time` is separated from `Insulin Peak Time` by `Duration of Insulin Action`.

The way @Sjoerd-Bo3 implemented this (and I'm not eager to changing this) is that every setting gets its own row, and technically there are two settings here: 1. use custom peak time, 2. the actual custom peak time.

> * I find it strange to have `Insulin Type` listed under `Therapy > Units and Limits`. Maybe it could be moved to `Devices > Insulin Type` right under `Devices > Pump Type`?

Agreed. Moved to Device category.

> * It would be nice to include the 4th version number when it's available, so `0.6.0.34` instead of just `0.6.0`

Agreed. Changed it from `Bundle.main.infoDictionary?["CFBundleShortVersionString"]` to our Trio specific `Bundle.main.appDevVersion`

> * For the times for Therapy Settings, we could drop the seconds which will always be `:00` — so just `Basal Rate (18:30)` instead of `Basal Rate (18:30:00)`

Agreed. Added a new helper function as extension to `String` (`formattedHourMinuteFromTimeString()`) that'll do that for us. It now displays rates in the label as `HH:mm` instead of `HH:mm:ss` (i.e. how it's stored).

Trio/Sources/Modules/Export/View/ExportRootView.swift
```diff
@@ -27,9 +27,13 @@ extension Export {
                                 HStack {
                                     Image(systemName: state.allCategoriesSelected ? "checkmark.square.fill" : "square")
                                         .foregroundColor(state.allCategoriesSelected ? .blue : .secondary)
-                                    Text("Select All")
-                                        .fontWeight(.bold)
-                                        .foregroundColor(.primary)
+                                    Text(
+                                        state
+                                            .allCategoriesSelected ? String(localized: "Deselct All") :
```
MikePlante1: typo: `Deselect All`

Trio/Sources/Modules/SettingsExport/SettingsExportStateModel.swift
```diff
@@ -24,7 +24,7 @@ extension Export {
         let viewContext = CoreDataStack.shared.persistentContainer.viewContext
 
         override func subscribe() {
-            versionNumber = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String ?? "Unknown"
+            versionNumber = Bundle.main.appDevVersion ?? "Unknown"
```
MikePlante1, line 27: When, say, main gets updated to 0.7.0, will `appDevVersion` be `0.7.0`?

Trio/Sources/Helpers/String+Extensions.swift
```diff
@@ -8,6 +8,21 @@ extension String {
     mutating func capitalizeFirstLetter() {
         self = capitalizingFirstLetter()
     }
+
+    func formattedHourMinuteFromTimeString() -> String {
```
MikePlante1, line 12: Looks like this could be reformatted to the following, but I suppose it doesn't really matter:

```swift
func formattedHourMinuteFromTimeString() -> String {
    let input = DateFormatter()
    input.dateFormat = "HH:mm:ss"
    
    guard let date = input.date(from: self) else {
        return self
    }
    
    return date.formatted(date: .omitted, time: .shortened)
}

```

Trio/Sources/Modules/SettingsExport/SettingsExportStateModel.swift
```diff
@@ -0,0 +1,1333 @@
...
+                if let pumpManager = provider.deviceManager.pumpManager {
+                    addSetting(category: devicesCategory, name: String(localized: "Pump Type"), value: pumpManager.localizedTitle)
+                } else {
+                    addSetting(
+                        category: devicesCategory,
+                        name: String(localized: "Pump Type"),
+                        value: String(localized: "Not Connected")
+                    )
+
+                    // Get insulin type from pump manager if available, otherwise from preferences
```
MikePlante1, line 154: Should this be moved outside of the `else` statement? As-is, "Insulin Type" would only get exported if pumpManager is not found and Pump Type is set to Not Connected.

MikePlante1 (changes requested):
Added some in-line comments/questions.

Trio/Sources/Modules/SettingsExport/SettingsExportStateModel.swift
dnzxy: yes

Trio/Sources/Modules/SettingsExport/SettingsExportStateModel.swift
dnzxy: You cannot really set a pump type without a fully onboarded pump manager. I did not regard a simulator pump as "properly" onboarded, so no, it shouldn’t imho.

MikePlante1: Okay, I just built this to my phone to test in-vivo then, and it still doesn't show Insulin Type.

It also just says "plugin" for "CGM" instead of "Dexcom G6".

dnzxy: Yeah I was being an idiot and added in to the else condition rather than to the if condition. Getting your initial comment now, too, lol 😂 

Fixed!

image

Trio/Sources/Modules/Export/View/ExportRootView.swift
dnzxy: Fixed

dnzxy:
Should be good to go now, @MikePlante1  :) 

If you're happy with it, lets go for it. Given the # of (semi useless) commits of this one, I'd suggest to probably squash-merge this one.

MikePlante1:
Hmm, when I run this connected to Xcode, it breaks repeatedly on this line. When I hit the "Continue program execution" play button, it breaks on that line again for the same temp target. When not connected to Xcode, it seems to export fine, though.



---

Also, the TT and OR export lines are a bit confusing to look at:



Maybe the second column could be the preset's name since currently it's just the same as the first column, like this?



Or if that's not possible, at least add a separator like `, ` or ` | ` between the name of the preset and the setting. Also, "Basal Rate Adjustment" and "Target Glucose" don't seem to have names in the list, but maybe that's part of the Xcode issue at the beginning of this comment?

---

Lastly, I seem to have an extra Override Preset that doesn't have a name and isn't displayed when I go to the Adjustments > Overrides view in my Trio app. I doubt it's because of this PR, just that testing this PR made me discover it. 🤔

Trio/Sources/Modules/SettingsExport/SettingsExportStateModel.swift
MikePlante1: Just confirming this now works fine with in-vivo testing with the latest PR update, and now shows the Insulin Type.

It still just shows "plugin" for CGM instead of "Dexcom G6", but I suspect that would be a more difficult fix and is not a blocker for this PR imho

dnzxy:
Definitely did not have those crashes on my end. 
Back to the drawing board, let me fix that 👀😂

MikePlante1 (approved):
Tested in-vivo and everything seems to work great!

dnzxy:
Force push just to correct typo in commit message.

MikePlante1 (approved):

marionbarker:
## Test

I had to download @Sjoerd-Bo3 branch and then merge that branch into dev (because the PR is out of date with base branch).

There were no errors on merge.

This worked with two test phones: SE 2nd gen with iOS 18 and SE 3rd gen with iOS 26.2.

marionbarker (approved):
Approve from test.

marionbarker (approved):
## Status

I did not do a code review but I approve by test.

## Second Test

Now that the PR has no conflicts, I did a second test following `git pull origin pull/701/head`.

I compared the csv values for several different options (selected items and full set) against the setting on the Trio app and they agreed in all cases.

Issue #608, opened 2025-05-29
Trio does not backfill properly from Share with G6, and Share credentials often prevent new readings from G6


MikePlante1:
## Attach a Log

[log dexcom share.txt](https://github.com/user-attachments/files/20510200/log.dexcom.share.txt)

## To Reproduce
*Steps to reproduce the behavior:*
1. Use G6 as your CGM source in Trio with a paired G6 but no Share credentials entered and wait for readings to start coming in. Readings should populate consistently every 5 minutes.
2. Tap the bobble in Trio to bring up the G6 module, enter/add your Share credentials
3. Notice that 5 minute readings are no longer consistent and often get skipped

## Expected behavior
When Share credentials are added, readings should continue to populate fine in Trio when they are populated in the G6 app.

If your G6 transmitter is out of range of your phone for over 5 minutes, G6 will miss the reading, but it will backfill in the G6 app the next time the transmitter successfully connects to the G6 app. When this happens, Trio _should_ backfill the missed reading from Share and then populate the latest reading.

## Screenshots

https://github.com/user-attachments/assets/76e68fb1-6179-45db-9ed5-565bab2b8ebb

https://github.com/user-attachments/assets/86dd07fe-866f-4dc8-9041-4985a6e2131c

<img width=333px src=https://github.com/user-attachments/assets/583795bc-ffe4-4fae-a522-f8951ed29f6d>
<br><br>
<img width=555px src=https://github.com/user-attachments/assets/13638300-5bbd-4e8b-a615-169d2d9676e7>
<br><br>
<img width=555px src=https://github.com/user-attachments/assets/dcfeb2b6-1c15-4a22-b640-6329d6d5867a>

### Smartphone:
* Hardware: iPhone 14 Pro Max
* OS Version: iOS 18.5

### CGM:
* Device: Dexcom G6
* Manager app: official Dexcom G6 app

### Trio Version:
* Version Number: 0.5.0.18

## Additional context

It used to work fine, but stopped functioning properly sometime this year. 

I think this regression was introduced in [PR 386](https://github.com/nightscout/Trio-dev/pull/386) in the private repo, but @dnzxy thinks it was [PR 409](https://github.com/nightscout/Trio-dev/pull/409).

Here's a very old Issue regarding G6 backfill: https://github.com/nightscout/Trio/issues/287 which was fixed in https://github.com/nightscout/Trio/pull/282

Here's another issue that affects G6 backfill: https://github.com/nightscout/Trio/issues/293

[...]

PR #703, opened 2025-07-05
Fixes a race condition between share client and BLE for G6 CGMs

kingst:
This PR fixes a race condition in PluginSource when using a G6 CGM and share client credentials are set. For more details, please see this issue: https://github.com/nightscout/Trio/issues/608

This fix simply ignores share client values to avoid the race condition. I confirmed that none of the other plugin CGMs add CGM readings using the `fetchDataIfNeeded` code path, so this will only impact G6.

This is a simple one-line fix for the problem, but the downside is that we won't get backfilled CGM values from share client any longer. To retain this capability we will need to do a major refactor of this whole bit of functionality, which is beyond the scope of this simple fix.

### DIFF
--- a/Trio/Resources/InfoPlist.xcstrings
+++ b/Trio/Resources/InfoPlist.xcstrings
@@ -457,18 +457,6 @@
         }
       }
     },
-    "NSCalendarsFullAccessUsageDescription" : {
-      "comment" : "Privacy - Calendars Full Access Usage Description",
-      "extractionState" : "extracted_with_value",
-      "localizations" : {
-        "en" : {
-          "stringUnit" : {
-            "state" : "new",
-            "value" : "To create events with BG reading values, so that they can be viewed on Apple Watch and CarPlay"
-          }
-        }
-      }
-    },
     "NSCalendarsUsageDescription" : {
       "comment" : "Privacy - Calendars Usage Description",
       "extractionState" : "extracted_with_value",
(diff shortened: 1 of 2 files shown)


bjornoleh:
I wonder if this will break a use case for testing purposes that has been very useful:

We use a spent G7 or a working G6 transmitter for a heartbeat to allow background operation of the app in testing (sim pump) mode. The BLE device name for the G7 or G6 is entered in the G6 sensor ID field. This lets Trio receive the BLE heartbeat, but there is of course no sensor data to be received. CGM data is instead fetched from Dexcom Share. This allows the app to loop in the background based on real CGM data, and has several use cases for testing.

kingst:
@bjornoleh yes, this PR breaks this testing scenario. As I mentioned in discord, the Combine logic is really subtle and messy (which is why we had this bug in the first place) and needs to be refactored. My suggestion is that we break this into two steps, the first step fixes the bug, which many people face, and then add backfills / support for this testing scenario in a second step.

marv-out:
> My suggestion is that we break this into two steps, the first step fixes the bug, which many people face, and then add backfills / support for this testing scenario in a second step.

I think this sounds reasonable (if it really only affects G6, haven't confirmed that)

marv-out (approved):

dnzxy (approved):
Moving this forward as it fixes an actual issue for many users.

Issue #689, opened 2025-06-26
Decimal Separator hardcoded at top of Treatment view


MikePlante1:
## Describe the bug
The decimal separator at the top of the Treatment view uses a `.` even when the `Recommendation` and `Bolus` use a `,`

## To Reproduce
*Steps to reproduce the behavior:*
1. Set Region to something like Germany
2. Open Treatment view in Trio
3. Add something like `0.3 U` of insulin
4. *See error*

## Expected behavior
I'd expect all decimal separators to be the same

## Screenshots

| . | |
| --- | --- |
| ![Image](https://github.com/user-attachments/assets/9073459a-a48d-4f42-b394-551a79d1088b) | ![Image](https://github.com/user-attachments/assets/5b4c2908-ef19-46da-8cdc-54f4b74ab838) |

### Smartphone:
* Hardware: iPhone 14 Pro Max
* OS Version: iOS 18.5

### Pump:
* Manufacturer: Medtronic
* Model: 722

### Trio Version:
* Version Number: 0.5.1
* Repo: nightscout/Trio

## Technical Details
*If applicable, provide any technical details that might help in diagnosing the problem. This could include logs, error messages, or relevant configuration details.*

## Additional context
*Add any other context about the problem here.*

bjornoleh:
This will be fixed when https://github.com/nightscout/Trio/pull/707 is merged. That PR is good to go from my perspective.

PR #707, opened 2025-07-07
fixes decimal separator on the top of treatment view according region 

Lamercho:
addresses issue #689 

Simulator Screenshot - iPhone 16 - 2025-07-08 at 10 52 23

### DIFF
--- a/Trio/Sources/Helpers/Formatters.swift
+++ b/Trio/Sources/Helpers/Formatters.swift
@@ -32,6 +32,7 @@ extension Formatter {
     static let decimalFormatterWithTwoFractionDigits: NumberFormatter = {
         let formatter = NumberFormatter()
         formatter.numberStyle = .decimal
+        formatter.locale = .current
         formatter.maximumFractionDigits = 2
         return formatter
     }()
@@ -51,6 +52,7 @@ extension Formatter {
     static let decimalFormatterWithOneFractionDigit: NumberFormatter = {
         let formatter = NumberFormatter()
         formatter.numberStyle = .decimal
+        formatter.locale = .current
         formatter.maximumFractionDigits = 1
         return formatter
     }()
@@ -69,6 +71,7 @@ extension Formatter {
 
         switch units {
         case .mmolL:
+            formatter.locale = .current
             formatter.minimumFractionDigits = 1
             formatter.maximumFractionDigits = 1
         case .mgdL:
--- a/Trio/Sources/Helpers/Formatters.swift
+++ b/Trio/Sources/Helpers/Formatters.swift
@@ -84,9 +84,10 @@ extension Formatter {
     static let bolusFormatter: NumberFormatter = {
         let formatter = NumberFormatter()
         formatter.numberStyle = .decimal
+        formatter.locale = .current
         formatter.minimumIntegerDigits = 0
         formatter.maximumFractionDigits = 2
-        formatter.decimalSeparator = "."
+        // formatter.decimalSeparator = "."
         return formatter
     }()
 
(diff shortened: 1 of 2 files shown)


bjornoleh:
I can confirm that this works. But I wonder if this file for shared formatters could possibly be where this should be addressed? I really don’t know, this is just a hunch. 

https://github.com/nightscout/Trio/blob/dev/Trio/Sources/Helpers/Formatters.swift

Lamercho:
> I can confirm that this works. But I wonder if this file for shared formatters could possibly be where this should be addressed? I really don’t know, this is just a hunch.
> 
> https://github.com/nightscout/Trio/blob/dev/Trio/Sources/Helpers/Formatters.swift
You are absolutely right. Sorry that I missed that. Done with last commit. 
Do I have to touch bolusFormatter also ? There is hardcoded separator there.

bjornoleh:
Where is the bolusFormatter used? Maybe you could just use this one instead of decimalFormatterWithTwoFractionDigits?

I would think that separator is localised already, could you try without changing this?

Lamercho:
> Where is the bolusFormatter used? Maybe you could just use this one instead of decimalFormatterWithTwoFractionDigits?
> 
> I would think that separator is localised already, could you try without changing this?

If I use bolusFormatter without any changes this is the result:

Simulator Screenshot - iPhone 16 - 2025-07-10 at 13 59 08

Lamercho:
> Ok, thanks for testing. Then perhaps use bolusFormatter with the changes for localisation of decimal separators, and I suppose it’s relevant to add that change to all formatters in that file.

Something like this?

bjornoleh:
Yes, this looks good, I tested and it seems to work fine. Using the bolusFormatter makes this a minimal change by applying a formatter that seems to be made for this. 

If you delete the commented out hardcoded separator and perhaps apply the localisation to the other formatters, I suppose this is good to go.

Trio/Sources/Helpers/Formatters.swift
```diff
@@ -81,9 +84,10 @@ extension Formatter {
     static let bolusFormatter: NumberFormatter = {
         let formatter = NumberFormatter()
         formatter.numberStyle = .decimal
+        formatter.locale = .current
         formatter.minimumIntegerDigits = 0
         formatter.maximumFractionDigits = 2
-        formatter.decimalSeparator = "."
+        // formatter.decimalSeparator = "."
```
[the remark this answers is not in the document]
Lamercho: just removed it

bjornoleh (approved):
LGTM!

Trio/Sources/Modules/Treatments/View/ForecastChart.swift
```diff
@@ -73,8 +59,11 @@ struct ForecastChart: View {
 
             HStack {
                 Image(systemName: "syringe.fill")
-                Text("\(state.amount.description) U")
+                Text(
+                    "\(Formatter.bolusFormatter.string(from: state.amount as NSNumber) ?? state.amount.description) U"
```
dnzxy, line 63: the U here actually needs to be a localized string, as for example in German it would be `IE`

I realize this is not your "mistake", but while you are at it, please fix it.

dnzxy (changes requested):
Minor comment. 
Was the `glucoseFormatter` in `Trio/Sources/Modules/Treatments/View/ForecastChart.swift` because it was unused?

Trio/Sources/Modules/Treatments/View/ForecastChart.swift
Lamercho: please check last commit

dnzxy (changes requested):

dnzxy (approved):

PR #709, opened 2025-07-07
fix: Handle nil IDs in CarbsEntry.convertSyncCarb to prevent Tidepool upload crashes

### LINKED ISSUE #656: Crash during background carbohydrate uploads in BaseTidepoolManager.uploadCarbs(_:)
## Describe the bug

When the Trio app (v0.5.0 (22)) processes pending carbohydrate uploads on its background network queue, it crashes instead of completing the sync. The fault occurs in the `BaseTidepoolManager.uploadCarbs(_:)` closure on the `BaseNetworkManager.processQueue` thread, rather than successfully sending the data.

## Attach a Log

I’ve attached the Crashlytics stack trace for the June 5 crash session:
`org.nightscout.trio_issue_28e30203958fd16bda3d7870c0bddeac_crash_session_97189b2a7bbc4873822e27f4ba40c3b2_DNE_0_v2_stacktrace.txt`&#x20;

## To Reproduce

1. Launch the Trio app (version 0.5.0 (22)).
2. Create or have a pending carbohydrate entry in HealthKit (e.g., log a meal in the app).
3. Ensure the device has network connectivity.
4. Wait for the automatic background sync to trigger.
5. Observe the crash on the `BaseNetworkManager.processQueue` thread.

## Expected behavior

Carbohydrate entries should upload to the Tidepool backend without crashing, and the app should continue operating normally after the sync completes.

## Screenshots

*No screenshots available; crash happens during a background network operation.*

### Smartphone:

* Hardware: iPhone 15 Pro
* OS Version: iOS 17.5

### Pump:

* Manufacturer: Insulet
* Model: Omnipod Dash

### CGM:

* Device: Dexcom G7
* Manager app: Dexcom App

### Trio Version:

* Version Number: 0.5.0 (22)
* Repo: nightscout/trio
* Git Reference: *commit hash unknown*

## Technical Details

* **Thread:** `BaseNetworkManager.processQueue`
* **Faulting frame:**

  ```
  0  Trio    0x4449a4 closure #1 in BaseTidepoolManager.uploadCarbs(_:) + 4367796644
  1  Trio    0x3a64d0 <deduplicated_symbol> + 4367148240
  …
  ```
* **Likely cause:** an unexpected nil or race condition inside the carb-upload closure leading to an uncaught exception or memory fault during network‐queue processing&#x20;

## Additional context

This issue appears only when the background sync processes carbohydrate data. Other network operations (e.g., uploading insulin or glucose) seem to complete without error. It may be triggered by a malformed carb entry or a threading issue within `BaseTidepoolManager.uploadCarbs(_:)`.

Sjoerd-Bo3:
## Summary
- Fix crash in Tidepool carb upload when CarbsEntry has nil ID
- Replace force unwrapping with safe UUID generation
- Add debug logging for tracking when fallback UUIDs are used
- Resolves #656

## Problem
The app crashes during background carbohydrate data uploads to Tidepool when `CarbsEntry.convertSyncCarb()` force-unwraps a nil `id` value:
- `CarbsStorage.getCarbsNotYetUploadedToTidepool()` can return entries with `id: nil`
- `convertSyncCarb()` calls `UUID(uuidString: id\!)` which crashes on nil
- Occurs specifically in `BaseTidepoolManager.uploadCarbs()` during background sync

## Solution
- Replace `UUID(uuidString: id\!)` with safe UUID handling
- Generate new UUID when original ID is nil or invalid
- Add debug logging to track fallback UUID generation
- Maintain backward compatibility while preventing crashes

## Test plan
- [x] Build compiles without errors
- [x] No compilation warnings or diagnostics
- [ ] Test with actual carb entries that have nil IDs
- [ ] Verify Tidepool sync continues working after fix
- [ ] Test background sync operations don't crash

### DIFF
--- a/Trio/Sources/Models/CarbsEntry.swift
+++ b/Trio/Sources/Models/CarbsEntry.swift
@@ -42,13 +42,22 @@ extension CarbsEntry {
 
 extension CarbsEntry {
     func convertSyncCarb(operation: LoopKit.Operation = .create) -> SyncCarbObject {
+        // Generate a new UUID if id is nil to prevent crash
+        let safeUUID: UUID
+        if let id = id, let uuid = UUID(uuidString: id) {
+            safeUUID = uuid
+        } else {
+            safeUUID = UUID()
+            debug(.service, "Warning: CarbsEntry has nil or invalid ID, generating new UUID for Tidepool sync")
+        }
+
         SyncCarbObject(
             absorptionTime: nil,
             createdByCurrentApp: true,
             foodType: nil,
             grams: Double(carbs),
             startDate: createdAt,
-            uuid: UUID(uuidString: id!),
+            uuid: safeUUID,
             provenanceIdentifier: enteredBy ?? "Trio",
             syncIdentifier: id,
             syncVersion: nil,


kingst (commented):
Your change looks reasonable, but I'm curious -- how can we get carb entries that are missing an ID? Also, from looking at the return type the uuid property is optional, so we can just leave it as `nil` if ID isn't set. I'm not sure when this code gets called, but if we create new UUIDs each time it's converted it might mistake the same carb entry as different due to the unique UUIDs.

kingst (changes requested):
added suggestion to PR...

Sjoerd-Bo3:
Also true:

I'm not sure when this code gets called, but if we create new UUIDs each time it's converted it might mistake the same carb entry as different due to the unique UUIDs.

But how to fix something like this then? Just catch it and ignore the id less item?

kingst:
I think all you need is a one-line code change. Replace this from the current code:

```swift
SyncCarbObject(
            absorptionTime: nil,
            createdByCurrentApp: true,
            foodType: nil,
            grams: Double(carbs),
            startDate: createdAt,
            uuid: UUID(uuidString: id!),
```

With this:

```swift
SyncCarbObject(
            absorptionTime: nil,
            createdByCurrentApp: true,
            foodType: nil,
            grams: Double(carbs),
            startDate: createdAt,
            uuid: id.flatMap({ UUID(uuidString: $0) }),
```

You'd need to test it to make sure that Tidepool can handle carbs with no uuid set, but that property is an optional so hopefully it "just works" and Tidepool has some other way to identify carb entries uniquely when the UUID isn't set.

dnzxy:
Status here?

dnzxy:
Status here few months later?

dnzxy:
@Sjoerd-Bo3 I will leave this up until Sunday, December 14. If there is still no progress on it by then, I will close this as stale.

PR #710, opened 2025-07-07
fix: Remove redundant backgroundContext.perform calls in ContactImageManager

### LINKED ISSUE #655: BaseContactImageManager.fetchlastDetermination() EXC_BAD_ACCESS possibly crashing chart
## Describe the bug

When navigating to the blood-glucose chart in the Trio iOS app, the app crashes immediately. Instead of seeing the interpolated glucose-level plot, the process terminates with a low-level memory fault (`EXC_BAD_ACCESS` in `libsystem_platform.dylib` during a `memmove`), deep inside the Charts framework’s axis-mark initialization.

## Attach a Log

I’ve attached the full Crashlytics stack trace from the June 5 crash session:
[org.nightscout.trio_issue_0a959a416a747627e4ce0961765d32d7_crash_session_7f01ba5e4f0a4b7495189c4b1f1a06ee_DNE_0_v2_stacktrace.txt](https://github.com/user-attachments/files/20697603/org.nightscout.trio_issue_0a959a416a747627e4ce0961765d32d7_crash_session_7f01ba5e4f0a4b7495189c4b1f1a06ee_DNE_0_v2_stacktrace.txt)

## To Reproduce

None, intermittent.

## Expected behavior

Not crashing.

### Smartphone:

* Hardware: iPhone 15 Pro
* OS Version: iOS 17.5

### Pump:

* Manufacturer: Insulet
* Model: Omnipod Dash

### CGM:

* Device: Dexcom G7
* Manager app: Dexcom App

### Trio Version:

* Version Number: 0.5.0.18
* Repo: nightscout/trio
* Git Reference: *commit hash unknown*

## Technical Details

* **Crash type**: `EXC_BAD_ACCESS` in `_platform_memmove`
* **Faulting frame**:

  ```
  0  libsystem_platform.dylib       0xc5c  _platform_memmove + 428  
  1  libswiftCore.dylib             0x37b88c swift_cvw_initWithCopyImpl(...) + 616  
  2  Charts                         0x6a174 _Charts_getAxisMarkWitnessTable + 100296  
  …  
  ```
* **Likely cause**: corrupted or unexpectedly large Swift value during copy-in of axis-mark metadata in the Charts framework, triggering an out-of-bounds memory move&#x20;

Sjoerd-Bo3:
## Summary
- Fixes ContactImageManager crash in issue #655 by removing redundant Core Data context operations
- Eliminates memory access violations in BaseContactImageManager.fetchlastDetermination() at line 101

## Problem
The crash was caused by redundant context operations where `fetchEntitiesAsync()` already processes results on the background context, but the code was calling `backgroundContext.perform` again, leading to potential context invalidation issues.

## Solution
- Removed unnecessary `backgroundContext.perform` blocks in `fetchlastDetermination()`
- Removed unnecessary `backgroundContext.perform` blocks in `fetchGlucose()`
- Simplified the code while maintaining proper error handling

## Test plan
- [x] Build successfully without compilation errors
- [ ] Verify ContactImageManager functions work without crashes
- [ ] Test contact image updates don't cause memory access violations

## Note
This PR addresses the ContactImageManager crash from issue #655. The Charts rendering crash mentioned in the same issue will be addressed in a separate PR.

### DIFF
--- a/Trio/Sources/Services/ContactImage/ContactImageManager.swift
+++ b/Trio/Sources/Services/ContactImage/ContactImageManager.swift
@@ -98,13 +98,12 @@ final class BaseContactImageManager: NSObject, ContactImageManager, Injectable {
             fetchLimit: 1
         )
 
-        return try await backgroundContext.perform {
-            guard let fetchedResults = results as? [OrefDetermination] else {
-                throw CoreDataError.fetchError(function: #function, file: #file)
-            }
-
-            return fetchedResults.map(\.objectID)
+        // The results are already processed on the background context, no need for another perform block
+        guard let fetchedResults = results as? [OrefDetermination] else {
+            throw CoreDataError.fetchError(function: #function, file: #file)
         }
+
+        return fetchedResults.map(\.objectID)
     }
 
     private func fetchGlucose() async throws -> [NSManagedObjectID] {
@@ -117,13 +116,12 @@ final class BaseContactImageManager: NSObject, ContactImageManager, Injectable {
             fetchLimit: 3 /// We only need 1-3 values, depending on whether the user wants to show delta or not
         )
 
-        return try await backgroundContext.perform {
-            guard let glucoseResults = results as? [GlucoseStored] else {
-                throw CoreDataError.fetchError(function: #function, file: #file)
-            }
-
-            return glucoseResults.map(\.objectID)
+        // The results are already processed on the background context, no need for another perform block
+        guard let glucoseResults = results as? [GlucoseStored] else {
+            throw CoreDataError.fetchError(function: #function, file: #file)
         }
+
+        return glucoseResults.map(\.objectID)
     }
 
     private func getCurrentGlucoseTarget() async -> Decimal? {


kingst (changes requested):
I'm not an expert on CoreData, but don't we need `perform` calls for thread safety? At least that's what I was told when I made this same mistake :)

Sjoerd-Bo3:
Mmm, maybe you're right. I just asked Claude what it thought of this fix and look for info on the web, and maybe my .NET brain thought about this too fast...

```  The Current PR is Incorrect

  KIngst's concern is valid. While the PR removes redundant perform blocks, it creates a thread safety violation
  by accessing Core Data objects outside their context.

  Recommended Fix

  The code should be refactored to properly handle Core Data thread safety:

  private func fetchlastDetermination() async throws -> [NSManagedObjectID] {
      return try await backgroundContext.perform {
          let request: NSFetchRequest<NSManagedObjectID> = NSFetchRequest(entityName: "OrefDetermination")
          request.predicate = NSPredicate(format: "deliverAt >= %@", Date.halfHourAgo as NSDate)
          request.sortDescriptors = [NSSortDescriptor(key: "deliverAt", ascending: false)]
          request.fetchLimit = 1
          request.resultType = .managedObjectIDResultType

          return try backgroundContext.fetch(request)
      }
  }

  This approach:
  1. Fetches only NSManagedObjectIDs directly, avoiding object materialization
  2. Keeps all Core Data operations within the perform block
  3. Eliminates the double-perform issue
  4. Maintains proper thread safety

  Alternative Approach

  If the current fetchEntitiesAsync API must be used, ensure it returns object IDs directly rather than managed
  objects, or modify it to handle the ID extraction internally.

  The memory access violation likely occurs because Core Data objects are being accessed across thread boundaries
   when the Charts framework tries to render data fetched on a background context.```
   
   Still need to verify this ofc. So will do this later, putting back to draft.

marv-out:
We cannot remove this special perform call here since the map operation needs to access the nsmanagedobject on the same context. This change will most likely introduce a crash when the map operation gets executed (with core data multithreading flag enabled) 

Moreover a  “redundant” perform call won’t cause any problems. You can’t have too many perform calls :) there are some cases where you won’t need it but here it is needed. I made the attempt to introduce an actor that automatically ensures thread safety for core data without the perform calls to achieve concurrency operations similar to swift data and it worked, but isn’t in use. So long story short, we need it xD

dnzxy:
Given @marv-out 's comment, I'll close this as not being an acceptable solution.

PR #712, opened 2025-07-08
fix: Accidentally tapping on suffix of a field does not focus input

### LINKED ISSUE #660: Accidentally tapping on suffix of a field does not focus input
Currently when tapping on a carbs field in the bolus calculator (i.e. "0 g"), the focus may actually go to the "g" text, as they're so close together. Currently nothing will happen when you do this, which is frustrating.

Send focus to the input field when tapping suffix text.

Do nothing- workaround is tap to the left of the "0 g", however, that's counterintuitive, especially for new and infrequent users, such as carers.
Alternatively, we could put the value and the suffix in a dark rounded rectangle box, like with the recommended bolus, and focus based on that. However this may lead to visual clutter.

## Additional context
**I'm happy to implement the solution.**

Image

## Technical Details
Could be implemented for just the bolus calculator on a field by field level first just to trial it, however that would lead to inconsistencies in the UX. Could be just for weight / "g" input first but what other fields should be considered? Appears logical to create a new view that supports this use case and reuse that throughout the UI.

## User Impact
This is currently frustrating new and infrequent users, such as carers.

aidanlane:
## Summary
Fixes the "Accidentally tapping on suffix of a field does not focus input" from issue #660.

## Solution
- Extended TextFieldWithToolBar to add optional suffixText (with optional suffixForegroundColor). When this suffixText is tapped, then focus will be sent to the text field.

## Test plan
- [x] Build successfully without compilation errors
- [x] Unit tests run successfully
- [x] Verify (1) that tap gestures send focus to text field, (2) input values are persisted, and (3) live changes work:
  - ~~[ ] Calibrations -> Meter glucose - dead code?~~
  - [x] CarbEntryEditorView -> Carbs
  - [x] CarbEntryEditorView -> Protein
  - [x] CarbEntryEditorView -> Fat
  - [x] DataTable -> New Glucose
  - ~~[ ] ManualTempBasal -> Amount - unsure how to test~~
  - [x] Treatments -> Carbs
  - [x] Treatments -> Protein
  - [x] Treatments -> Fat
  - [x] Treatments -> Bolus
  - [x] AddMealPresetView -> Carbs
  - [x] AddMealPresetView -> Protein
  - [x] AddMealPresetView -> Fat
- [x] Multi day use in-vivo

## Note
This PR addresses the "Accidentally tapping on suffix of a field does not focus input" from issue #660.


bjornoleh (approved):
Approved based on testing. I could test the same items that you had checked, but not libre calibration or manual temp basal. 

No code review on my my part.

bjornoleh:
I think perhaps we don’t use the input field for manual temp basal any longer? In Trio 0.2, this could be set when in open loop, a dedicated button appeared on the bottom of the screen then. If so, the manual temp basal code should probably be removed (not necessarily in this PR). 

Regarding Libre, there used to be a Libre simulator one could choose when setting up Libre as CGM. I think perhaps this has been hidden behind a feature flag now?

aidanlane:
Thanks @bjornoleh 

It indeed looks like manual temp basal is dead code. I was able to remove the files and then there were only a few dead code looking references left that were quick to remove. After that it complied and ran fine. That said, I would suggest another PR and test plan for that, just to keep things clean, test surface minimal, and changes traceable.

I’ll have to dig deeper for the libre simulator.

aidanlane:
>Regarding Libre, there used to be a Libre simulator one could choose when setting up Libre as CGM. I think perhaps this has been hidden behind a feature flag now?

Do you know @kingst? (picking on you as you made the last commit to https://github.com/LoopKit/LibreTransmitter 😉)

bjornoleh:
There’s some info about hidden simulators behind a debug flag here, but this is actually still a draft PR, so this isn’t merged. 
https://github.com/nightscout/Trio/pull/573#issue-3088645414

aidanlane:
Oh, that just appears to allow disabling of the main simulators, not the ability to enable the Libre Demo one.

dnzxy:
Status here?

dnzxy:
Calibrations view can be tested by hard-coding the CGM to be of type Libre afaik.
The manual temp basal input is no longer in existence.  There used to be a dialog / popover that could be opened from home view's navigation. You can still check that out if you build 0.2.x and do not close the loop IIRC.

bjornoleh (approved):
I'd say that testing the input field for Libre calibration isn’t really needed. Seems like a low risk, zero consequence thing to merge this as is.

aidanlane:
Hey @dnzxy, is there anything left blocking this from being merged in?

dnzxy:
Probably not. I have not tested this at all though.

dnzxy (changes requested):
All values for the property `unitsText` need to be the appropriate `String(localized: '...')` that are used throughout Trio.

I realize, you only changed / adjusted what was present, but reviewing it again now, this is actually an issue for stuff like "U/hr", because various languagues abbreviate this differently. 

Please make this final adjustment, then this is good to go.

aidanlane:
Hi @dnzxy, could you please review the string localizations please?

Also, have you a preferred approach for the string that set at runtime:
`unitsText: state.units.rawValue`
https://github.com/nightscout/Trio/pull/712/files#diff-165eb2712d13497fa2093a18e9f0a55f990d2d2e7f0e52c35fd2b48d80450f2eR52

dnzxy (approved):

dnzxy:
I believe there‘s no localized version of that. I‘d go ahead and merge this one now, pending unit tests run.

Thank you.

aidanlane:
Fantastic, thanks @dnzxy

PR #714, opened 2025-07-10
Adds handling for plugin CGM managers to backfill glucose

kingst:
This PR adds support for handling plugin CGM manager backfill glucose readings. In the previous approach, CGM readings from BLE and fetched from the network both go through a Combine pipeline that is invoked using polling. Although this methodology works for network-fetched glucose, it is a poor fit for the event-driven nature of BLE glucose readings.

To address this issue, this PR enables CGM managers to send glucose readings directly to the FetchGlucoseManager, bypassing the Combine pipeline. This change is useful both pragmatically, as Combine is awkward for this type of communication, and theoretically as we don't need to poll for event-driven BTE CGM readings since the driver already notifies Trio when it has a new value, or backfill readings.

To test, disable Bluetooth on your device and let enough time go by for it to miss a CGM reading. Turn Bluetooth back on and confirm that the missed reading gets added.

I am running this in-vivo currently and so far so good.

### DIFF
--- a/Trio/Sources/APS/CGM/PluginSource.swift
+++ b/Trio/Sources/APS/CGM/PluginSource.swift
@@ -15,8 +15,6 @@ final class PluginSource: GlucoseSource {
 
     var cgmHasValidSensorSession: Bool = false
 
-    private var promise: Future<[BloodGlucose], Error>.Promise?
-
     init(glucoseStorage: GlucoseStorage, glucoseManager: FetchGlucoseManager) {
         self.glucoseStorage = glucoseStorage
         self.glucoseManager = glucoseManager
@@ -34,25 +32,12 @@ final class PluginSource: GlucoseSource {
     /// - Parameter timer: An optional `DispatchTimer` (not used in the function but can be used to trigger fetch logic).
     /// - Returns: An `AnyPublisher` that emits an array of `BloodGlucose` values or an empty array if an error occurs or the timeout is reached.
     func fetch(_: DispatchTimer?) -> AnyPublisher<[BloodGlucose], Never> {
-        Publishers.Merge(
-            callBLEFetch(),
-            fetchIfNeeded()
-        )
-        .filter { !$0.isEmpty }
-        .first()
-        .timeout(60 * 5, scheduler: processQueue, options: nil, customError: nil)
-        .replaceError(with: [])
-        .eraseToAnyPublisher()
-    }
-
-    func callBLEFetch() -> AnyPublisher<[BloodGlucose], Never> {
-        Future<[BloodGlucose], Error> { [weak self] promise in
-            self?.promise = promise
-        }
-        .timeout(60 * 5, scheduler: processQueue, options: nil, customError: nil)
-        .replaceError(with: [])
-        .replaceEmpty(with: [])
-        .eraseToAnyPublisher()
+        fetchIfNeeded()
+            .filter { !$0.isEmpty }
+            .first()
+            .timeout(60 * 5, scheduler: processQueue, options: nil, customError: nil)
+            .replaceError(with: [])
+            .eraseToAnyPublisher()
     }
 
     func fetchIfNeeded() -> AnyPublisher<[BloodGlucose], Never> {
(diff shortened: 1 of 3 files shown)


MikePlante1:
I tested this in-vivo with the Dexcom G6 app, and it backfills great for both disabled bluetooth as well as closed-trio/open-dexcom... but I'm getting multiple double entries since installing this.

IMG_7800

[log 27.txt](https://github.com/user-attachments/files/21169014/log.27.txt)

mountrcg:
Great backfill while testing on g6, I have no dups like Mike.

marionbarker:
## Summary - Test with G7

I also see a duplicate glucose with G7. Otherwise the backfill appears to work as expected. I did not review the code.

### Details

Build this branch onto my personal (Loop) phone and attach my G7 to Trio (as well as Loop).
I have done this before with no issues - both apps seem to work fine in parallel.
* The Loop app is connected to my G7 and DASH pod
* The Trio app is connected to my G7 and in-app pump simulator

Keep the phone connected to Xcode so I can see log files in real time.
Wait for 3 G7 readings to come into Trio and get one loop.
Disable Bluetooth - wait 20 minutes then enable again.
The G7 should backfill on next reading.

It backfilled and I got one duplicate (the last reading before turning off Bluetooth) was repeated.

Saved the log file and some screenshots.
* [20250710_Trio_PR714_Xcode_log.txt](https://github.com/user-attachments/files/21172000/20250710_Trio_PR714_Xcode_log.txt)

Left: Glucose before disabling Bluetooth; Right: Glucose after enabling Bluetooth
trio_pr0714_screenshots

The log for Loop shows the same messages as the Trio log, starting at 2025-07-10 22:21:53 in Loop (which reports UTC) and in Trio (2025-07-10 15:21:52.768774-07:00) which reports in local time and includes microsec.

kingst:
@MikePlante1 and @marionbarker I just pushed a fix for the deduplication check -- it was using dates for matching and it was losing precision somewhere. So I check if there are any current values within 1 second for matches and it's filtering out the duplicates now in my testing.

MikePlante1:
Installed the updated PR and ran the following test:

* Turned Bluetooth off to miss one reading
* Then turned Bluetooth back on but swiped Trio closed and missed one more reading.
* I then opened Trio right before the next reading came in and it read the current reading and backfilled one reading, but not the first missed reading.
* I then deleted the two most recent readings in Trio (so missing 3 readings now) and waited until the next reading, which read the current reading but only backfilled one reading.

Is it possible to extend this to backfill more than just one missed reading? Or is that the limit?

I then installed latest dev with PR #703 reverted so I could backfill from Share, and then reinstalled this PR. I did not see any double readings in my testing since the "fix filtering logic" commit, but will continue to run this PR in-vivo. I might do some further testing tomorrow.

kingst:
I don't know the details of how the LoopKit driver backfills data and I didn't have to configure anything as it was already sending us these values -- we were just ignoring them due to the way we had our poll / Combine setup previously. So all this PR does is process all of the glucose values we were already getting from the CGM driver.

But since we're getting this data directly from the transmitter, in my testing when I leave bluetooth off for multiple readings then turn it back on it'll backfill all of them.

marionbarker:
## Test Summary

Successful test. I approve from testing.

### Configuration

Run both Loop and Trio, commit `e44909c0e`, branch: `backfill-glucose-from-cgm-manager` on iPhone 15, iOS 18.5 with Dexcom G7 app v2.9.0.

### Test 1 - quit Dexcom G7 only, leave Loop and Trio running

Times below are local time (-7 UTC) on 13 July 2025.

* Record the 11:16 am reading for G7, 89 mg/dL
* Quit the Dexcom G7 app a 11:20, leave trio and loop running
   * Notified that Dexcom G7 will get no sensor readings or alerts until it is open
   * Loop has Searching for Sensor icon at 11:23
   * Trio shows a gap in glucose with no forecast plotted
   * Wait 15 minutes and then restart the Dexcom G7 app
   * Get notification from both Loop and Trio that they are not looping
* 11:38 Open Dexcom G7 app and see Signal Loss warning
   * by the time I recorded that, it has backfilled showing current glucose of 90 mg/dL
   * next reading should be at 11:41
      * both Loop and Trio picked up 91 mg/dL at 11:41 and did not fill in the gap
      * and Dexcom G7 also picked up the 91 mg/dL
   * next reading should be at 11:46
      * Dexcom gets 93 mg/dL
      * Loop and Trio also pick up the reading with no backfill
* The gap, from 11:21 through 11:36, inclusive, when Dexcom G7 was not operating, remains for both Loop and Trio
 
### Test 2: Disable Bluetooth while Dexcom G7, Loop and Trio are running

Disable BLE at 11:50.
Wait 15 minutes.
Enable BLE at 12:10
* next reading should be at 12:11
   * Dexcom G7: 101
   * Loop 101 and filled in
   * Trio 101 and filled in

Confirm there are no duplicated glucose in Trio.

I believe the updated commit fixed the duplicated glucose problem.

dnzxy (approved):
Approving based on in-vivo testing with G6 + Anubis + official Dexcom app.

PR #718, opened 2025-07-14
Decoupling IoB from determine basal

kingst:
This PR fixes a regression that happened after this [PR](https://github.com/nightscout/Trio/pull/590) landed and it is the case again where people who are stuck at 400 mg/dl will not get updates to their IoB. This lack of updates can be problematic since people will dose while they are at 400 manually and they need to know their IoB to do this safely.

The approach is to decouple the display of IoB values from determine basil. IoB will update if (1) there is a new pump event (2) there is a new determination, or (3) 5 minutes has elapsed since the last IoB update and the app is in the foreground.

There are two optimizations to ensure the IoB calculation doesn't get called too often: limiting the update frequency from timer events and only honoring timer events when the app is in the foreground.


kingst:
I'm going to close this as it requires a much larger architectural change. In particular, these following modules all need IoB information and depend on determinations (in addition to the home view).

Listen for changes to "OrefDetermination" in CoreData:
- `Trio/Sources/Services/WatchManager/AppleWatchManager.swift`
- `Trio/Sources/Services/WatchManager/GarminManager.swift`
- `Trio/Sources/Services/ContactImage/ContactImageManager.swift`
- `Trio/Sources/Services/LiveActivity/LiveActivityManager.swift`

And these fetch it on demand (still using OrefDetermination records from CoreData):
- `Trio/Sources/Services/Calendar/CalendarManager.swift`
- `Trio/Sources/Services/RemoteControl/TrioRemoteControl+Bolus.swift`
- `Trio/Sources/Shortcuts/State/StateIntentRequest.swift`

The architectural change is to have a new IobManager that can notify these components when IoB changes and has a query interface to get the latest IoB or historical IoB values.

PR #720, opened 2025-07-16
chore: update DanaKit

bastiaanv:
Changelog:
- fix: notifyDelegateOfDeactivation correctly implemented now
- chore: removed force unwrapping at DanaKitHUDProvider.swift
- feat: add `replaceComponent` PumpEvent for PR #682 

Tagging: @MikePlante1


MikePlante1 (approved):

marionbarker:
I do not think this was sufficiently tested before merging.

I ran into problems testing the addition of this version of DanaKit to LoopWorkspace.

Please see [feat:/add DanaKit](https://github.com/LoopKit/LoopWorkspace/pull/247#issuecomment-3148667626)

As part of the troubleshooting, I decided I should test Trio as well.

## Trio results

Because Trio uses the same repository as LoopWorkspace, I repeated this test with Trio 0.5.1.10.

### Trio using dev (`59f4853`)

The signature is a little different - there is not a crash, but the pump is not added. After tapping on Finish, the user returns to the main screen with the Add Pump icon showing. Tapping on that icon brings up the Add Pump menu.

### Trio using `41b592e`

When building with the commit (`41b592e`) before the problem commit (`dcc4495`), the Dana pump can be added or removed from Trio as expected.

marionbarker:
I added one more test:

* I built the latest onto an instance where Trio was already running on a Dana pump and there was no issue. 

This Issue is when adding or deleting a Dana pump.

PR #735, opened 2025-08-05
Add steps in Onboarding for uploading to Nightscout

### LINKED ISSUE #624: Onboarding: Uploading to Nightscout is not enabled when users establish connection

MikePlante1:
Adds two settings, `Allow Uploading to Nightscout` and `Upload Glucose`, to the Onboarding.

https://github.com/user-attachments/assets/71827339-bb9f-4ef6-8926-1d825bdf3b87

If you choose to Setup Nightscout, it will default them both to ✅ but allow toggling either off.
If you choose to Skip Nightscout Setup, it will use whatever settings were selected in 0.2.x if updating, or the defaults (Upload to NS ❎ and Upload Glucose ✅) if it's a fresh install.


Trio/Sources/Modules/NightscoutConfig/View/NightscoutUploadView.swift
```diff
@@ -35,7 +35,7 @@ struct NightscoutUploadView: View {
                 VStack(alignment: .leading, spacing: 10) {
                     Text("Default: OFF").bold()
                     Text(
-                        "The Upload Treatments toggle enables uploading of the following data sets to your connected Nightscout URL:"
+                        "This toggle enables uploading of the following data sets to your connected Nightscout URL:"
```
dnzxy: I'd suggest not to change this, as the pre-change text is not necessarily wrong, and part of many localizations already.

Trio/Sources/Modules/Onboarding/OnboardingStateModel.swift
```diff
@@ -100,6 +100,8 @@ extension Onboarding {
         var isConnectedToNS: Bool = false
         var nightscoutImportError: NightscoutImportError?
         var nightscoutImportStatus: ImportStatus = .none
+        var isUploadEnabled: Bool = true
+        var uploadGlucose: Bool = true
```
dnzxy, line 104: `shouldUploadGlucose`

dnzxy (commented):
Only reviewed code, haven't run this.
Seems mostly okay, some small nitpicks.
What happens if you were to navigate back? Does it properly navigate back? Are you not overwriting the states for those 2 settings?

Trio/Sources/Modules/NightscoutConfig/View/NightscoutUploadView.swift
MikePlante1: I find it confusing as there is no toggle named “Upload Treatments”, but I’ll revert it.

Trio/Sources/Modules/Onboarding/OnboardingStateModel.swift
MikePlante1: Shouldn’t it match the name of the variable in Settings, though?

MikePlante1:
> Only reviewed code, haven't run this. Seems mostly okay, some small nitpicks. What happens if you were to navigate back? Does it properly navigate back?

Yes, it navigates back normally.

https://github.com/user-attachments/assets/d49de58b-10ba-44d4-ad71-ae45450dad13

> Are you not overwriting the states for those 2 settings?

Not sure what you're asking. When `applyToSettings()` is called, if the user chose to setup Nightscout in Onboarding, then those 2 settings are set from the toggles chosen in the onboarding. `applyToSettings()` is only called from `saveOnboardingData()` which is only called once onboarding is completed.

dnzxy:
> Not sure what you're asking. 

Navigate to the login step. Login. Navigate further. Options will default to on. Toggle both off. Navigate further, then back again to the login step. Then further again. Will it properly keep the toggled off state or will it override?

MikePlante1:
Just synced this to the latest `dev`.

> > Not sure what you're asking.
> 
> Navigate to the login step. Login. Navigate further. Options will default to on. Toggle both off. Navigate further, then back again to the login step. Then further again. Will it properly keep the toggled off state or will it override?

It keeps the toggled off state. I added a video to the previous comment to document that.

> I'd suggest not to change this, as the pre-change text is not necessarily wrong, and part of many localizations already.

This has now been reverted to the previous text.

> shouldUploadGlucose

I thought it better to match the local variable's name with the actual variable's name from the settings, and I think changing the name of the actual variable for the settings would probably have too many consequences. But I will update the local variable of `uploadGlucose` in onboarding to this if you think that's better.

dnzxy (approved):

Issue #737, opened 2025-08-06
CGM Backfill logic causes issues when you delete entries or change CGMs


kingst:
The current CGM backfill logic isn't sophisticated enough to cope with CGM managers that send historical values. Historical values can happen on each loop with G6 and potentially xdrip. The fundamental problem with the current algorithm is that (1) it will add deleted glucose values back and (2) it will produce duplicate values if the new device has data that overlaps with your old one and the CGM manager sends historical data.

To fix it, we need to keep track of deleted glucose values and have more nuanced logic to only backfill CGM values when there is a gap.

(1) To reproduce, uncomment the backfill logic
(2) install Trio on a fresh simulator
(3) use Nightscout CGM and wait until you get some data
(4) switch to the glucose simulator

You'll see a bunch of overlapping data.

t1dude:
Not sure how this is progressing, but it might be interesting to have a look at how this is solved in AndroidAPS. Whenever readings are too close due to backfill issues, overlapping sensors, or changing source, the user will get a notification and be urged to clean up the readings. There's a separate view for readings and a "delete mode" where readings that are too close to each other are highlited and easy to mark for deletion. I'm not saying this is a perfect solution, just wanted to point out that this is a way to have the user solve the issue manually when it occurs.

kingst:
Hi @t1dude I haven't taken a look at the AAPS experience, I was planning to handle it algorithmically. I think there are a few key cases where we get duplicates that we can detect and handle correctly. But, I do think that the AAPS experience sounds interesting and something we should do if the automatic approach doesn't work.

kingst:
Removing stale label -- I'm going to take this on but need a few more weeks before I'll have the time

MikePlante1:
Resolved in #851

PR #738, opened 2025-08-06
Comment out CGM backfill logic

kingst:
This PR temporarily removes the backfill logic. Backfilling was causing issues with CGM managers that send historical values commonly, like G6 and potentially xdrip. In particular, it would add duplicate readings and would re-add deleted CGM values.

For now I'm going to disable this logic, but will fix it soon https://github.com/nightscout/Trio/issues/737

### DIFF
--- a/Trio/Resources/InfoPlist.xcstrings
+++ b/Trio/Resources/InfoPlist.xcstrings
@@ -457,18 +457,6 @@
         }
       }
     },
-    "NSCalendarsFullAccessUsageDescription" : {
-      "comment" : "Privacy - Calendars Full Access Usage Description",
-      "extractionState" : "extracted_with_value",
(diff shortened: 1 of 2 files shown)


t1dude (commented):
You seem to be changing calendar access as well. Is that intentional?

kingst:
@t1dude One of our build scripts changed the `InfoPlist.xcstrings` this shouldn't impact access only the strings we show. To be honest I usually let localization changes through without scrutinizing them too much and if you want me to back that change out I can, but we'll need to figure out why our build script is making the change.

dnzxy (approved):
Not sure about the InfoPlist and Localizable changes, but since this is a semi-critical issue, approving to get along.

PR #741, opened 2025-08-09
fix(DanaKit): prevent pump deactivation after onboarding

### LINKED ISSUE #733: Pairing Dana-i in Trio v0.5.1.10 (1) completed but pump not shown
I started pairing a Dana-i pump an within process i entered 6 digits shown on pump and Trio shows finish message and pump display shows bluetooth data exchange, but pump is not visible in section devices. 

I tried it several times now, always with deleting pump from iPhone bluetooth device list, where it was always shown. 
I restarted iPhone in between too. 

iOS 18.5, iPhone 14 Pro
Trio built today from dev branch. 

Dana-i formerly run well in aaps setup on android phone where pump connection is deleted. 

What can i try to get pump connected?

Image
Image
Image
Image

bastiaanv:
closes #733 

Context:
PR #720 changed the way DanaKit handles offboarding/removing the pumpKit. Now when a user tries to onboard the pump, this offboarding got triggered.

This PR fixes that by now directly calling the code to dismiss the modal, instead of the stepFinished function (which can invoke off-boarding)


dnzxy:
Could you share some context please? The Onboarding workflow is not manipulating any pump state.

marionbarker:
@bastiaanv 

I updated loopandlearn:DanaKit:dev to be same as bastiaanv:DanaKit:dev - whereas you had added the same code to the two repositories independently (and thus the same code had 2 different SHA). 

Please update this PR to use the SHA, `910e339` from loopandlearn (which is now the same SHA as bastiaanv for the same code).

I will be testing the actual code modification next.

marionbarker:
When using the updated SHA (from loopandlearn, `910e339`), this PR works as expected when testing with an rPi dana-simulator. 

* [y] Able to add Dana pump
* [y] Able to sync to pump
* [y] Able to achieve a green loop with Trio
* [y] Able to remove Dana pump and add a different pump (simulator pump)
* [y] Able to remove different pump and add Dana pump successfully

Thanks for the fix.

dnzxy (approved):

PR #764, opened 2025-09-07
Improve tabBar functionality for treatment sheet selection

MikePlante1:
When Trio is built to iOS 26 Beta with Xcode 26 Beta, it uses the new liquid glass for the tabBar. This creates an issue where the spacer in the middle of the tabBar can be accidentally selected instead when trying to open the Treatment sheet, which brings up a completely empty view instead of the Treatment sheet.

This PR makes it so if the middle tabBar slot is accidentally selected instead of the Treatment sheet button, it will revert back to the currently selected tab and open the Treatment sheet.

Tested using Xcode 26 Beta 7 with iOS 18.5 and iOS 26.0 simulators.

### DIFF
--- a/Trio/Sources/Modules/Home/View/HomeRootView.swift
+++ b/Trio/Sources/Modules/Home/View/HomeRootView.swift
@@ -1063,9 +1063,12 @@ extension Home {
                 .tint(Color.tabBar)
                 .onChange(of: selectedTab) { oldValue, newValue in
                     if newValue == 2 {
-                        // Middle tab was selected – revert to previous tab and open sheet
-                        selectedTab = oldValue
+                        // Middle tab was selected - open treatment sheet and after delay revert to previous tab
                         state.showModal(for: .treatmentView)
+                        Task {
+                            try await Task.sleep(for: .seconds(1))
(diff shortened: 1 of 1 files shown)


AndreasStokholm:
This technically works and solves the issue. But from a UX perspective it might be very confusing to see the tabbar selector go back to main and then opening the sheet.

Additionally the positioning of the + feels very off with Liquid Glass.

MikePlante1:
> This technically works and solves the issue. But from a UX perspective it might be very confusing to see the tabbar selector go back to main and then opening the sheet.

I just pushed a slight delay to the revert to hopefully improve this.

> Additionally the positioning of the + feels very off with Liquid Glass.

I agree it looks "off" with Liquid Glass, but I didn't touch the positioning in this PR. 

Screenshot 2025-09-08 at 09 00 09

__________________

But also, I just found a way to disable Liquid Glass, so I'll probably close this PR in favor of that one.

AndreasStokholm:
The delay works well, nice fix!

However, I think just sticking to the non-liquid glass design is the best option for now, until the tabbar can have a rework that works better with the floating design for liquid glass.

PR #776, opened 2025-09-23
upgrade rexml per dependabot security warning

marionbarker:
See warning here: https://github.com/nightscout/Trio/security/dependabot/8


marionbarker:
## Test

Confirm build works as expected:
* ✅ using Xcode 16.4 onto an iOS 26 phone
* ✅ using [Browser Build](https://github.com/docs-test/Trio/actions/runs/17954955892)

dnzxy (approved):

PR #781, opened 2025-09-27
Fixes #778 — ISF value not saved

### LINKED ISSUE #778: ISF value not saved
## Describe the bug
When I select the ISF value and change it then it doesn't select the new value it still remembers the old value. But if you add the same time schedule that already exists  for instance 12:00 is 3,5 mmol/L/U and add 12:00 3,0 mmol/L/U then the old value will change after it's saved. Very odd behavior. The version of Trio I use is v0.5.1.25(1). 

## Attach a Log
*Tap the Trio settings icon at the bottom of the screen, then tap 'Share logs' on the bottom of the list and attach it to this ticket.*

## To Reproduce

1. Go to 'Settings'
2. Click on 'Therapy '
4. Click on 'Insulin Sensitivities'
5. Tap an existing time schedule and change the ISF value. The value won't change!! 
6. Tap back and open the same time schedule as described in step 4

https://github.com/user-attachments/assets/b3ae976a-d415-4351-87d7-265b229245bc

7. The values hasn't changed
8. Go back
9. Go to the same time schedule the rate shows the value you have changed.
10. Go back and the ISF rate show the old value
11. Go to the same time schedule and the set rate shows the new value
12. Go back and the ISF shows the old value
13. Go back to therapy settings 
14. Select Insulin Sensitivity
15. Select the time schedule with the changed value
16. The Set Rate screen shows the changed value 
17. The choice (highlighted Save button) to Save the new ISF setting won't appear (that's probably logical, the old value is still present).
18. Sometimes it's hard to reproduce but the screencapture is more clear (I think).
19. *See error*## Expected behavior*A clear and concise description of what you expected to happen.*## ScreenshotsIf applicable, add screenshots to help explain your problem.*## Setup Information (please complete the following information):### Smartphone:* Hardware: *[e.g. iPhone 15 Pro]** OS Version: *[e.g. iOS 17.5]*## Pump:* Manufacturer: Insulet* Model: Omnipod Dash

### CGM:
* Device: Dexcom G6
* Manager app: Dexcom App or xDrip4iOS

### Trio Version:
* Version Number: v0.5.1.25(1)
* Repo: *nightscout/trio*
* Git Reference: *[e.g. commit hash]*

## Technical Details

## Additional context

MikePlante1:
Fixes #778 — the issue in Xcode 26 where ISF values were not saved properly.


MikePlante1:
Note: Does not resolve build time automations for NSCalendar or localization or MinimedKit, which were present before this PR.

Trio/Sources/Modules/ISFEditor/ISFEditorStateModel.swift
```diff
@@ -115,11 +115,16 @@ extension ISFEditor {
             DispatchQueue.main.async {
                 DispatchQueue.main.async {
                     let uniq = Array(Set(self.items))
-                    let sorted = uniq.sorted { $0.timeIndex < $1.timeIndex }
-                    sorted.first?.timeIndex = 0
+                    var sorted = uniq.sorted { $0.timeIndex < $1.timeIndex }
+
+                    if !sorted.isEmpty {
```
dnzxy, line 120: Why this change?

Trio/Sources/Modules/ISFEditor/ISFEditorDataFlow.swift
```diff
@@ -3,7 +3,7 @@ import Foundation
 enum ISFEditor {
     enum Config {}
 
-    class Item: Identifiable, Hashable, Equatable {
+    struct Item: Identifiable, Hashable, Equatable {
```
dnzxy, line 6: Why are you only changing this for ISF, but not the other editors?

Trio/Sources/Modules/ISFEditor/ISFEditorStateModel.swift
MikePlante1: > When Item was a class, sorted.first returned an optional reference to the actual object in memory. Even though first itself is read-only, the object it points to is mutable. So you could modify the timeIndex property of that object directly.
> 
> When Item is a struct, sorted.first returns an optional copy of the struct value. Since first is read-only, you get a read-only copy that you cannot modify at all.

Trio/Sources/Modules/ISFEditor/ISFEditorDataFlow.swift
MikePlante1: Because only ISF was broken, but I'll update the others as well.

MikePlante1:
Closing, as this is not a proper fix and could lead to unforeseen consequences. Per dnzxy:

> The proper "fix" to this would probably be to tackle refactoring the settings in its entirety and bring over the handling we use in Onboarding

PR #794, opened 2025-10-06
Shift GitHub to check for updates every Sunday and build 2nd Saturday of each month

marionbarker:
## Purpose

Try to avoid times when GitHub actions are historically busy.
Thanks Carol Vachon for this report:

> The most significant and consistent historic pattern impacting GitHub activity is the week-to-weekend cycle, as most development work happens during the standard workweek.

> * Lowest activity: Weekends, especially Sundays, see the least amount of activity, as most developers are off work.
> * Highest activity: Activity is typically at its peak from Tuesday to Thursday.
> * Moderate activity: Mondays and Fridays have moderate activity as developers ease into and out of the workweek.

On Wed, Oct 1, GitHub actions were impacted and all builds across the Open Source community all failed.

## Method

* Shift the weekly build in case of code updates from Wednesday to Sunday
* Shift the monthly builds, regardless of whether code was updated, from the 1st of the month to the second Saturday of each month

### DIFF
--- a/.github/workflows/build_trio.yml
+++ b/.github/workflows/build_trio.yml
@@ -7,9 +7,9 @@ on:
   #push:
 
   schedule:
-    # avoid starting an action at xx:00 when GitHub resources are more likely to be impacted
-    - cron: "43 8 * * 3" # Checks for updates at 08:43 UTC every Wednesday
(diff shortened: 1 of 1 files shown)


MikePlante1 (approved):
I agree with this scheduling change.

PR #798, opened 2025-10-09
Update AddOverrideForm.swift

quiktom:
extending range for overrides

### DIFF
--- a/Dependencies/OmniBLE/OmniBLE/OmnipodCommon/Pod.swift
+++ b/Dependencies/OmniBLE/OmniBLE/OmnipodCommon/Pod.swift
@@ -79,7 +79,7 @@ public struct Pod {
     // Checked to verify it agrees with value returned by pod during the pairing process.
     public static let cannulaInsertionUnits = 0.5
 
-    public static let cannulaInsertionUnitsExtra = 0.0 // edit to add a fixed additional amount of insulin during cannula insertion
+    public static let cannulaInsertionUnitsExtra = 0.25 // edit to add a fixed additional amount of insulin during cannula insertion
 
     // Default and limits for expiration reminder alerts
     public static let defaultExpirationReminderOffset = TimeInterval(hours: 2)
--- a/Trio/Sources/Modules/Adjustments/View/Overrides/AddOverrideForm.swift
+++ b/Trio/Sources/Modules/Adjustments/View/Overrides/AddOverrideForm.swift
@@ -110,7 +110,7 @@ struct AddOverrideForm: View {
                                 set: { state.overridePercentage = Double($0) }
                             ), label: Text("")
                         ) {
-                            ForEach(Array(stride(from: 40, through: 150, by: percentageStep)), id: \.self) { percent in
+                            ForEach(Array(stride(from: 10, through: 220, by: percentageStep)), id: \.self) { percent in
                                 Text("\(percent) %").tag(percent)
                             }
                         }
--- a/Trio/Sources/Modules/Adjustments/View/Overrides/AddOverrideForm.swift
+++ b/Trio/Sources/Modules/Adjustments/View/Overrides/AddOverrideForm.swift
@@ -110,7 +110,7 @@ struct AddOverrideForm: View {
                                 set: { state.overridePercentage = Double($0) }
                             ), label: Text("")
                         ) {
-                            ForEach(Array(stride(from: 10, through: 220, by: percentageStep)), id: \.self) { percent in
+                            ForEach(Array(stride(from: 10, through: 200, by: percentageStep)), id: \.self) { percent in
                                 Text("\(percent) %").tag(percent)
                             }
                         }
--- a/Trio/Sources/Modules/Adjustments/View/Overrides/EditOverrideForm.swift
+++ b/Trio/Sources/Modules/Adjustments/View/Overrides/EditOverrideForm.swift
@@ -172,7 +172,7 @@ struct EditOverrideForm: View {
                             label: Text("")
                         ) {
                             ForEach(
-                                Array(stride(from: 40.0, through: 150.0, by: Double(percentageStep))),
+                                Array(stride(from: 10.0, through: 200.0, by: Double(percentageStep))),
                                 id: \.self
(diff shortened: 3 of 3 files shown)


dnzxy:
Was this meant as a general PR or as a personal code customization ?

quiktom:
> Was this meant as a general PR or as a personal code customization ?

As general PR. On v0.2.x was range up to 200% on v0.6 only up to 40-150%, which could be a limitation factor for many situations.

Sjoerd-Bo3:
@quiktom If that is your intention, then this needs to be proper: Please adhere to the feature template in the new issue section. We cannot merge something without the proper motivation explanation of steps etc. 

You can do this with opening an issue with the feature you want and linking it to this PR. And explaining in the PR what you changed and which side effects you have though of and how to combat them.

Thanks in advance!

Sjoerd-Bo3:
And now OT: You also need to edit the EditOverrideform.swift:175 There is also a cap there.

(I also have the minimum set to 10% btw on my own build, so personally I'm for it!)

dnzxy:
I'll second Sjoerd in his comments for proper form, proper change log / PR description and providing some context. 
I'll furthermore second that this change is incomplete and is not addressing all necessary instances of percentage thresholds for Overrides.

The maximum threshold changes/modifications that we allow based on what oref allows for Temporary Targets is 15% at 200 mg/dL and 200% at 80 mg/dL when choosing to do a custom modification, not a calculated one.

I am very much opposed to offering users *whatever* target paired with *whatever* percentage adjustment (which is how Overrides *can* be used), so there needs to be a proper way of safeguarding this – i.e., pair the adjustment with "allowed" targets or something to that extent, or something entirely different – before this is considered as a *general adjustment* in Trio.

What the user does in their own personalized builds, like Sjoerd has mentioned, is up to them, but it's a different thing offering this for the general user base.

Sjoerd-Bo3:
> I am very much opposed to offering users _whatever_ target paired with _whatever_ percentage adjustment (which is how Overrides _can_ be used), so there needs to be a proper way of safeguarding this – i.e., pair the adjustment with "allowed" targets or something to that extent, or something entirely different – before this is considered as a _general adjustment_ in Trio.

I agree with this! We need to think this through before changing this simply. Thats also a reason I didnt propose this yet. And only changed this in my build because I dont really use targets

quiktom:
> And now OT: You also need to edit the EditOverrideform.swift:175 There is also a cap there.
> 

Thanks for advice :)

quiktom:
> I'll second Sjoerd in his comments for proper form, proper change log / PR description and providing some context. I'll furthermore second that this change is incomplete and is not addressing all necessary instances of percentage thresholds for Overrides.
> 
> The maximum threshold changes/modifications that we allow based on what oref allows for Temporary Targets is 15% at 200 mg/dL and 200% at 80 mg/dL when choosing to do a custom modification, not a calculated one.
> 
> I am very much opposed to offering users _whatever_ target paired with _whatever_ percentage adjustment (which is how Overrides _can_ be used), so there needs to be a proper way of safeguarding this – i.e., pair the adjustment with "allowed" targets or something to that extent, or something entirely different – before this is considered as a _general adjustment_ in Trio.
> 
> What the user does in their own personalized builds, like Sjoerd has mentioned, is up to them, but it's a different thing offering this for the general user base.

Sure, we can keep it only in personalized builds. I was just considering that many my patients needs (eg in preperiod/period time) more than 150% and in previous version was up to 200%. ;)

dnzxy:
> my patients 

Say what?

Previous versions did not have any guardrails of any sort for any of the settings of Trio. Bad comparison, sorry to say. 

@quiktom , I had listed the maximum allowed adjustments we offer users for temporary targets in my reply — so those can be used theoretically. However, it will need safeguarding around the extremes. You are more than welcome to propose a thorough feature  set like that :) 

Edit to add:
Example for extremes: 
- 10% @ 270 mg/dL
- or, 200% @ 75 mg/dL
- or anything in that realm

dnzxy:
Hi @quiktom 

are you planning on addressing any of the raised points here in the comments? If not, I‘ll proceed to close this PR. 

Side note: you have since also introduced a change to one of our submodules of the LoopKit ecosystem (Omnipod drivers, initial insulin dosing on pod insertion) — this absolutely cannot be pushed here and also doesn’t belong in this PR. 

Regards.

quiktom:
> Hi @quiktom
> 
> are you planning on addressing any of the raised points here in the comments? If not, I‘ll proceed to close this PR.
> 
> Side note: you have since also introduced a change to one of our submodules of the LoopKit ecosystem (Omnipod drivers, initial insulin dosing on pod insertion) — this absolutely cannot be pushed here and also doesn’t belong in this PR.
> 
> Regards.

Hey, yes you can close this PR please. Maybe I will get to it later. 

Interesting that the dosing on insertion showed up here. I didn’t make a pull request, and it’s supposed to be only on my testing dev version.

MikePlante1:
> Example for extremes:
> 200% @ 75 mg/dL

But TTs allow 200% at 80 mg/dL — is that really so much different?

Here's the combos allowed by Temp Targets in the latest Trio-dev:

| Target Glucose | Custom Sensitivity Adjustment |
|---|---|
| 80 – 99 mg/dL | 105 – 200 % |
| 101 – 200 mg/dL | 15 – 95 % |

And here's the combos allowed by Overrides in the latest Trio-dev:

| Target Glucose | Custom Sensitivity Adjustment |
|---|---|
| 72 – 270 mg/dL | 40 – 150 % |

I can see why oref was originally designed like this, but I've always found the 100 mg/dL pivot point kinda strange, and if we implemented the same restrictions for Overrides someone who usually has a target of 120/dL would have to decrease their target by 20 mg/dL if they even wanted a slight increase in their sensitivity.

MikePlante1:
> Interesting that the dosing on insertion showed up here. I didn’t make a pull request, and it’s supposed to be only on my testing dev version.

In the future, please make a new branch before PR'ing it. The problem you ran into here is that you PR'd your `dev` branch, and then made changes to your `dev` branch, which automatically updates the PR.

PR #807, opened 2025-10-13
Implement snooze action for notifications and eliminate duplicate notification

cachrisman:
I used to get several notifications about the same high (or low) glucose alert and I had to tap the notification to open the app to then be prompted to snooze the alert. This PR introduces a lastGlucoseAlertToken to dedup multiple notifications and also introduces notification actions for both iOS and WatchOS to make it even easier to respond to these notifications. I've tested this on my Trio app and the notification actions work for me on both my phone and my watch. I also have only seen one notification rather than 3 or 4 that I used to get.


cachrisman:
Here's what it looks like when you get a notification and tap and hold on it. 

image

And here's a video of the notification on the lock screen showing the long tap and selecting a snooze duration

video

dnzxy:
Hi @cachrisman 
what CGM are you using? Libre by any chance?
May it be that there are just pending stale glucose notifications that are not being cleared properly from earlier instances?

cachrisman:
> Hi @cachrisman what CGM are you using? Libre by any chance? May it be that there are just pending stale glucose notifications that are not being cleared properly from earlier instances?

Hi @dnzxy I'm using Dexcom G7. 

The changes in this PR mitigate the problem I found that duplicates were coming from multiple observers firing for the same reading without a dedupe token, and snooze wasn’t clearing delivered/pending glucose alerts. The patch adds a per‑alert token, clears it when alarms resolve or snooze is set, removes queued alerts on snooze, and wires snooze actions (phone + watch) through categories/quick actions, eliminating stale notifications instead of leaving them pending.

dnzxy:
Thanks for the response @cachrisman .
There is a snoozing function already present in Trio (long press the "glucose bobble" on main screen), that lets the user snooze way longer than this.
How does your PR play into that (I haven't tested it, tbf) and why is the PR only giving the user the option to snooze up to 1 hr?

cachrisman:
@dnzxy good point - I only added in 3 choices in the notification action (20mins, 40mins, and 1 hour) to make the UX simpler when responding to a notification on the watch or from the notification center. There are 15 choices available as detailed `pickerTimes` in SnoozeRootView.swift lines 16-40 (including one that doesn't make sense to me - snooze for 0 seconds 🤷)

[Apple's Human Interface Guidelines related to this](https://developer.apple.com/design/human-interface-guidelines/notifications#Notification-actions) mentions 
> A notification can present a customizable detail view that contains up to four buttons people use to perform actions without opening your app.

I've not tested it, but sounds like there's a system limitation of 4 actions possible with notification actions. 

I've also realized after looking at this again that the Snooze module (specifically the "Click to Snooze Alerts" button that you mentioned) doesn't clear delivered/pending alerts as-is. I'll figure that out and push an update to this PR today or tomorrow to accommodate for that as well.

I'm certainly open to suggestions and feedback if you or anyone else thinks it should be different in someway. Just let me know! 😄 

edit - I realized I didn't directly answer your question. These notification actions to snooze alerts play directly into the existing Snooze module system; it's just another interface by which to interact with the app and snooze alarms - just as if the user opened the app and long pressed on the glucose bobble and choose snooze for 20 minutes or 40 minutes or 1 hour.

dnzxy:
> I've also realized after looking at this again that the Snooze module (specifically the "Click to Snooze Alerts" button that you mentioned) doesn't clear delivered/pending alerts as-is. I'll figure that out and push an update to this PR today or tomorrow to accommodate for that as well.

That would be great, thank you for looking into it!

> I'm certainly open to suggestions and feedback if you or anyone else thinks it should be different in someway. Just let me know! 😄

Since we get up to four buttons, would it maybe make sense to just give longer snooze intervals? E.g.
- 20min
- 60min
- 3 hours
- 6 hours

This would have to go hand-in-hand with the existing snooze module, so we should clean up and remove the other selection options, I reckon? What do you think?

I'll convert this PR to a draft until this gets to a ready state again 😊 thanks for looking into it @cachrisman

cachrisman:
Hi @dnzxy, I've reviewed and updated this PR to reflect those 4 options for both the notification actions as well as the Snooze module. I've built this based on the latest dev branch and am running it on my phone and everything is working as expected. Should be good to go. Let me know if there are any questions or issues.

Trio Watch App Extension/Helper/WatchNotificationHandler.swift
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    private func sendSnoozeRequest(for action: NotificationResponseAction) {
+        guard WCSession.isSupported() else { return }
+
+        let payload: [String: Any] = [WatchMessageKeys.snoozeDuration: action.minutes]
+        let session = WCSession.default
+
+        if session.delegate == nil {
+            session.delegate = PassiveSessionDelegate.shared
```
dnzxy: I'm not sure if this may cause problems. My understanding is that our `BaseWatchManager` owns the `WCSession` delegate. Now if this watch extension runs first (or later), it may steal the delegate or prevent expected delegate callbacks elsewhere

It might be safer to rely on the existing watch connectivity manager to activate the session at Trio startup via `transferUserInfo`. So how about doing something like `WCSession.default.transferUserInfo(payload)` here?

Trio/Sources/Services/UserNotifications/UserNotificationsManager.swift
```diff
@@ -331,6 +391,18 @@ final class BaseUserNotificationsManager: NSObject, UserNotificationsManager, In
...
+    private func alertToken(from glucose: GlucoseStored?) -> String {
+        if let id = glucose?.id?.uuidString {
+            return id
+        }
+
+        if let date = glucose?.date {
+            return "date-\(date.timeIntervalSince1970)"
+        }
+
+        return UUID().uuidString
```
dnzxy: If `id` and `date` are missing here, this generates/returns a new token every time, i.e. the actual dedupe will never work for those objects and we might end up unintentionally re-alerting. Should we return some kind of `unknown` state here and harden its handling for all callers?

Trio Watch App Extension/Helper/WatchNotificationHandler.swift
```diff
@@ -0,0 +1,103 @@
...
+                options: []
+            )
+
+            var categories = existingCategories
+            categories.update(with: glucoseCategory)
+            center.setNotificationCategories(categories)
+        }
+    }
+
+    private func title(for action: NotificationResponseAction) -> String {
```
dnzxy: This logic is also present, identical in `BaseUserNotificationsManager`. Same titles, same action list, same category. Let's either put this in some shared helper (e.g. a `NotificationCategoryFactory`) or at least move the localized titles into NotificationResponseAction via a `var localizedTitle: String { ... }`.

dnzxy (commented):
Some comments and questions. Also, a bit hard to currently grasp if everything properly runs on MainThread &/or is properly thread-safe, i.e., when called off-main like it happens for UNUserNotificationCenterDelegate, hopping to main with a `Task { @MainActor in ... }`. 

Overall great addition, good fixes.

Trio/Sources/Services/UserNotifications/UserNotificationsManager.swift
cachrisman: Agreed. Returning a fresh UUID when both id and date are missing guarantees the token never matches and can cause repeated alerts. I’m replacing that with a stable token: prefer glucose.id, else a minute-bucketed timestamp from glucose.date, else a stable Core Data objectID URI (when available) and, if identifiers are missing, returning "unknown" as the stable fallback. This keeps dedupe deterministic and avoids notification spam in the “missing fields” case.

```
    @MainActor private func sendGlucoseNotification() async {
...
                let token = alertToken(from: glucoseObjects.first)

                if token == "unknown" {
                    warning(.service, "Missing glucose token fields; skipping notification to avoid re-alerting")
                    return
                }
...
    }

    private func alertToken(from glucose: GlucoseStored?) -> String {
        if let id = glucose?.id?.uuidString { return id }

        if let date = glucose?.date {
            let roundedMinute = Int((date.timeIntervalSince1970 / 60).rounded())
            return "date-\(roundedMinute)"
        }

        // Stable fallback for Core Data objects:
        if let glucose, !glucose.objectID.isTemporaryID {
            return "objectID-\(glucose.objectID.uriRepresentation().absoluteString)"
        }

        // Stable “unknown” fallback: prevents repeated alarms when identifiers are missing
        return "unknown"
    }
```

one potential undesired side effect is that by returning 'unknown', it prevents spam, but it also means a real alarm might be missed if the reading lacks id/date and has a temporary objectID. I don't know how likely that is to happen, but I suspect it would probably be quite rare.

Trio Watch App Extension/Helper/WatchNotificationHandler.swift
cachrisman: Agreed — having parallel action/title/category logic is drift-prone. I’ve centralized these into shared definitions: action identifiers + localized titles live on NotificationResponseAction, and both iOS + watch register the glucose category via a shared NotificationCategoryFactory. This removes duplication and ensures any future changes to actions/titles/categories happen in one place and apply to both targets.

both iOS and watch use this to register the category
```
let glucoseCategory = NotificationCategoryFactory.createGlucoseCategory()
```

and here's the factory in NotificationIdentifiers.swift:

```
enum NotificationCategoryFactory {
    static func createGlucoseCategory() -> UNNotificationCategory {
        let snoozeActions = NotificationResponseAction.allCases.map { action in
            UNNotificationAction(
                identifier: action.rawValue,
                title: action.localizedTitle,
                options: []
            )
        }

        return UNNotificationCategory(
            identifier: NotificationCategoryIdentifier.trioAlert.rawValue,
            actions: snoozeActions,
            intentIdentifiers: [],
            options: []
        )
    }
}
```

I've also updated the NotificationIdentifiers to use a localizedTitle

Trio Watch App Extension/Helper/WatchNotificationHandler.swift
cachrisman: You’re right. WatchNotificationHandler shouldn’t ever set WCSession.default.delegate (or try to “fill in” a missing delegate) since BaseWatchManager is the intended owner. I’m updating the handler to only enqueue the snooze request payload and removing the PassiveSessionDelegate fallback entirely. This keeps delegate/activation lifecycle owned by the watch connectivity manager and prevents ordering-dependent behavior.

```
    /// Sends snooze request to iPhone via WatchConnectivity.
    /// WCSession.transferUserInfo is thread-safe and can be called from any thread.
    private func sendSnoozeRequest(for action: NotificationResponseAction) {
        guard WCSession.isSupported() else { return }

        let payload: [String: Any] = [WatchMessageKeys.snoozeDuration: action.minutes]
        WCSession.default.transferUserInfo(payload)
    }
```

cachrisman:
Thanks for the thoughtful review.

On the thread-safety / MainThread concern: agreed that it’s not obvious at a glance, especially since UNUserNotificationCenterDelegate callbacks can arrive off-main. In the follow-up changes I’m making the threading model explicit by ensuring all state mutations and side-effectful paths that touch UI-adjacent/shared state (e.g., updating snooze state / last-alert tokens / scheduling notifications) funnel through @MainActor entrypoints. Where work can run off-main (building payloads, computing tokens, etc.) it stays off-main, but the “commit” step is always on MainActor.

I’ll also add brief comments (and/or @MainActor annotations) around the delegate-handling methods to document the intended thread, so future changes don’t accidentally bypass the main-actor hop.

Appreciate the callout — it’s a good point and I’ll tighten up the annotations for clarity and safety.

cachrisman:
just leaving a comment here to say that I found a few bugs to fix/issues to improve recently that my latest commit addresses:

- **Threading standardization**: Unified snooze-related async work to use `Task { @MainActor [weak self] }` instead of mixing `DispatchQueue.main.async` and `Task` patterns for consistency.
- **Watch connectivity fix**: The watch was sending snooze via `transferUserInfo`, but the phone was only listening in `didReceiveMessage`. Updated to:
  - Use `sendMessage` when the session is reachable and activated (immediate delivery)
  - Fall back to `transferUserInfo` when not reachable or if `sendMessage` fails (queued delivery)
  - Handle snooze in both `didReceiveMessage` and `didReceiveUserInfo` on the phone side

Trio/Sources/Modules/Snooze/SnoozeStateModel.swift
```diff
@@ -4,12 +4,53 @@ import SwiftUI
...
+            // Allow any duration chosen in the Snooze UI, while keeping validation for quick actions elsewhere.
+            snoozeUntilDate = duration > 0 ? Date().addingTimeInterval(duration) : .distantPast
+            alarm = glucoseStorage.alarm
+            await notificationsManager.applySnooze(for: duration)
+        }
+    }
+}
+
+extension Snooze.StateModel: SnoozeObserver {
+    func snoozeDidChange(_ untilDate: Date) {
```
dnzxy, line 40: This is being called via the `SnoozeObserver` decorated with `@MainActor` on the main thread/process, and then this function's signature is sync and starts an async `Task`, also on `@MainActor`. It's my understanding this is redundant now and the `Task` here could be omitted, could it not?

Trio/Sources/Services/UserNotifications/UserNotificationsManager.swift
```diff
@@ -409,6 +470,19 @@ final class BaseUserNotificationsManager: NSObject, UserNotificationsManager, In
         }
     }
 
+    @MainActor func applySnooze(for duration: TimeInterval) async {
+        let untilDate = Date().addingTimeInterval(duration)
```
dnzxy: If I understand your implementation correctly, in Snooze.StateModel.applySnooze you treat `duration` > 0 as a snooze, else `.distantPast`, but here it always gets applied.

How about doing
```swift
let untilDate = duration > 0 ? Date().addingTimeInterval(duration) : .distantPast
``` 
here to make it more explicit?

Nit comment though.

dnzxy (approved):
Two questions / comments.
Overall looks great and would be fine to go along for me.

I'd ask @marv-out and @kingst to please also give this a proper review.

j-ktz:
A bell/notification icon on the Home Screen UI to signify if notifications are muted or not. Currently, when you are high, you can open the alert and snooze. But once you do that there is no way to unsnooze until the time frame has elapsed. If there was an icon you could quick toggle between mute and unmute.

Trio/Sources/Modules/Snooze/View/SnoozeRootView.swift
```diff
@@ -87,10 +70,14 @@ extension Snooze {
                     let interval = pickerTimes[selectedInterval]
                     let snoozeFor = formatter.string(from: interval)!
```
kingst: I realize that you didn't add this, but can you remove this forced unwrap? We can't have these in production code.

Trio/Sources/Modules/Snooze/SnoozeStateModel.swift
```diff
@@ -4,12 +4,53 @@ import SwiftUI
...
+        // Add validation helper inside the class
+        private func validateSnoozeDuration(_ duration: TimeInterval) -> Bool {
+            // Only allow durations matching our defined actions
+            NotificationResponseAction.allCases
+                .map(\.duration)
+                .contains(duration)
+        }
+
+        // Add handleSnoozeResponse inside the class
+        func handleSnoozeResponse(_ duration: TimeInterval) {
```
kingst: Is this called anywhere? If not, consider removing it

kingst (commented):
Overall this looks like a solid improvement, deduping notifications and adding actionable snooze buttons is a real UX win. I didn't test it, my review is just from reading the code and I'm not that familiar with this code.

Trio/Sources/Modules/Snooze/View/SnoozeRootView.swift
cachrisman: I fixed this and another forced unwrap earlier in the file.

Trio/Sources/Modules/Snooze/SnoozeStateModel.swift
cachrisman: You are correct. I've removed it.

Trio/Sources/Services/UserNotifications/UserNotificationsManager.swift
cachrisman: Good catch — aligned with the StateModel logic for consistency.

Trio/Sources/Modules/Snooze/SnoozeStateModel.swift
cachrisman: you are right about that. I've updated it to take out the unnecessary Task.

dnzxy (dismissed):

dnzxy:
Hi @cachrisman, are you on Discord by any chance?
The PR does not compile for me (or our unit tests) on my end.

dnzxy (approved):

PR #813, opened 2025-10-16
Fix typos in Onboarding

### LINKED ISSUE #747: Incorrect ISF explanation in therapy settings (mmol/L per unit)
### Trio Version:

* Version Number: *0.5-dev*
* Repo: *nightscout/Trio*
* Git Reference: *(installed via Xcode manually – no specific commit hash available)*

Hi,
I noticed a possible mistake in the explanation of the Insulin Sensitivity Factor (ISF) within the therapy settings of the app (Trio/Loop). The current text says:

“A higher value means you’re less sensitive to insulin.”

However, when using mmol/L per iE, a higher ISF value means that one unit of insulin lowers blood glucose more strongly – in other words, you are actually more sensitive to insulin.

For example:
– ISF = 3.6 → 1 iE lowers BG by 3.6 mmol/L (high sensitivity)
– ISF = 2.6 → 1 iE lowers BG by 2.6 mmol/L (lower sensitivity)

So the statement in the app appears to be inverted. It might confuse users, especially in international settings using mmol/L.

Thanks for your amazing work – just wanted to help improve clarity!

Image

MikePlante1:
This fixes the mixup between `less` and `more` sensitive in the Onboarding's description for ISF, as well as some formatting errors. I copy/pasted the localizations for the formatting errors, but let the ones reset where the original string changed meanings.

| 0.6.0.2 | This PR | localization reset |
|---|---|---|
| "• A higher number means you're less sensitive to insulin" | "• A higher number means you're more sensitive (less resistant) to insulin" | yes |
| "• A lower number means you're more sensitive to insulin" | "• A lower number means you're less sensitive (more resistant) to insulin" | yes |
| "• A ratio of 10 g/U means 1 unit of insulin covers 10g of carbs" | "• A ratio of 10 g/U means 1 unit of insulin covers 10 g of carbs" | no |
| "• An ISF of %@ means 1 U lowers your glucose by %@" (50 mg/dL, 50 mg/dL) | "• An ISF of %@ %@/U means 1 U lowers your glucose by %@ %@" (50, mg/dL, 50, mg/dL) | no |
| "For 45g of carbs, you would need:" | "For 45 g of carbs, you would need:" | no |
| "Welcome to Trio - an automated insulin delivery system for iOS based on the OpenAPS algorithm with adaptations." | "Welcome to Trio — an automated insulin delivery system for iOS based on the OpenAPS algorithm with adaptations." | no |

### DIFF
--- a/Trio/Sources/Modules/Onboarding/View/OnboardingSteps/TherapySettings/CarbRatioStepView.swift
+++ b/Trio/Sources/Modules/Onboarding/View/OnboardingSteps/TherapySettings/CarbRatioStepView.swift
@@ -69,7 +69,7 @@ struct CarbRatioStepView: View {
                             .padding(.horizontal)
 
                         VStack(alignment: .leading, spacing: 8) {
-                            Text("For 45g of carbs, you would need:")
+                            Text("For 45 g of carbs, you would need:")
                                 .font(.subheadline)
                                 .padding(.horizontal)
 
@@ -79,7 +79,7 @@ struct CarbRatioStepView: View {
                                         .carbRatioRateValues[state.carbRatioItems.first!.rateIndex] as NSNumber
                                 )
                             Text(
-                                "45 \(String(localized: "g", comment: "Gram abbreviation")) / \(formatter.string(from: state.carbRatioRateValues[state.carbRatioItems.first!.rateIndex] as NSNumber) ?? "--")  = \(String(format: "%.1f", insulinNeeded))" +
+                                "45 \(String(localized: "g", comment: "Gram abbreviation")) / \(formatter.string(from: state.carbRatioRateValues[state.carbRatioItems.first!.rateIndex] as NSNumber) ?? "--") \(String(localized: "g/U")) = \(String(format: "%.1f", insulinNeeded))" +
                                     " " + String(localized: "U", comment: "Insulin unit abbreviation")
                             )
                             .font(.system(.body, design: .monospaced))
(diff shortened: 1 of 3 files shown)


MikePlante1:
@t1dude Is manually fixing localization like this in Xcode okay? Or does that mess anything up with crowdin?

t1dude:
> @t1dude Is manually fixing localization like this in Xcode okay? Or does that mess anything up with crowdin?

English is the main language for Trio and all source strings are written in English directly in code. By the time we got to discussing this, it was too late to move English out of code and into the string catalog (@dnzxy). Because of this, we should not fix English spelling mistakes or do improvements to the English language by using the string catalog. We need to do the improvements in code, where the string is located.

These are the steps needed:

1. Make the language improvement directly in the code.
2. Build the app in Xcode, this will update the string catalog, updating all the source strings that have been changed.
3. Make a PR to update the repository. All files from steps 1 and 2 will be updated. The English language will be updated immediately, because it does not need translation.
4. Upload the new string catalog to Crowdin so that translators get a notification with the new source strings, and they will update all translations.
5. Pull an updated string catalog from Crowdin after some time, and make a PR to have the updated string catalog added to the repository.

I am currently doing steps 3 and 4 every two months or whenever someone asks me to :)

MikePlante1:
@t1dude I did steps 1-3 for the ones that say "yes" for "localization reset"

But the ones that are just formatting updates I also did the English translations in code, which made all the translations for those strings stale, so I went into the string catalog and copy/pasted the translations available from the stale string into the new string. Is that okay?

t1dude:
> @t1dude I did steps 1-3 for the ones that say "yes" for "localization reset"
> 
> But the ones that are just formatting updates I also did the English translations in code, which made all the translations for those strings stale, so I went into the string catalog and copy/pasted the translations available from the stale string into the new string. Is that okay?

This might work, but it might also break some Crowdin translations. Whitespaces etc f things up. My experience after 20+ rounds of this, is that we should not edit the string catalog manually, apart from removing stale strings. These are strings that exist in the catalog but have been removed from code. But hey, worst case is that some strings show up for translations again in Crowdin.

t1dude (approved):

PR #819, opened 2025-10-20
Refactor Therapy Settings Editors 

marv-out:
# Refactor Therapy Settings Editors to Use Shared Component

## Summary
This PR refactors all four therapy settings editors (Glucose Targets, Basal Rates, Carb Ratios, and Insulin Sensitivities) to use the unified `TherapySettingEditorView` component, bringing consistency with the onboarding flow.

## Changes Made

1. Fixed some problems in onboarding where the user can't really add multiple entries and edit them
2. Refactored All Therapy Settings Editors to match the onboarding UI using the `TherapySettingEditorView` component


Sjoerd-Bo3:
Already great rework, I think it's much clearer already!

Few things I encountered while testing in settings:
- It does allow you to tap the 'add' button even when there is a 23:30 entry, which results in not an extra row being added.

<video width="280" height="800" src="https://github.com/user-attachments/assets/29e5991c-4dd3-40ca-947e-aa7aaa6ea8b0.mp4" controls preload></video>

- When changing an entry, and sliding one of the values to another entry. It auto closes pretty abruptly. Also I have no chance to change the other item. Is this intended? I think it would be more logical to have it open until you close it by clicking the row?

<video width="280" height="800" src="https://github.com/user-attachments/assets/10c3ba50-16a4-4ed1-a780-dc0679139078.mp4" controls preload></video>

dnzxy:
Thanks for taking on the port of my Onboarding work and refactoring the Therapy Settings, plus fixing iOS 26 SDK bugs.

As for the UI: this is the expected look of the chart and content (Screenshot taken from Onboarding).

image

- The text and double headlines are unnecessary.
- Users have already gone through Onboarding, they know their settings and what each setting does. 

There is a change to `Info.plist` that seems unnecessary.

MikePlante1:
> There is a change to `Info.plist` that seems unnecessary.

I think that’s one of those things that if you build with Xcode 26 then it makes that change when you build. But then if you build with Xcode 16.4 it reverts that change.

marv-out:
>  The text and double headlines are unnecessary.

I just wanted to recreate the onboarding UI in the first step and get some feedback on it. The double headlines aren’t really necessary, but I wanted to include that emoji in the navigation title — which isn’t quite possible without a custom nav title. As you know, I’m a fan of those emojis (even though I know you’re not xD).

Anyway, which texts should stay, and which ones can go? I actually like the explanatory texts in the CR Editor view and the ISF view.

> There is a change to Info.plist that seems unnecessary.

As Mike said I haven't changed anything there. Must have been Xcode 26

marv-out:
> When changing an entry, and sliding one of the values to another entry. It auto closes pretty abruptly. Also I have no chance to change the other item. Is this intended? I think it would be more logical to have it open until you close it by clicking the row?

No, thats not intended. I will have a look

> It does allow you to tap the 'add' button even when there is a 23:30 entry, which results in not an extra row being added.

Ok, we can disable the button in that case

dnzxy:
> The double headlines aren’t really necessary, but I wanted to include that emoji in the navigation title

I’d suggest to keep it default SwiftUI, and remove the double title, so the one with the icon.

> which texts should stay, and which ones can go? 

Imho all of them. The goal would be to replace the current setup (Add button top right corner, chart, then rates) where you need to tap on a table item, it opens a NavView, you need to change time and rate etc. and make it into the inline-editing UI we have in Onboarding. 

Whether we keep the Add button above the list (and sticky) have it be a toolbar item could be discussed.

> As Mike said I haven't changed anything there. Must have been Xcode 26

Thanks for clarifying that to you and Mike. My bad. 

The current (before PR) therapy setting views also have a small warning when the schedule becomes "full". This may be something to keep, possibly. Open for discussion. 

image

The main thing we need to fix and we fixed that with what was built for Onboarding, is the whole back and forth navigation stuff. That a) bad UX and b) buggy for iOS26, as it doesn’t keep changed entry values.

Sjoerd-Bo3:
I like the message about full rates.

marv-out:
> Imho all of them. 

just to clarify, you are talking about removing (texts) here, aren't you?

> Whether we keep the Add button above the list (and sticky) have it be a toolbar item could be discussed.

when we change it here we also need to change it in onboarding. I like it as it is right now, but I needed to change it with this PR to move it outside the list

> The current (before PR) therapy setting views also have a small warning when the schedule becomes "full".

I like that, will add it back

dnzxy:
> > Imho all of them.
> 
> just to clarify, you are talking about removing (texts) here, aren't you?
> 
Yes, thats why I quoted you.

> > Whether we keep the Add button above the list (and sticky) have it be a toolbar item could be discussed.
> 
> when we change it here we also need to change it in onboarding. 

Not really, as the UI in Onboarding is different and you cannot fit a toolbar button there. I was specifically talking about this PR and therapy settings.

marv-out:
> Yes, thats why I quoted you.

Yeah, but I asked which texts could be deleted AND also which texts should stay and you simply said `all` 😂 nvm

> Not really, as the UI in Onboarding is different and you cannot fit a toolbar button there. I was specifically talking about this PR and therapy settings.

Well...doesn't matter that much to me where this button is. The most important is that its no longer in the List and doesn't scroll away when there are too many items in the List. One argument in favor of the toolbar would be that its more Swift-like

dnzxy:
> Yeah, but I asked which texts could be deleted AND also which texts should stay and you simply said `all` 😂 nvm

My suggestion was to: Remove all texts. Remove one of the duplicate titles.

> One argument in favor of the toolbar would be that its more Swift-like

Agree absolutely, but we don't have that "screen real estate" in Onboarding. Not really relevant for this PR / your work here though.

dnzxy:
Side note: There are merge conflicts as latest translations were merged to dev today.

MikePlante1:
> Side note: There are merge conflicts as latest translations were merged to dev today.

@marv-out Are you okay with me doing and pushing this merge to make sure some of the onboarding localization changes don't get reverted?

MikePlante1:
My language was imprecise. To clarify — I was tested merging 0.6.0.5 in with Xcode 16.4 vs Xcode 26 because development should be done in Xcode 16.4 given that's what Trio's fastlane is currently using... and in doing so I discovered most of #813 got reverted when it was merged earlier in this PR. So I'd like to correct that:

https://github.com/marv-out/Trio/compare/therapy...MikePlante1:819?expand=1

marv-out:
> Are you okay with me doing and pushing this merge to make sure some of the onboarding localization changes don't get reverted?

Yes xD

dnzxy:
Have the two "issues" in [this comment](https://github.com/nightscout/Trio/pull/819#issuecomment-3423141455) been addressed yet? I'm not sure, I don't see any commits. @marv-out ? 👀

marv-out:
Ok please retest/ review again @Sjoerd-Bo3 @dnzxy @MikePlante1 

I've removed the texts, re-added the basal coverage box, fixed the picker behaviour and disabled the add button when the last entries' time was at 11:30 p.m.

should be good to go imo (unless you wanna move the add button to the toolbar in trailing position, haven't done that yet)

MikePlante1:
Not sure why part of #813 got reverted again... but I'll push a fix for that shortly, along with some Xcode 16.4 fix-its, including the linting that was introduced with #705 (0.6.0.7).

Missing space between value and units and ISF's units should be mg/dL/U:



EDIT: I figured it out. I missed this part in my previous revert-revert-813 commit. Oops.

marv-out:
Whats the status of this PR? Is it good to go?

dnzxy:
Will try and get a thorough review in tonight (pacific time, so in roughly 8-10 hrs).

dnzxy:
I've cleaned up the UI a bit and pushed a few adjustments. 
Looks good to me so far :)

Trio/Sources/Modules/ISFEditor/View/ISFEditorRootView.swift
```diff
@@ -67,143 +69,100 @@ extension ISFEditor {
...
-                    Picker(selection: $state.items[index].timeIndex, label: Text("Time")) {
-                        ForEach(0 ..< state.timeValues.count, id: \.self) { i in
-                            Text(
-                                self.dateFormatter
-                                    .string(from: Date(
-                                        timeIntervalSince1970: state
-                                            .timeValues[i]
-                                    ))
-                            ).tag(i)
+                                    Text("Swipe to delete a single entry. Tap on it, to edit its time or rate.")
```
MikePlante1: > `time or rate.`

*`time or value.`

MikePlante1:
I find it strange how Basal Rates all get set to 0.05 U/hr when a pump is not connected. Maybe this isn't something we can easily fix, though, since unlike other therapy settings, editing basal rates requires a pump to be connected.

---

Screenshot 2025-11-14 at 06 02 03

This appears in Therapy settings in the app, but not in Onboarding.

---

Screenshot 2025-11-14 at 05 18 20

I went through onboarding in mg/dL, set the target to 180 mg/dL, went back switched to mmol/L, and Trio crashes out.

I reopened the app, which reset onboarding, started with mmol/L this time, set a target of 10.0 mmol/L, went back to switch it to mg/dL, then went forward in onboarding and it used 132 mg/dL instead of 180 mg/dL. Also tried 8.0 mmol/L, which it converted to 112 mg/dL instead of 144 mg/dL.

Same thing with ISF too.

This wasn't introduced with this PR, though, as I checked `dev` 0.6.0.9 and got the same results.

---

Another thing not introduced in this PR, but I find it odd that setting a Scheduled Basal Rate > 2 U/hr doesn't either automatically increase `Maximum Basal Rate` to the highest rate or just give an error and force the user to edit it themselves.

---

Will resuming testing after I board my flight...

dnzxy:
> I find it strange how Basal Rates all get set to 0.05 U/hr when a pump is not connected. Maybe this isn't something we can easily fix, though, since unlike other therapy settings, editing basal rates requires a pump to be connected.

Good find, good testing haha. That’s why for the Onboarding we have introduced the pump selection step. The pump you choose there is just a mock pump that is used to get an increment value for the subsequent settings screens. We don’t really have a good way of doing this for basal once onboarded when the user has no onboarded (i.e. instantiated PumpManager via one of the submodules) pump. It’s annoying but not sure if it’s important. 

We _could_ persist the pump selection the user chose during Onboarding but what happens if the user then uses a pod and then switches their pump. 

When the user has a pump onboarded, we set the increment correctly. It’s a bit of a shitty situation, tbh. 

I‘ll fix the typo and would open new issues for the other stuff, plus wait for whatever you may find later :)

MikePlante1:
Nothing else came up in my testing, so once that typo is fixed, I think this is good to merge.

Trio/Sources/Modules/ISFEditor/View/ISFEditorRootView.swift
dnzxy: Addressed. 

I actually cross compared and in the current implementation (as of latest `dev`), it says `rate` for all the settings, so targets, basal, carb ratio and isf. It now says value everywhere. It only seemed to say `rate` for ISF, as I had already adjusted it to say `value` everywhere else.

MikePlante1 (approved):

PR #821, opened 2025-10-22
Fix iOS 26.1 LoopStatusView Detent Issue

### LINKED ISSUE #840: Fuzzy screen and displaced loop status popup
I'm in my first sensor warm-up period since installing v0.6.0.7 and iOS26.1, so I'm not sure which might be causing this.

- Start without a CGM signal.  (It's OK to be in warmup.)
- When I tap on the greyed out loop status pill in the HUD, the screen goes fuzzy, and it looks like the loop status popup is being displaced to the bottom of the screen.  
- It's challenging to grab the handle, and it won't 'pull up' all the way.  
- Tapping in the main screen dismisses the fuzzy background and the displaced popup.

The pod info popup appears correctly.  This behavior has persisted despite app and phone reboot.  I'll update this issue with findings after my warmup period has completed.

I'm using an iPhone 15 Plus. 

[Update: yes, this problem persists after the CGM has completed its warmup period.]

https://github.com/user-attachments/assets/b4e61fa0-b533-44e0-a3f0-6fc417bac5da

MikePlante1:
There was a change (or bug?) in the iOS 26.1 Beta which broke the LoopStatusView from expanding to be viewable. Xcode says "Invalid sheet detent height: 0.0". 

The hardcoded 90% does not look as good as when the View is working properly (like on the latest public release of iOS — 26.0.1) since it hides more of the screen and has an awkward gap below the Got It button, but since this PR just forces a 90% detent if the detent height is 0, users that aren't experience this problem won't be affected. 

Once iOS 26.1 is officially released, if it's still falling back to the 90% version we should look into a more elegant solution.

| dev 0.6.0.4 | This PR |
|---|---|
| image | 41C2C1F5-EBF2-4DE6-A7AD-C253ED7C7A75_1_101_o |

### DIFF
--- a/Trio/Sources/Modules/Home/View/Header/LoopStatusView.swift
+++ b/Trio/Sources/Modules/Home/View/Header/LoopStatusView.swift
@@ -129,7 +129,11 @@ struct LoopStatusView: View {
         .onAppear {
             lastDetermination = state.determinationsFromPersistence.first
         }
-        .presentationDetents([.height(sheetContentHeight)])
+        .presentationDetents([
+            sheetContentHeight > 0
+                ? .height(sheetContentHeight)
+                : .fraction(0.9)
+        ])
         .presentationDragIndicator(.visible)
         .onPreferenceChange(ContentSizeKey.self) { newSize in
             sheetContentHeight = newSize.height
--- a/Trio/Sources/Modules/Home/View/Header/LoopStatusView.swift
+++ b/Trio/Sources/Modules/Home/View/Header/LoopStatusView.swift
@@ -130,9 +130,8 @@ struct LoopStatusView: View {
             lastDetermination = state.determinationsFromPersistence.first
         }
         .presentationDetents([
-            sheetContentHeight > 0
-                ? .height(sheetContentHeight)
-                : .fraction(0.9)
+            sheetContentHeight > 0 ? .height(sheetContentHeight) : .fraction(0.9),
+            .large
         ])
         .presentationDragIndicator(.visible)
         .onPreferenceChange(ContentSizeKey.self) { newSize in


dnzxy:
I don’t think we need to provide any fixes or support, or introduce code changes for users running unreleased beta-type iOS versions. If this persists to be an issue in an upcoming official release of iOS, this becomes a different thing. Until then, I‘d not do any changes.

t1dude (commented):
Let's point affected users to this PR, but I agree that we should wait and see if this issue still exists when 26.1 is released.

MikePlante1:
I just installed the iOS 26.1 Release Candidate (23B82) and this is still an issue, so looks like we'll need that more elegant solution eventually, and this temporary patch-job if another solution is not found before 26.1 is officially released.

MikePlante1:
Just updated to the full release of 26.1 and confirmed this bug is still present with the latest dev 0.6.0.7

MikePlante1:
Updated it a bit. Now,

* If it can set a calculated size it'll set it to that (< iOS 26.1)
* If it can't, it sets 0.9 which seems to be fine for most cases, but might cut off the button for SEs in certain cases
* Either way, you can now swipe to make it full screen.


https://github.com/user-attachments/assets/a07fc7ac-90b0-4b36-9aea-e01cf75551d5

dnzxy (approved):
I think this about as good as it will get for now. 

I have been trying various things to enforce a rerendering of the sheet once content detent height can properly be calculated by the `GeometricReader` but it seems like it’s not doing what it should on iOS 18.x and iOS 26 (26.0.1 nor 26.1). 

Let’s get this in for now :)

pepijnjoost:
Is working just like @MikePlante1 posted!

g5WS:
Big heaps of thanks, @MikePlante1 !

🙏

PR #598, opened 2025-05-27
fix: Live Activity color calculation for MMOL units with static color scheme

Sjoerd-Bo3:
## Summary
- Fixed incorrect color display in Live Activities for MMOL users with static color scheme
- Fixes #537

## Problem / Solution
**Problem**: When using MMOL units with a static color scheme, the Live Activity was comparing MMOL glucose values against mg/dL thresholds, resulting in incorrect colors being displayed.

**Solution**: Modified `LiveActivityView.swift` to properly convert `highGlucose` and `lowGlucose` thresholds from mg/dL to mmol/L when the user has selected MMOL units, ensuring correct color calculation for the static color scheme.

## Test plan
- [x] Test with MMOL units and static color scheme enabled
- [x] Verify colors display correctly for high/low glucose values
- [x] Confirm no regression for mg/dL users
- [ ] Test across different glucose ranges
- [ ] Verify with dynamic color scheme still works

## Technical Details
The fix ensures that when using static color scheme with MMOL units, the threshold values are converted using the same unit conversion logic as the glucose values themselves.

PR #828, opened 2025-10-29
feat: update watch os Live Activity

### LINKED ISSUE #537: Live activities wrong colour.
BG value in live activities is the wrong colour. 

Version 0503

This apparently is an MMOL conversation issue

Apologies for omitting - Static colouring

Image

### LINKED ISSUE #772: Ios26 CarPlay LA
The LA on CarPlay 26 is rather bland. Whilst it has the basic BG and trend, COB and IOB would be very useful. Doesn’t need the graph like the phone LA as that would be too much info whilst driving, but COB and IOB, I think, would be essential.

bastiaanv:
With iOS 18, Apple released a new feature: [supplementalActivityFamilies](https://developer.apple.com/documentation/SwiftUI/WidgetConfiguration/supplementalActivityFamilies(_:)). This allows us to render a different LA based on wheter the LA is running on your iPhone or Watch/CarPlay.

<details><summary><b>Before:</b></summary>
<p>
Example provided by Marion, Trio used the same layout and data, but the title would say Trio instead of Loop:

incoming-5D0598E8-F68B-406E-8D25-DFB5173AD193

</p>
</details> 

<details><summary><b>After:</b></summary>
<p>

### WatchOS

**Simple view:**
Screenshot 2025-10-29 at 11 22 12

**Detailed view:**
Screenshot 2025-10-29 at 11 23 38

### Carplay
**Simple view:**
507165721-7eaa738c-c3e3-49f9-b3f2-4fd53f8083a3

**Detailed view:**
507165677-5a92c3a1-ab73-469b-ad11-cd5c15e0cffd

</p>
</details> 

<details><summary><b>LA settings:</b></summary>
<p>

IMG_0999

</p>
</details> 

TODO's:
- [x] Localizations in LiveActivitySettingsRootView
- [x] Try to detect is Widget is in a CarPlay environment (NOT POSSIBLE)
- [x] Process feedback

closes #772, #598

### DIFF
--- a/LiveActivity/LiveActivity+Helper.swift
+++ b/LiveActivity/LiveActivity+Helper.swift
@@ -232,10 +232,9 @@ struct LiveActivityModifiers: ViewModifier {
 
     func body(content: Content) -> some View {
         content
-            .padding(.all, isWatchOS ? 8 : 14)
+            .padding(.all, isWatchOS ? 10 : 14)
             .frame(minHeight: 0, maxHeight: .infinity)
             .privacySensitive()
-            .foregroundStyle(Color.primary)
             // Semantic BackgroundStyle and Color values work here. They adapt to the given interface style (light mode, dark
             // mode)
             // Semantic UIColors do NOT (as of iOS 17.1.1). Like UIColor.systemBackgroundColor (it does not adapt to changes of
--- a/LiveActivity/LiveActivity+Helper.swift
+++ b/LiveActivity/LiveActivity+Helper.swift
@@ -179,6 +179,7 @@ func bgAndTrend(
             case .minimal:
                 let scaledText = text.scaleEffect(x: 0.7, y: 0.7, anchor: .leading)
                 scaledText.foregroundStyle(hasStaticColorScheme ? .primary : glucoseColor)
+
             case .compact:
                 text.scaleEffect(x: 0.8, y: 0.8, anchor: .leading).padding(.trailing, -3)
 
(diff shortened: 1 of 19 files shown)


dnzxy:
Cool stuff! I also was genuinely not aware (I actually looked at it, but didn't find info on it) that [supplementalActivityFamilies](https://developer.apple.com/documentation/SwiftUI/WidgetConfiguration/supplementalActivityFamilies(_:)) can also be used or rather affects the CarPlay LA widget.

I really don't want more options/settings, but would it make sense to change to glucose value's color in the detailed screen to reflect the dynamic color or does that look off?

You already have 3 different styles next to each other (bold and big, regular, secondary and bold), so it may obscure clarity. That'd be something that needs to be tried.

bastiaanv:
@dnzxy Something like this?
No bold, less colors, only primary & glucose color. Also decreased the padding between the arrow & current BG

Screenshot 2025-10-29 at 20 40 10

dnzxy:
Something like that, for example, yes. 

Just curious, if you make the glucose and arrow bold, leave the time regular and then use the same string in a `VStack` that we use for the lockscreen detailed one (time + "Updated" beneath) in smaller text and `.secondary`?

You don’t have to push it; just an idea haha 😅 Might look super off. No idea. 

But I‘d 👍 this. With or without color, up to you.

bastiaanv:
Screenshot 2025-10-30 at 21 30 37

@dnzxy Something like this? With this, the time & updated are `caption2` which is the smallest size Apple allows:
https://developer.apple.com/design/human-interface-guidelines/typography#Specifications

The bold feels better. I am not bothered by it, so we could consider to push that part

Screenshot 2025-10-30 at 21 33 39

This specific state is not good, due to the updated label

dnzxy:
@bastiaanv I like it. Maybe let’s get rid of the "updated" again but introduce some SF symbol that indicates it was a loop status? Like Green O and timestamp, in one line? 

Either way, it’s cool!

dnzxy:
I believe the addition of EversenseKit here was by mistake, right? 

If you remove that one I think this PR can go in.

bastiaanv:
Yes it was. Need a clean PR?
It was already removed fro the PR

dnzxy:
Nope all good, seems like I had been looking at a dated file diff

dnzxy (approved):

marv-out (approved):
LGTM

PR #842, opened 2025-11-06
chore: update Medtrum & sync dev

bastiaanv:
Full list of changes:
- MedtrumKit: https://github.com/jbr7rr/MedtrumKit/pull/48
- Sync Trio dev-branch

### DIFF
--- a/Config.xcconfig
+++ b/Config.xcconfig
@@ -19,7 +19,7 @@ TRIO_APP_GROUP_ID = group.org.nightscout.$(DEVELOPMENT_TEAM).trio.trio-app-group
 
 // The developers set the version numbers, please leave them alone
 APP_VERSION = 0.6.0
-APP_DEV_VERSION = 0.6.0.8
+APP_DEV_VERSION = 0.6.0.9
 APP_BUILD_NUMBER = 1
 COPYRIGHT_NOTICE =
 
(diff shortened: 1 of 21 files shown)


dnzxy:
I realize that https://github.com/jbr7rr/MedtrumKit/pull/48 which contains https://github.com/jbr7rr/MedtrumKit/issues/27 was already merged, but just to comment on this:

> feat: low reservoir notification https://github.com/jbr7rr/MedtrumKit/issues/27 (disabled by default)

The other PMs let you select an insulin amount when to warn "at". It is not disabled by default, it's enabled basically always and you can choose at what reservoir fill to get the warning. I'd suggest do adhere to this behavior and cross compare with Omnipod and the other PMs.

Edit to add: Unless I am misunderstanding this feature request, as from running MedtrumKit on my in-vitro device here, there is the low reservoir setting present.

bastiaanv:
@dnzxy

Normally the user will be notified for the following:
- The patch has x hours remaining (if the patch is in normal mode)
- The patch is in a faulty state: occlusion, fault, empty reservoir, etc

That PR adds the ability to also get notified earlier than 0U reservoir level. This PR adds the bottom one to the feature branch, maybe the text in the label needs to be finetuned.

Screenshot 2025-11-09 at 16 27 39

Sjoerd-Bo3 (approved):
LGTM also Builds and runs!

Sjoerd-Bo3:
@dnzxy @bastiaanv I’m merging, you can continue the discussion here or take it into an issue. It’s not related to something that was introduced in this PR but earlier. So this can still merge.

PR #846, opened 2025-11-15
Remove 400 / shouldProtectDueToHIGH guard

dnzxy:
### Summary

This PR removes the "400 guard" (via `TrioCustomOrefVariables` `shouldProtectDueToHIGH`) added in #622. That logic disabled SMBs and forced a neutral temp basal whenever the CGM reported flatlined HIGH values (400 mg/dL) [to avoid dosing on potentially invalid readings].

### Changes

* Remove `shouldProtectDueToHIGH` from `TrioCustomOrefVariables` (model + decoding + constructor).
* Remove `glucoseIsHIGH` helper and all references.
* Remove warning message from Loop Status View when glucose is 400 mg/dL, telling user dosing has been disabled.
* Delete 400-guard logic in `determine-basal.js` (no forced neutral TBR, no SMB block).
* Update webpack bundle and `oref0_source_version.txt` accordingly.

### Rationale

The original guard was only meant as a temporary safety measure for suspicious HIGH readings. It’s no longer needed and diverges from upstream oref, so we’re removing it to restore standard oref behavior, to aid FCL users in underserved populations that semi-frequently run into situations where they get stuck on 400 mg/dL for a prolonged time period.


oddst:
Really happy to see this change likely being done. The current behavior has been a problem for teenagers with Trio, when they get into troubles and their parents cannot contact them to remind them that Trio stops automatic corrections at 400 mg/dL . Of the two options, I believe this new one is the least worst choice - as being really stuck high happens more often than what the previous behavior protection...! Thanks...

oref0_source_version.txt
```diff
@@ -1,6 +1,7 @@
-oref0 branch: dev - git version: 37896e5
```
MikePlante1, line 1: ~~Looks like this could be updated to dev (83fd5de) given that https://github.com/nightscout/Trio/pull/622 was merged.~~

dnzxy: Uhm, this seems like the exact opposite of what we want to do here.
This PR removes the code that was merged via #622. 
Unless we merge https://github.com/nightscout/trio-oref/pull/49 I cannot update to `trio-oref:dev`.

MikePlante1: My mistake, I see I got my wires twisted. Will look at this again with fresh eyes tomorrow.

dnzxy:
> Really happy to see this change likely being done. The current behavior has been a problem for teenagers with Trio, when they get into troubles and their parents cannot contact them to remind them that Trio stops automatic corrections at 400 mg/dL . Of the two options, I believe this new one is the least worst choice - as being really stuck high happens more often than what the previous behavior protection...! Thanks...

Hi @oddst, thanks for the feedback and for sharing some user experience.

To clarify the intent: the guard introduced in #622 was a Trio-specific protection to avoid dosing on potentially invalid, prolonged "flat HIGH" (~400 mg/dL) CGM readings. Trio would continue dosing as normal as long as there were not three consecutive 400 mg/dL readings. Once three consecutive HIGH values occurred, SMBs/TBRs would stop until either a finger stick value or a CGM reading < 400 mg/dL was received, which reset the guard and allowed dosing to resume; again until the next sequence of three HIGH readings.

In practice, this still led to the same problematic outcome in many real-world cases: the loop would stop corrections for extended periods and users remained stuck high. Simply because users would not enter finger stick values. It led to exactly the scenario you described.

The change in this PR removes that guard, thus this potential stall. It isn’t risk-free (sensor failures and miscalibration can happen, especially with some Libre setups), but the trade-off is clearer: prolonged high glucose with no corrections has been consistently more harmful in practice. We discussed this guard in detail at the Vancouver hackathon and agreed that removing it is the better path forward.

We’ll continue to monitor real-world data and feedback, and can introduce a more targeted safety check later if needed.

Thanks again for highlighting the user-impact perspective.

dnzxy:
I have had this run for the last ~2.5 hours with a modified glucose simulator that will keep producing stale 400 mg/dL values.
For the first ~1 hr, basal was kept at default 0.1 U/hr with MaxIOB at the maximum, 20 U. 
I then massively increased the profile basal to 2.25 U/hr and decreased MaxIOB to 2 U. 
Trio has kept looping nominally, while respecting other given safety thresholds such as MaxIOB, adjusting SMB and TBR dosing accordingly.

| Header | Header | Header |
|--------|--------|--------|
| image | image | image

From my testing here, this LGTM.

kingst (approved):
The code LGTM and I tested it in the simulator.

marionbarker:
## Test Details

Results are visible at this Nightscout URL:
* https://mope.strangled.net/?token=read-238a3eb0eaf97343

Test started at 2025-Nov-18 at 20:30 PST.

### Configuration

Test phone is an SE 3rd gen phone running iOS 26.1
* pump is rPi DASH simulator
* CGM is Nightscout fed by a cgm_max_500 test pattern
   *  glucose ramps up to reach a max value of 500 and remains there before ramping down
   * the pattern exceeds 400 for one hour
   * configure the pattern to repeat 2 times and capture the results in Nightscout
* start with this branch and merge latest dev (0.6.0.11) into it (there were no merge conflicts)

### Therapy Settings

| Setting | Schedule | Value |
|:--|:--|:--|
| max IOB | n/a | 20 U|
| max Bolus | n/a | 12 U |
| max Basal Rate | n/a | 2 U/hr |
| Glucose Target | 00:00 | 110 mg/dL |
| Basal Rates | 00:00<br>08:00 | 0.65 U/hr<br>0.85 U/hr |
| Carb Ratio | 00:00 | 10 g/U |
| ISF | 00:00<br>08:00 | 51 mg/dL/U<br>41 mg/dL/U |

### Algorithm

all nominal with these modifications:
* enable SMB always
* enable UAM
* part-way through first ramp up in glucose, increase SMB/UAM minutes from 30 to 90 minutes

Notice that max allowed glucose rise for SMB is 20%
* Some of the ramp up will be too big for SMB to be issued (the ramp is about 40 mg/dL per 5 minutes)
* Once glucose is above 230, then SMB should be possible and that is visible in the Nightscout URL.

MikePlante1:
I merged this PR'd branch into Trio:dev v0.6.0.10 and built it to the following setup:

iPhone SE (2nd gen)
iOS 18.7.2
OrangeLink
Medtronic 723
Anubis + G6 Sensor Simulator

I used the following settings:

| Setting | Value |
|---|---|
| Glucose Targets | 110 mg/dL |
| Basal Rates | 1 U/hr (24 h) |
| Carb Ratios | 10 g/U |
| Insulin Sensitivities | 40 mg/dL/U |
| Max IOB | 20 U |
| Max Bolus | 10 U |
| Max Basal Rate | 4.5 U/hr |
| Autosens Min | 70 % |
| Autosens Max | 120 % |
| Dynamic ISF | Disabled |
| Enable SMB Always | ON |
| Enable UAM | ON |
| Max SMB Basal Minutes | 60 min |
| Max UAM Basal Minutes | 60 min |
| Max Allowed Glucose Rise for SMB | 30 % |

I then cranked up the glucose level on the sensor simulator until it read HIGH, but there were periods of "Sensor Error" caused by the glucose jumps.

SMBs and Increased Basals continued to happen throughout it's nearly 2 hour HIGH line.





https://mike1ns.robot-armies.com/?token=pr846-12372ca1f69b0c0f

marionbarker:
## Test Results

This test was a success.

## Test Configuration

See https://github.com/nightscout/Trio/pull/846#issuecomment-3550826989 for test configuration.

### Summary

I ran 3 sets of ramps that went from 110 mg/dL to 500 mg/dL and flat-topped at 500 mg/dL for either 13 or 12 readings.

* The first ramp used a Delivery Limits setting of max IOB = 20 U and that value was never approached
   * max value of IOB reported was 10.64 U
   * there were several times the rPi needed to be restarted - those are seen as times when SMB stopped and the Temp Basal halted at the end of 30 minutes
* The second ramp used a Delivery Limits setting of max IOB = 8 U and that value was approached but never reached
   * max value of IOB reported was 7.61 U
   * the rPi worked without issue for this test
* The third ramp was run after a pause to allow the active IOB to decay 
   * the third ramp used a Delivery Limits setting of max IOB = 6 U and that value was approached but never reached
   * max value of IOB reported was 5.82 U
   * there was one time the rPi needed to be restarted during this test

Note: it is expected that max IOB will never be reached because after the app decides how much increased insulin can be delivered at time t0, the insulin is delivered, but when IOB is reported at time t1, additional active insulin decay is considered. This behavior is also seen with Loop. Using automated delivery, max IOB is never reached.

### Nightscout Graphic

Trio-PR-846-Nightscout-results

marionbarker (approved):
Approve from Code Review and Test.

kingst:
Quick note: I forgot to merge the JS first! We'll take care of that and do a mini PR to fix the sha text file after.

Issue #847, opened 2025-11-16
HUD insulin delivery display incorrect for scheduled basal and suspensions


itsmojo:
## Describe the bug

Trio’s HUD insulin delivery rate display will show incorrect rate info when the pump is currently running a Scheduled Basal Rate (SBR) or is suspended (or not available). In these cases, Trio will generally display the last TBR run (even if from a previous pod) and in certain rare startup cases “No Data” or “Pump suspended”.

## Attach a Log
Not needed

## To Reproduce

Generally you can use a pod (or pod simulator) and easily reproduce these issues by comparing the Insulin Delivery info in the Omnipod DASH view (which is always correct) to Trio’s HUD insulin delivery (which generally is the last TBR enacted) and comparing the differences when alternating between the doing pod Suspends & Resumes (before the next loop is enacted if Closed Loop is enabled).

*Steps to reproduce the behavior:*
Here’s an easy method to show reproduce the incorrect HUD insulin delivery display for both a suspended pump and a SBR. You can also get varying results on a newly paired pod or a first pod after adding a pump manager that I'll skip. It should be possible to run the same basic tests on other pump types and get similar results (for this example at least).

1. Pair a test DASH pod (either real or pod sim). Allow to run in closed loop mode until at least one loop operation is enacted and Trio sets a TBR whose value will be shown next to a blue drop on the right side of the HUD.
1. Touch the reservoir info on the left side of the Trio home view to go to the Omnipod DASH view. The Insulin Delivery rate should match the value shown in the Trio HUD insulin delivery rate next to the blue drop on the right side.
2. Tap “Suspend Insulin Delivery”, select 30 minutes and wait for operation to complete. Note that the Omnipod DASH view will now say “Insulin Suspended”. Now tap “Done” to return to the Trio home view.
3. Observe that the left hand side of the HUD does say “Insulin Suspended”, but that the right hand side of the HUD by the blue drop will incorrectly & inconsistently still show insulin delivery of the last TBR.
4. Alternately go back and forth between the Trio home view and the Omnipod DASH view and see that the Trio HUD continues to show an incorrect insulin delivery rate of the last TBR.
5. From the Omnipod DASH view tap “Resume Insulin Delivery” and wait for the display to be updated. The Omnipod DASH should now say “Scheduled Basal” and give the current scheduled insulin delivery rate right above the Activity label (if a loop doesn’t get enacted right away).
[...]

Issue #845, opened 2025-11-15
Pumps are incorrectly initialized with a flat 1 U/hr basal schedule


itsmojo:
## Describe the bug

Pump managers are incorrectly initialized with default fixed values instead of using the values from Trio’s Therapy Settings. This means that pump’s basal schedules and maximum basal rate & bolus (where applicable) will initialized to incorrect values. These incorrect values will probably remain in place until new Therapy values are set and saved in Trio while the pump manager is configured and must be repeated for each subsequent new pump type configured.

## Attach a Log
None needed

## To Reproduce

Use the PumpManager Details diagnostic to see that it didn’t receive the basal schedule and max basal values from Trio Therapy settings when adding an OmniBLE DASH (or Omnipod) pump manager.

*Steps to reproduce the behavior:*

1. Note Trio’s Therapy Settings for the Basal Schedule and Maximum Basal Rate, changing them to something else and saving them if they are still the default flat 1 U/hr basal schedule and 2 U/hr values respectively. 
2. Delete the current pump (e.g., for Omnipod DASH scroll to the bottom of the pump settings after deactivating any pod and selecting “Switch to other insulin delivery device”).
3. Click on “Add pump” and select Omnipod DASH. Go thru the setup screens using any value to advance to the “Pair Pod” Screen and the hit Cancel. Tap on “No Pod” in the HUD to reenter the Omnipod DASH view.
4. Select Pod Diagnostics->PumpManager Details option and find the ## OmniBLEPumpManagerState values.
5. Note that basalSchedule value contains a single entry with a rate 1.0 and a startTime of 0.0 and *not* the scheduled basal from Trio’s Therapy Settings.
6. Also note that the maximumTempBasalRate value is 2.0 and *not* the value from Trio’s Therapy Settings.

## Expected behavior
The Pod Diagnostics->PumpManager Details for OmniBLEPumpManagerState.basalSchedule and maximumTempBasalRate should match those in Trio’s Therapy Settings in steps 5 & 6 above.

## Screenshots
Example screenshots taken for the above sequence.

Trio's Therapy Settings

<img width="1284" height="2778" alt="Image" src="https://github.com/user-attachments/assets/0629271c-3bf4-405b-be04-da505289814b" />

![Image](https://github.com/user-attachments/assets/17cd42b4-0a0d-4c7a-9204-6bfb359bfc86)

[...]

PR #849, opened 2025-11-17
HUD insulin delivery display changes for Trio issue #847

itsmojo:
Fixes various Trio HUD issues mentioned in Trio issue #847

Updated Trio HUD view for a suspended pump
1) Omnipod DASH - pod suspended
2) Trio HUD - insulin delivery has  Pump suspended

Updated Trio HUD view for a pump running a scheduled basal
3) Omnipod DASH - scheduled basal
4) Trio HUD - insulin delivery has scheduled basal rate


Trio/Sources/APS/DeviceDataManager.swift
```diff
@@ -411,6 +415,25 @@ extension BaseDeviceDataManager: PumpManagerDelegate {
             bolusTrigger.send(false)
         }
 
+        /// The new suspended and scheduledBasal PassthroughSubject var's are set here based on the PM's basalDeliveryState.
```
dnzxy: This is a comment that explains a code change, but does not need to be here going forward.
Please remove.

Trio/Sources/APS/Storage/TDDStorage.swift
```diff
@@ -692,3 +659,49 @@ extension Decimal {
         return result
     }
 }
+
+/// Moved out from findBasalRate()
```
dnzxy: Remove this line please.

Trio/Sources/Modules/Home/HomeStateModel+Setup/PumpHistorySetup.swift
```diff
@@ -41,15 +41,14 @@ extension Home.StateModel {
         insulinFromPersistence = insulinObjects
 
         manualTempBasal = apsManager.isManualTempBasal
-        tempBasals = insulinFromPersistence.filter({ $0.tempBasal != nil })
+        tempBasals = insulinFromPersistence.filter { $0.tempBasal != nil }
 
+        /// The suspensions variable is the list of all pump suspend and resume events
```
dnzxy: Let's remove this comment and rename the variable to `suspendAndResumeEvents`

Trio/Sources/Modules/Home/HomeStateModel+Setup/PumpHistorySetup.swift
```diff
@@ -41,15 +41,14 @@ extension Home.StateModel {
...
         suspensions = insulinFromPersistence.filter {
             $0.type == EventType.pumpSuspend.rawValue || $0.type == EventType.pumpResume.rawValue
         }
-        let lastSuspension = suspensions.last
 
-        pumpSuspended = tempBasals.last?.timestamp ?? Date() > lastSuspension?.timestamp ?? .distantPast && lastSuspension?
-            .type == EventType.pumpSuspend.rawValue
+        /// Determining if the pump is currently suspended is now handled in DeviceDataManager
```
dnzxy: Again, a comment that explains a change brought in with this PR; not really a sensible documentation in code.

Trio/Sources/Modules/Home/HomeStateModel.swift
```diff
@@ -92,7 +91,7 @@ extension Home {
         var fetchedTDDs: [TDD] = []
         var insulinFromPersistence: [PumpEventStored] = []
         var tempBasals: [PumpEventStored] = []
-        var suspensions: [PumpEventStored] = []
+        var suspensions: [PumpEventStored] = [] // all of the pump suspend and resume events
```
dnzxy: See my other comment; rename this variable + remove comment.

Trio/Sources/Modules/Home/View/HomeRootView.swift
```diff
@@ -467,31 +496,38 @@ extension Home {
...
-                            if tempBasalString.count > 5 {
-                                Text(tempBasalString)
-                                    .font(.callout).fontWeight(.bold).fontDesign(.rounded)
-                                    .lineLimit(1)
-                                    .minimumScaleFactor(0.85)
-                                    .truncationMode(.tail)
-                                    .allowsTightening(true)
+                            if let basalString = self.basalString {
+                                // If running a scheduled basal, display basalString in loopGray instead of black
+                                let color: Color = state.apsManager?.isScheduledBasal == true ? .loopGray : .black
```
dnzxy: The color can stay the same if it's the correct value, no need to change it.

We will very likely remove this information entirely in upcoming home screen changes to enable better accessibility of Trio.

Trio/Sources/Modules/Home/View/HomeRootView.swift
```diff
@@ -467,31 +496,38 @@ extension Home {
                         .font(.callout)
                 } else {
                     HStack {
-                        if state.pumpSuspended {
+                        if state.apsManager?.isScheduledBasal == nil {
+                            /// The pump is not currently available (e.g., no pod) so
+                            /// display no insulin delivery info rather than "Pump suspended"
+                        } else if state.apsManager.isSuspended {
```
dnzxy: If the pump is suspended, let's display the last rate that was active and style it striked through. 

Depending on localization and text enlargement / zoom factor, this label here will break the display. 

I realize this isn't a thing brought in with this PR, but please make this change, i.e., when pump is suspended, display the TBR / SBR rate as

~~💧 1.0 U/hr~~

dnzxy (changes requested):
I have added a few comments and change requests to this PR.

I am not certain whether the changes introduced in this PR will seamlessly work with all other pumps Trio supports. I am aware that `MinimedKit`, `DanaKit` and `MedtrumKit` all technically provide a `BasalDeliveryState` via `PumpManagerStatus` but this definitely needs testing. 

> The graphic insulin display also is missing code for handling SBR’s (and possibly suspends) which should also be looked into.

This will require bigger changes in all pump drivers (including `OmniBLE` / `OmniKit`), so that the respective pump managers actually emit scheduled basal events, not "just" temporary basal and bolus.

itsmojo:
> I am not certain whether the changes introduced in this PR will seamlessly work with all other pumps Trio supports. I am aware that `MinimedKit`, `DanaKit` and `MedtrumKit` all technically provide a `BasalDeliveryState` via `PumpManagerStatus` but this definitely needs testing.

Agreed that this code should be tested for all pump types. But there's no reason why this shouldn't work as expected since Loop has always been using `BasalDeliveryState`.

> > The graphic insulin display also is missing code for handling SBR’s (and possibly suspends) which should also be looked into.
> 
> This will require bigger changes in all pump drivers (including `OmniBLE` / `OmniKit`), so that the respective pump managers actually emit scheduled basal events, not "just" temporary basal and bolus.

Agreed that more work will be needed (outside the scope of this PR) to improve the graphic insulin display for SBR, but this shouldn't be a blocker for getting the HUD insulin delivery information correct (which is the purpose of this PR). It might be possible to identify periods of SBR within Trio after some pending Omni PR's land which might be sufficient for the graphic insulin delivery graph. In the even longer run, having all the PM's use a new basal pump event with dosing amounts is another item that will require more discussion, development and testing for all pump types.

Trio/Sources/APS/Storage/TDDStorage.swift
itsmojo: Will do. This was meant as a placeholder since I never got an answer about where this routine might be moved if was a separate file. I'll take it that I should just leave this utility function here for now unless told to do move it to a definitive location.

Trio/Sources/Modules/Home/HomeStateModel+Setup/PumpHistorySetup.swift
itsmojo: Will do.

Trio/Sources/Modules/Home/View/HomeRootView.swift
itsmojo: Not sure what the "if it's the correct value" reference is actually about (it'll be the correct SBR as long as the Trio's basal schedule is synced the pump [note that this might not happen with https://github.com/nightscout/Trio/issues/845] and there isn't some timezone offset issue occurring). I believe that using an alternate color for SBR here is useful to people who might pay attention to such things at times without being much of a difference for anyone who doesn't care. In any case, if it is very likely that this information will be removed entirely (this is all news to me), I'd really prefer to keep this in as it is quite useful for testing, evaluation and understanding what the pump is actually doing at the present time.

Trio/Sources/Modules/Home/View/HomeRootView.swift
itsmojo: I believe that the HUD insulin delivery display should show of the correct current delivery state of the pump. Displaying a lined out value from the past for a suspended pump really seems like the wrong way to go. The old TBR could be from a long time in the past, a previous pod, or a manual TBR (which has its own variation for display). And in the case of a SBR, this rate from the past changes over time and the user could have updated the basal schedule resulting in a different SBR that that would have been run at the time the pump was suspended (or should it be at the current time). It's hard to even say what the semantics of value of the crossed out TBR / SBR should be for many edge cases. In any case, the "Pump suspended" text has always been there in iAPS and Trio, it just hasn't been getting displayed when it should recently because the code which sets the variable for this condition was changed and has been broken for most cases (although it can still happen with a new pump manager instantiation). I'd really like keeping Trio's long standing "Pump suspended" display here when a pump is suspended to keep this straight forward and consist and just make it work as it should (which the current code now does).

Trio/Sources/Modules/Home/HomeStateModel+Setup/PumpHistorySetup.swift
itsmojo: Will do

Trio/Sources/Modules/Home/HomeStateModel.swift
itsmojo: Will do

Trio/Sources/APS/DeviceDataManager.swift
itsmojo: Will do

Trio/Sources/Modules/Home/View/HomeRootView.swift
dnzxy: Before this PR, it would (incorrectly) display the latest TBR's value, even if before a suspension, it may have actually been running an SBR. You have fixed it with this PR. 

If you are interested in getting your current basal rate, you can at all times tap on the pump infos in the header (the entire area to the left of the "glucose bobble") and grab that info right from there. The currently running rate, regardless whether SBR or TBR, is a secondary set of information, and as hinted, may be removed in the future for further accessibility improvements of Trio's main screen UI. 

I don't agree that the type of basal (or that it's even basal) is important here, hence the color change isn't necessary IMO. Insulin is insulin. Generally, we need to simplify things.

Trio/Sources/Modules/Home/View/HomeRootView.swift
dnzxy: I agree, that just displaying a rate without _any_ change in the UI might implicate that the rate is still running (it creates inconsistency in the UI of the header), so would be wrong.

However, when the pump is suspended, it already displays a big fat "Insulin suspended"to the left to the "glucose bobble". IMO it's unnecessary to display the suspension state twice.  Hence, it will say that it's suspended to the left for the "pump info area", and when it is in this state, it displays the last "correct" rate (you introduced this fix with this PR), so tells the user what was happening just before the suspension, yet strikes this value, as the suspension withholds any insulin from being dosed.

dnzxy: My comment may be confusing when just looking at the current state of Trio's UI, so let me add a side note for context:
I am asking you to remove the duplicate suspension label because I am already thinking of future improvements for accessibility, and we will need to look into further simplifying the header and the amount of content/information in the screen, to allow for improved text enlargement support. It may not be an issue now, or for the way you and I run the app, but it is an issue for users using zoom and depending on the language of the user.

itsmojo: Done

Trio/Sources/Modules/Home/View/HomeRootView.swift
itsmojo: Used a different opacity instead of an alternate color for SBR display

marionbarker:
## Test

Summary: 
* ✅ for closed loop test
* ❓for canceling TBR on MDT pump - pretty sure that is an MDT pump manager feature

### Configuration

To make this easier to test, I created a local branch, `dev_plus_849_859_861`, that has code from this PR along with updates from PR #859 and #861. Actually - PR 861 isn't really relevant to this test, but it prevents accidents if this gets built onto the phone currently connected to a Medrum patch.

### Medtronic Test

### test 1 - initial add with closed loop enabled
Begin with MDT 523 with internal therapy settings of 11.5 U & 3.5 U/hr; single basal rate of 1.0 U/hr.
Begin with Trio app with Therapy settings of 8 U & 4 U/hr; 4 basal rates, 00:00 0.65 U/hr, 08:00 0.85 U/hr, 18:00 0.65 U/hr, 20:00 0.85 U/hr
* Add Medtronic Pump MDT 523
* Observe the main screen and grab screen shots
* Observe the settings inside the pump are correct (consequence of PR 859)
* From left to right
   * main screen following initial addition of the pump
   * ✅ MDT screen (the correct scheduled basal rate of 0.85 U/hr) 
   * ✅ main screen following suspend and then resume of MDT to force update
      * in this case, the pump is still running scheduled basal of 0.85 U/hr highlighted with red rectangle

pr_849_859_861

### test 2 - cancel temp basal set by Trio on the pump

(❓) - this test failed but pretty sure this is a feature of MDT pump manager (it behaves the same way on Loop)


Put Trio into open loop
* On pump, cancel the temp basal of 0.8 U/hr, pump thus resumed scheduled basal of 0.85 U/hr
* suspend and resume to force an update - do this on both the pump and using the app
* The display continues to show 0.8 U/hr for both the main screen and the MDT screen

### test 3 - closed loop test
Enable closed loop and observe screen before the next loop
* ✅ main screen immediately updated to correct scheduled basal of 0.85 U/hr
* ✅ MDT screen also updated to correct scheduled basal rate of 0.85 U/hr
* ✅ suspend MDT and observe the MDT screen and main screen all say suspended (the location that shows current basal rate is left blank during the suspend and is updated upon resume)

Observe several TBR set by Trio in closed loop, the 3 screens: main, MDT and pump itself agree each time.

### TBR Comment - Pump Specific Behavior

* With MDT
   * a suspend / resume does not modify temporary basal value or duration - it continues after delivery resumes 
   * an updated basal rate schedule does not modify temporary basal value or duration - it continues after delivery resumes
* With an Omnipod Eros or DASH
   * a suspend / resume returns the pod to the scheduled basal rate until the next cycle
   * an updated basal rate schedule is loaded by first suspending and then loading the new rate and resuming with the scheduled basal

The Medtrum instance does not have this PR applied but it looks like Medtrum behaves similar to the Omnipod

dnzxy (approved):

PR #851, opened 2025-11-19
[Part 1 of 2] Archive deleted glucose readings for CGM backfill

kingst:
This PR introduces a new table in CoreData to store deleted glucose readings. Storing deleted glucose values allows us to avoid backfilling already deleted items after we re-enable CGM backfills.

This is part 1 of 2 to address this issue: https://github.com/nightscout/Trio/issues/737


dnzxy (approved):
I have tested this PR in conjunction with #853 and can report that it works as expected.
More detailed test report on #853.

Optional/nit: I don't really like `deletedGlucoseDeletion`, but not sure what else to propose here. Maybe `archivedGlucoseDeletion`?

kingst:
@dnzxy I agree, I made the change. This one is ready to merge when you're ready.

dnzxy (approved):

PR #853, opened 2025-11-20
[Part 2 of 2] CGM Backfill Support

kingst:
This commit adds logic to support backfilled CGM values from CGM managers. There are a few key corner cases that it handles:
    - _Not_ backfilling previously deleted CGM readings
    - Deduplication for CGM readings
    - Avoiding adding CGM readings when values already exist
    
The way we handle it is by comparing timestamps from both the DeletedGlucoseStored and GlucoseStored CoreData tables.
    
Our logic isn't perfect, because we're only looking at timestamps. But it's simple and predictable. Since these are corner cases that we don't expect to happen often in practice, having a less than perfect but simple solution is preferred.

This PR depends on this one, which adds the deleted glucose archive: https://github.com/nightscout/Trio/pull/851


dnzxy (approved):
I have tested this in-vivo with the following setup

## Test Result: **successful**
* Backfill nominal
* Deletion of glucose entries nominally
* Respects previously deleted (backfilled) entries

| Key | Value |
|--------|--------|
| Phone | iPhone 15 Pro |
| iOS | iOS 18.7.1 |
| CGM | Dexcom G6 + Anubis |
| CGM App | xDrip4iOS v6.1.0 |
| Pump | Omnipod Dash |

The Trio version I ran is basically currently private `oref-swift-medtrum` with PR #851 and #853 applied. 

## Test Procedure
* Delete a few entries in the last 60 minutes
* Toggle off Bluetooth
* Force quit Trio and xDrip4iOS
* Wait for 15 minutes
* Toggle on Bluetooth again
* Open xDrip4iOS and Trio, wait for next CGM reading
* Observe: readings get filled in successfully; previously deleted entries are not backfilled
* Repeat test once more. This time delete one entry that was part of the previously backfilled 3 readings
* Observe: readings get filled in successfully; previously deleted + backfilled entries are not backfilled

I approve of this PR based on testing and functionality-wise.

I am uncertain how to address the various SonarQube issues.

MikePlante1:
Testing setup:
* iPhone SE (2nd gen)
* iOS 26.1
* Anubis + G6 Sensor Simulator Hardware
* Dexcom G6 app v1.14.0 (SW11677)
* OrangeLink + Medtronic 723
* Trio:dev v0.6.13 + this PR'd branch merged in

## ✅ Test 1
* Place CGM and iPhone on opposite sites of building out of reach for ~23 minutes
* **Observed:** 4 readings missed
* Place CGM and iPhone back next to each other
* **Observed:** When 5th reading came in, it populated in Trio as well as all 4 missed readings
* Note: I did confirm the 314 vs 315 readings matched between Trio and G6 app

| Trio Main | Trio History | Dexcom G6 |
|---|---|---|
|  |  |  |
|  |  |  |

## ❎ Test 2
* Close Trio App but leave Dexcom G6 app open for 12 minutes
* **Observed:** 2 readings missed in Trio but not in Dexcom G6 app
* Open Trio App
* **Observed:** Next reading populated in Trio but the missing readings did not backfill
* **_Not a blocker, as I did not expect this to succeed_**

| Trio Main | Trio History | Dexcom G6 |
|---|---|---|
|  |   |   |
|  |  |  |

## ✅ Test 3
* Delete last CGM reading from Trio
* Close G6 app (didn't want to walk the phone out of range of the G6)
* Wait a bit before re-opening G6 app
* **Observed:** When next reading came in, backfill was successful and did not include deleted reading

| Trio Main | Trio History | Dexcom G6 |
|---|---|---|
|  |  |  |
|  |  |  |

dnzxy (approved):

MikePlante1:
Couple more tests:

## ✅ Test 4
* Turn off phone for over 3 hours before turning it back on
* Ensure both Dexcom G6 app and Trio app are open and wait for next reading
* **Observe:** 3 hours of glucose data backfilled to both the Dexcom G6 app and Trio

| Trio Main | Trio History | Dexcom G6 |
|---|---|---|
|  |   |  |
|  |  |  |

## ⍰ Test 5
* Close the Dexcom G6 app to simulate the phone being out of range from the transmitter
* Wait for 2 missed readings
* Enter a manual glucose reading and force a loop cycle
* Open Dexcom G6 app and wait for the next reading
* **Observe:** First missed reading was backfilled, but not the second one that would be close to the same time as the manual glucose entry
* **_While strange, not that big of a deal. Probably not a blocker._**

| Trio Main | Trio History | Dexcom G6 |
|---|---|---|
|  |  |  |
|  |  |  |

MikePlante1:
To explain Test 5: Say you're at a concert or an apple store and your dexcom ain't reading for half an hour... so you fingerstick and enter that reading. When your Dexcom finally connects and backfills, it'll skip the backfill for the timeslot of the fingerstick reading that was manually entered. So again, not a big deal, and I think oref might pool cgm readings anyway? But if Trio is your only source of glucose readings in Nightscout, it'd be nice for it to include all of the readings.

MikePlante1 (approved):
Approving from testing. Feel free to address Test 5 in this PR, another PR, or just not address it at all.

kingst:
Thank you for the detailed testing, @MikePlante1 ! My vote is to move forward with this PR as is, here is my thinking:
- For test 2, we are limited to backfilling only what the CGM plugins tell us about and in this case the plugin doesn't know about the backfill so we are limited.
- For test 5, we could handle this but it would complicate the logic. With this feature I wanted to keep the logic very simple and make sure that we nail the common backfill case, like leaving your phone in another room for 10 minutes and making sure those two missing readings get logged. If we wanted to handle this case we'd track sources, filter out manual reading, etc but since we've already had one false start with it due to differences in how xdrip and dexcom readings get backfilled, so I wanted to keep the logic super simple. Would you be open to creating an issue describing the desired behavior around test 5?

MikePlante1:
Sounds good. With you, @dnzxy, and I in agreement, I will merge this now and open up an Issue for Test 5.

PR #858, opened 2025-11-23
chore: update submodules from upstream, translations and fixes

marionbarker:
## Purpose:

Update the Trio repositories to be aligned with the upstream repositories.

## Method:

In the case of all but 3 repositories, the `trio` branch of the loopandlearn org fork of the repositories is identical to the upstream branch (either dev or main at the LoopKit org).

The 3 repositories that differ are:
* CGMBLEKit and G7SensorKit - which have different values for the default for glucose upload
* LoopKit - which has support for Dana pump simulator
   * this is not merged into LoopKit because a modification was requested and has not been made yet

There are 2 repositories which, at one point did not align, and now do.
* RileyLinkKit and TidepoolService
* after merging dev into trio, it was noted that the code is identical (but commit history differed)
* to simplify things going forward, the trio branch at loopandlearn was deleted and created fresh to align with the dev branch from LoopKit
* only someone with a clone that pointed to the old trio branch will notice a commit history difference and they can delete their local trio branch and check it out fresh to align with loopandlearn

All other repositories are directly aligned with the upstream repo with trio matching the associated main or dev branch with identical commit history.

## Test:

Trio was built onto an iOS 18.7.2 device (SE Gen 3) using Xcode 16.4
* that phone already had a Medtronic pump attached
* Trio continues to behave nominally with that pump
* Delete the MDT 523 pump and Add an rPi simulated Dana pump
* Dana pump was added successfully
   * Not an issue with this PR and I think other PR are addressing this
   * Initially, reservoir continues to report the MDT value on the main screen while correct reservoir value is reported on the Dana screen - this was corrected at the next CGM cycle
   * The Dana basal rate is set to the Default Therapy setting (as noted in Issue #845)
      * Doing a basal rate change loaded the basal rates to the pump

## Unrelated Issue noted in this test

I'm recording this here out of basic laziness. Tagging @bastiaanv or maybe it's a Trio developer issue. 
The allowed Dana basal rates are for 0 to 3 U/hr with 0.01 U increment.
This is with an rPi simulated Dana pump.

Dana with Trio:
* Problem for a different PR: The basal rates are not properly formatted for DanaKit for some values
   * for example, picker shows 0.58999999999. . .
   * after save - it shows a value of 0 U/hr
   * if I pick one that shows a reasonable value in the picker, e.g., return to 0.80 U/hr, then that shows as the actual value following save

Dana with Loop:
* Loop uses a different picker system and had no issue setting an 0.59 U/hr basal rate.


marionbarker:
Sorry - got confused. I converted to a draft because updates are expected for MedtrumKit, then I remembered that MedtrumKit is in a different branch.

dnzxy (approved):

PR #860, opened 2025-11-24
Feature/dev/update type data object

nielsbjorn-da:
## What has changed:
- Switch BasalProfileEditor.Item and ISFEditor.Item
from reference types to value types (struct) to better match their use in Set and equality checks.
- Update validate() for both editors to:
Remove duplicate items,
Sort by timeIndex,
Force the first item’s timeIndex to 0 when items exist,
Only write back to items when the normalized list changed.

## Why it was changed:
To fix the UI.
The UI now registrates when events happens

### DIFF
--- a/Trio/Sources/Modules/BasalProfileEditor/BasalProfileEditorDataFlow.swift
+++ b/Trio/Sources/Modules/BasalProfileEditor/BasalProfileEditorDataFlow.swift
@@ -4,7 +4,7 @@ import Foundation
 enum BasalProfileEditor {
     enum Config {}
 
-    class Item: Identifiable, Hashable, Equatable {
+    struct Item: Identifiable, Hashable, Equatable {
         let id = UUID()
         var rateIndex: Int
         var timeIndex: Int
--- a/Trio/Sources/Modules/BasalProfileEditor/BasalProfileEditorStateModel.swift
+++ b/Trio/Sources/Modules/BasalProfileEditor/BasalProfileEditorStateModel.swift
@@ -139,14 +139,19 @@ extension BasalProfileEditor {
         }
 
         @MainActor func validate() {
-            let uniq = Array(Set(items))
-            let sorted = uniq.sorted { $0.timeIndex < $1.timeIndex }
-            sorted.first?.timeIndex = 0
-            if items != sorted {
-                items = sorted
-            }
-            calcTotal()
-        }
+                    var uniq = Array(Set(items))
+                    uniq.sort { $0.timeIndex < $1.timeIndex }
+
+                    if !uniq.isEmpty, uniq[0].timeIndex != 0 {
+                        uniq[0].timeIndex = 0
+                    }
+
+                    if items != uniq {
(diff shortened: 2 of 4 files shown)


dnzxy:
Hi, would you care to please share why this PR was done? What issue is it solving?

nielsbjorn-da:
> Hi, would you care to please share why this PR was done? What issue is it solving?

Hello @dnzxy,
Of course - I experienced trouble with saving new values in the ISF and basal editor. When I changed the values and went back to the window where i could hit save - the save button was still deactivated as it did not recognize the changes. This is a common issue when you use reference type instead of value type.

This PR does not change any of the behavior and do not have any consequences in regards to saving the data to the core.

dnzxy:
I‘m wondering whether you were on an older version of dev or something. This should have been addressed with #819. It was one of the iOS 26 / Xcode 26 issues, as newer SwiftUI seems to handle the state update (when referencing things) differently than it did on older versions.

MikePlante1:
This was essentially how I had [fixed](https://github.com/nightscout/Trio/pull/781) the ISF before it was refactored. The refactor seemed to have fixed it for me, but I remember it seeming like ~~class~~*struct was the “proper” type for this variable.

dnzxy:
Well, this PR then leaves me with two questions:
- Was this fixed with the refactor or not? 
- What about glucose targets and carb ratio?

dnzxy:
@nielsbjorn-da @MikePlante1 any insight?

MikePlante1:
I had addressed ISF, CR, Target, and Basal in my closed PR. I tested the current dev on my phone in both mg/dL and mmol/L and couldn’t get it to break, so I think 0.6.0.10 made this no longer an issue. That said, just because it currently works doesn’t mean it shouldn’t be updated to how it’s apparently supposed to be implemented (struct instead of class).

@nielsbjorn-da , do you have a specific ISF combo that breaks it so we can reproduce the error? (e.g.: set ISF to 2.5 mmol/L and then try changing it to 2.6 mmol/L)

nielsbjorn-da:
Hello @dnzxy and @MikePlante1,

I’ve just come back from travel. 🙂

I’ve tried different combinations for testing both ISF and basal without being able to break it or make the app shut down on dev on my phone.

I think you’re right, @dnzxy — I should implement a solution for both target and carb ratio as well, as from my perspective a struct is better to use than a class.

Is this something your guys want me to focus my time on?

nielsbjorn-da:
@MikePlante1 I can see that after the UI has been updated, the error is no longer there.

Previously, if I went in and changed the value in the editor, it didn’t register the changes, so I couldn’t save them. I had to delete an item before I was able to save (to force an update).

dnzxy:
Shall we close this PR then @nielsbjorn-da ?

dnzxy:
Closing this for now. Please feel free to re-open PR or discussion if you still come across this issue and feel it's a required change to be introduced.

Thanks for looking into this :)

PR #872, opened 2025-12-02
Improve time remaining string for <12 hrs #865

### LINKED ISSUE #865: Consider presenting pod hours remaining differently
Two minutes after a scheduled 12h expiration reminder, the pod information screen says that there are 11h remaining, which gives off buggy vibes. (11 ≠ 12)

After an expiration reminder has been triggered, consider presenting remaining time as:

- 11.9 h remaining or
- 11h 58m remaining 

This format could be applied to both the HUD and the pump information screen. 

Trio version 0.6.0.16

Image
Image
Image

dnzxy:
This PR refines the pod time-remaining display to be more accurate and less confusing, especially right after the 12-hour reminder.

### Changes

* For **1–11 hours remaining**, the UI now shows **hours + minutes** (e.g. `11h 58m`).
* For **>=12 hours**, behavior is unchanged, show **hours only** (e.g. `12h`).
* For **<1 hour**, still show **minutes only** (e.g. `47m`).

Addresses #865.

### DIFF
--- a/Trio/Sources/Modules/Home/View/Header/PumpView.swift
+++ b/Trio/Sources/Modules/Home/View/Header/PumpView.swift
@@ -142,7 +142,12 @@ struct PumpView: View {
         }
 
         if hours >= 1 {
-            return "\(hours)" + String(localized: "h", comment: "abbreviation for hours")
+            var remainingHoursString = "\(hours)" + String(localized: "h", comment: "abbreviation for hours")
+            if hours < 12 {
+                remainingHoursString += " " + "\(minutes)" +
+                String(localized: "m", comment: "abbreviation for minutes")
+            }
+            return remainingHoursString
         }
 
         return "\(minutes)" + String(localized: "m", comment: "abbreviation for minutes")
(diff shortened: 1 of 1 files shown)


MikePlante1 (approved):
Code review LGTM.

I was initially going to question why you didn't add minutes for 12h00m–23h59m, but then realized it's because that's when the hourglass switches from a full hourglass to a half-full hourglass. Makes sense.

I think you can change the time remaining on a DashPi, so I can attempt that to test this when I get home tonight, unless someone beats me to it.

dnzxy:
> I was initially going to question why you didn't add minutes for 12h00m–23h59m, but then realized it's because that's when the hourglass switches from a full hourglass to a half-full hourglass. Makes sense.

FWIW, I asked myself _the exact same thing_ when I implemented it, but came to the same conclusion as you did 😂

MikePlante1:
I’ll update this comment later with screenshots, but tested this with an Eros pod that had 17h left and then again at 6h38m left, and both looked good.

PR #874, opened 2025-12-04
Feature: barcodescanner for quick carb entry of processed food

EJM0:
### Shortcut 
adds the ability to open the scanner directly via the action button or lock screen.

### Search Bar (Treatment View & List View)
I recommend logging in via OpenFoodFacts because text search is currently rate-limited for anonymous users, although the barcode API is not affected. By logging in, users can also correct incorrect data, improving their own results while contributing to the quality of the OpenFoodFacts database. Justifying the added usage to their service.
<div style="display: flex; gap: 10px;">
  
  
</div>

---

### Barcode Scanner
<div style="display: flex; gap: 10px;">
  
</div>

---

### List View (Items Added to Meal)

- Items can be removed or edited via left and right swipe gestures.

<div style="display: flex; gap: 10px;">
  
</div>

---

### Edit View of Items

- This view appears after scanning an item to verify that the nutritional information is correct.
- To confirm, tap **Add**.
- The same view is used for updating nutritional information and can also be reused for future meal presets.
- If the user notices incorrect data, it can be edited.
- If logged in via OpenFoodFacts, users can submit corrected data to the database.

<div style="display: flex; gap: 10px;">
  
  
</div>

---
#### Treatmentview with added item
<div style="display: flex; gap: 10px;">
  
</div>

---

### Settings
<div style="display: flex; gap: 10px;">
  
  
</div>


kingst (changes requested):
I'm concerned about the use of AI/ML here without any sort of design doc. Given that insulin is a dangerous hormone, anything that uses AI/ML, especially via cloud services, should have a well articulated safety and security discussion.

I love the initiative of this feature, but from a quick glance at the code I'm less enthusiastic about the maintenance overhead given the benefit vs complexity that I can see. Long term I'd like to see Trio move toward no meal announcements vs tools to help with meal announcements, but that's just my own personal opinion.

Architecturally, have you considered using a separate app for this and interfacing with Trio via a universal link or some other IPC mechanism?

As a concrete next step I'd like to suggest starting with a short design document.

EJM0:
I thought about implementing it as a separate app, but with that would come unnecessary complications.
In my opinion, it’s just a better quality of life to have an all-in-one approach as long as it is not too much in one.
I think it would be a good improvement if we implement the barcode feature and discard the ai feature.
Additionally, I’d look into nutritional label detection to eventually strengthen the reliability of the feature since the openfoodfact db does have some inconsistencies in the data it provides for the food that is scanned.
Also im aware of the poor code quality that is a product of vibe coding it together in 1h but id like to refine the code to make it also better for maintaining.

EJM0:
https://github.com/user-attachments/assets/311f1f6e-ff84-4d4d-b1f3-31eb125979c4

MikePlante1:
At this time, we will not be accepting any PRs that require linking an LLM to use them. Please strip the AI parts out. Thank you.

EJM0:
Hi @MikePlante1 , does that mean specifically using LLM for dosing? If yes, I stripped that out of the code. The only "ai" used here is [openfoodfact-extractor](https://huggingface.co/openfoodfacts/nutrition-extractor). Which is only used for extracting nutritional labels from food that is not covered by the barcode db. Therefore, it does not do any carb predictions and also not falling under this [case](https://github.com/nightscout/Trio/pull/888#issuecomment-3651397154).

EJM0:
https://github.com/user-attachments/assets/77eb01f4-8b8a-475f-9803-980462b34f4c

Current state

MikePlante1:
> Hi @MikePlante1 , does that mean specifically using LLM for dosing? If yes, I stripped that out of the code. The only "ai" used here is [openfoodfact-extractor](https://huggingface.co/openfoodfacts/nutrition-extractor). Which is only used for extracting nutritional labels from food that is not covered by the barcode db. Therefore, it does not do any carb predictions and also not falling under this [case](https://github.com/nightscout/Trio/pull/888#issuecomment-3651397154).

To quote the comment you linked from @dnzxy :

> We will not accept anything into Trio that requires users to add Claude (or any other LLM) to the app at this point in time.

So please strip it out for this PR. If this policy changes in the future, you'd be welcome to PR it again.

EJM0:
Okay I'll do so :)

dnzxy:
I haven’t had time to check this out in detail or run in in-vitro/in-silico, but could you give a brief overview of the workflow for the user. How do you access the scanner? How do you use fiber vs. grams of carbs vs. fat and protein when confirming a scanned article? 

When it comes to the UI, please have a look at how we‘ve designed various elements in Trio thus far. The NS setting views give a good overview of the use of `Section`s and various button styles. Similar for swipe actions (compare w/ carb entries or adjustment presets). 

I can take a closer look around this upcoming weekend and give more dedicated feedback ☺️ just from your recording in your second-to-last comment, it feels a bit out of place and not Trio-esque, yet. We can work on that though.

EJM0:
I will review the NS settings to understand the UI structure and improve the design coherence. For this I would appreciate your feedback this weekend.

Further, here is the workflow overview:

1. User long presses the boluscalc button to open the scanner UI.
2. The user can scan barcodes and directly enter the amount consumed in the Editorview.
3. Scanned items are combined in the listview and can be edited or removed if incorrectly added.
4. The user can access the treatmentview directly from the scannerview or listview if one item is in the list.
5. Based on settings, only carbohydrates, fat and protein are summarised and passed to the treatmentview. Other nutritional facts are obsolete and can be theoretically removed from the views.(if carb only it toggled only carbs are send to treatmentview. This behavior could also be toggled by setting FPUs True or false )

MikePlante1:
I gave this a test, and it's pretty cool! A lot of good stuff in there. Here's a demo.

https://github.com/user-attachments/assets/b6c5eeb6-f2d5-44e4-a2f7-72d3fb2f80ed

Some constructive criticism:

I noticed several typos in the mini and verbose hints.

The verbose hint also doesn't tell you how to access the food scanner. While long-pressing the `+` button does streamline it to make the scanner faster to access, it's not very intuitive. Maybe a barcode.viewfinder SF Symbol to the right of the `Carbs` label? Just an idea, not sure if that's good UX. Maybe both?



I didn't have many barcodes within reach to scan, so I just tried it with three items. I suppose I could just bring up some labels on my computer screen to test/scan.

While "per 100 g" info is good and useful, it should probably have a "per servings" options as well, as all three products I tried only have their nutritional info printed in "per serving", which makes it difficult to confirm it's accuracy.

I did notice it automatically used the suggested serving size for two of my items, but my third item was liquid so it just defaulted to 100 g instead of 3 mL (which is equal to 3 g). Probably just a bad item to test, though, as it's a no-calorie item. I tested it with some drink barcodes I found online and it did properly auto-enter the serving size for those (such as 330 mL).

I didn't like how tapping "Add to List" automatically re-opened the scanner. Maybe have a text header that says "Add to list and" and then three side-by side buttons with options "Open List", "Open Calculator", "Scan Another"?

For the Scanned Items list, it only shows carbs and kcal, not fat and protein.

I notice you've increased the maximum length of the note from 25 to 100 characters. This makes it so the label can be bigger than the UI allows, which can make it impossible to see the entire string on the screen, except if you open Meals in History after it's already logged.

Not sure if we want to just use the Total Carbohydrates count from the label like this PR currently does, or if we also should consider how Sugar Alcohols or Dietary Fiber affect Net Carbs.

dnzxy:
Have you considered adding something like this to the feature @EJM0 ? Plus have you considered rather than adding a hidden "trigger" (force touch of the (+) button) to add something in Treatment View? Example from another app:
image

The placeholder says "What did you have for breakfast?" and you can either free-text search or scan a bar code. The open foods database you are using might expose both, food names and barcodes (?).  

Will be able to do a more detailed write-up once I ran it. Just got me thinking after your reply and Mike‘s video.

EJM0:
I think the extra search line/button in the treatment view would be nice. I would have to look into keyword search with the OpenFoodFacts API. First, I’d try improving the ui so that it's working fine. (Currently, I have some inconsistencies in the editor view when it’s displayed in a sheet or in normal presentation.)

The defaulting to scanner view is something I was not certain about. It could be helpful if you're eating multiple foods at once. But I think the most likely case is that the users mainly entered one processed item on the fly and then proceed to the treatment view.

As for the inconsistencies of the serving size automations, I noticed that some products in the db sould have them in their JSON response, out going from their nutritional labels, but they simply are not given. If the serving portion is given, I use it, but if not, I just default to 100g/ml. Vice versa for serving units, there the default is g.

It would be helpful if someone could do the part of writing the explanation of the settings properly, as English is not my main language, so the explanations can be formalised more intuitively and condensed.

MikePlante1:
> It would be helpful if someone could do the part of writing the explanation of the settings properly, as English is not my main language, so the explanations can be formalized more intuitively and condensed.

I can help with that, but I’ll wait until the rest of the PR is finalized first to avoid having to re-write it if things change.

EJM0:
Yes, that would be great thanks a lot:)

EJM0:
image
Okay added the scanner to the treatment view directly

EJM0:
next I'd try to implement the search feature

EJM0:
IMG_2147
IMG_2148

dnzxy:
Just because I keep getting notifications for this PR (not your doing, that's just what GitHub does to code owners) – you'll notify and mark this as ready once you are done tweaking, right?

EJM0:
Yes I'll do

EJM0:
Ready @dnzxy,
Would love a review form somebody more knowledgeable than me. :)

dnzxy:
Sorry, very busy with life recently and attempting to get dev to main.
This won't make the cut for the dev-to-main push, but will put it on the list to look at ASAP after that.

EJM0:
Okay, good to hear. Looking forward for the big release:) 🍾

dnzxy:
Just a heads-up: Given massive work being done on the backend side and the frontend side of Trio, I don't realistically see this landing anytime soon. It's been open for the better part of 8 months now, and unless someone takes on the task to ensure this fits in line with the design concepts and look+feel of Trio, I don't think it will move in before v1.0 or in the next few months, whichever comes sooner. 

I genuinely do not have the capacity to review this and given it carries 100+ commits and almost 30 changed files, not sure if this needs to be cleaned up, and broken up more.

Potentially @MikePlante1 or @t1dude or @marv-out want to take on this review process — for the above reasons, I won't.

t1dude:
I am closing this PR for now, for three reasons:

1. It needs to be re-engineered to be mergable. It's been open for a very long time and the codebase has changed a lot since it was opened.
2. The code with all its commits and file changes is very hard to review properly.
3. The code seems to be entirely vibe coded. Using AI code assistance is fine, but features as complex as this should have at least a foundation that is not vibed.

A new PR is welcome, but please address the points above. If you rely on AI code assistants, I suggest to start with smaller improvements/features.

EJM0:
Since this was my first commit, I guess it was a bit overambitious. I'll wait for the heavy merging of the Homeview rework to finish and then rework the code to have a better base.(Since then, much was learned from my side.) Also some changes on OpenFoodFacts side like better search api are on the horizon, so this also needs time.

Issue #829, opened 2025-10-29
Temp Target Guardrails Not Always Enforced


MikePlante1:
## Describe the bug
If you alter settings in a specific order, you can set custom sensitivity outside the intended guardrails.

## To Reproduce
*Steps to reproduce the behavior:*
1. Add a new TT of 105 mg/dL
2. Choose Custom Sensitivity Adjustment
3. Lower to minimum of 15%
4. Increase Target Glucose
5. Notice Sensitivity Adjustment is now lower than the guardrail minimum

## Expected behavior
It should always respect the guardrail.

## Screenshots

https://github.com/user-attachments/assets/c5e7e281-7645-4713-b5d2-de9129deab46

### Smartphone:
* Hardware: iPhone 14 Pro Max
* OS Version: iOS 26.1 Beta 4

### Trio Version:
* Version Number: v0.6.0.1 (no customizations)
* Repo: *nightscout/trio*
* Git Reference: [99e4a0d](https://github.com/nightscout/Trio/commit/99e4a0d04019b66b0a55ee8fc7a000854a813f97)

MikePlante1:
Ah, I just found a Draft PR #716 which should hopefully address this Issue, as well as other edge cases I had not noticed.

mountrcg:
#716  is not a draft, I have been using it since it's inception and it works nicely.

MikePlante1:
> [#716](https://github.com/nightscout/Trio/pull/716) is not a draft, I have been using it since it's inception and it works nicely.

Sorry, looking again I’m not sure where I saw “draft”. I’ll take a closer look and do some tests.

dnzxy:
> [#716](https://github.com/nightscout/Trio/pull/716) is not a draft, I have been using it since it's inception and it works nicely.

If people would reply to comments on that PR, it could also move foward 🤷

PR #878, opened 2025-12-07
Fix Add & Edit TT to have target value change HBT instead of sensitivity

mountrcg:
To fix issue #829 and replace #716.

needs extensive testing. Basically it only calculates the sensitivity (Insulin percentage) when the target value is choosen and the current HBT is the same as the one in Settings, which is always the case when you switch from a standard sensitivity TT to a customized one. After that, when changing target values the current state sensitivity is kept and the HBT calculated to be aligned for this new TT and current sensitivity.

### DIFF
--- a/Trio/Sources/APS/OpenAPS/OpenAPS.swift
+++ b/Trio/Sources/APS/OpenAPS/OpenAPS.swift
@@ -549,9 +549,14 @@ final class OpenAPS {
 
                 if let effectiveHBT, effectiveHBT != defaultHalfBasalTarget {
                     adjustedPreferences.halfBasalExerciseTarget = effectiveHBT
+                    let percentage = Int(TempTargetCalculations.computeAdjustedPercentage(
+                        halfBasalTarget: effectiveHBT,
+                        target: targetValue,
+                        autosensMax: preferences.autosensMax
+                    ))
                     debug(
                         .openAPS,
-                        "checkStandardTT: Updated halfBasalExerciseTarget to \(effectiveHBT) for target \(targetValue) (stored HBT: \(String(describing: activeTempTarget.halfBasalTarget?.decimalValue)), settings HBT: \(defaultHalfBasalTarget))"
+                        "TempTarget: target=\(targetValue), effectiveHBT=\(effectiveHBT), percentage=\(percentage)%, adjustmentType=Custom"
                     )
                 }
             }
--- a/Trio/Sources/APS/OpenAPS/OpenAPS.swift
+++ b/Trio/Sources/APS/OpenAPS/OpenAPS.swift
@@ -534,15 +534,26 @@ final class OpenAPS {
 
         // Check for active Temp Targets and adjust HBT if necessary
         try await context.perform {
-            // Check if a Temp Target is active and if its HBT differs from user preferences
+            // Check if a Temp Target is active
             if let activeTempTarget = try self.fetchActiveTempTargets().first,
                activeTempTarget.enabled,
-               let activeHBT = activeTempTarget.halfBasalTarget?.decimalValue,
-               activeHBT != defaultHalfBasalTarget
+               let targetValue = activeTempTarget.target?.decimalValue
             {
-                // Overwrite the HBT in preferences
-                adjustedPreferences.halfBasalExerciseTarget = activeHBT
-                debug(.openAPS, "Updated halfBasalExerciseTarget to active Temp Target value: \(activeHBT)")
+                // Compute effective HBT - handles both custom HBT and standard TT (where HBT might need adjustment)
+                let effectiveHBT = TempTargetCalculations.computeEffectiveHBT(
+                    storedHBT: activeTempTarget.halfBasalTarget?.decimalValue,
+                    settingHalfBasalTarget: defaultHalfBasalTarget,
+                    target: targetValue,
+                    autosensMax: preferences.autosensMax
+                )
+
+                if let effectiveHBT, effectiveHBT != defaultHalfBasalTarget {
+                    adjustedPreferences.halfBasalExerciseTarget = effectiveHBT
+                    debug(
+                        .openAPS,
+                        "checkStandardTT: Updated halfBasalExerciseTarget to \(effectiveHBT) for target \(targetValue) (stored HBT: \(String(describing: activeTempTarget.halfBasalTarget?.decimalValue)), settings HBT: \(defaultHalfBasalTarget))"
+                    )
+                }
             }
             // Overwrite the lowTTlowersSens if autosensMax does not support it
             if preferences.lowTemptargetLowersSensitivity, preferences.autosensMax <= 1 {
--- a/Trio/Sources/APS/OpenAPS/OpenAPS.swift
+++ b/Trio/Sources/APS/OpenAPS/OpenAPS.swift
@@ -556,7 +556,7 @@ final class OpenAPS {
                     ))
                     debug(
                         .openAPS,
-                        "TempTarget: target=\(targetValue), effectiveHBT=\(effectiveHBT), percentage=\(percentage)%, adjustmentType=Custom"
+                        "TempTarget: target=\(targetValue), HBT=\(defaultHalfBasalTarget), effectiveHBT=\(effectiveHBT), percentage=\(percentage)%, adjustmentType=Custom"
                     )
                 }
             }
--- a/Trio/Sources/APS/OpenAPS/OpenAPS.swift
+++ b/Trio/Sources/APS/OpenAPS/OpenAPS.swift
@@ -541,7 +541,7 @@ final class OpenAPS {
             {
                 // Compute effective HBT - handles both custom HBT and standard TT (where HBT might need adjustment)
                 let effectiveHBT = TempTargetCalculations.computeEffectiveHBT(
-                    storedHBT: activeTempTarget.halfBasalTarget?.decimalValue,
+                    tempTargetHalfBasalTarget: activeTempTarget.halfBasalTarget?.decimalValue,
                     settingHalfBasalTarget: defaultHalfBasalTarget,
                     target: targetValue,
                     autosensMax: preferences.autosensMax
--- a/Trio/Sources/APS/OpenAPS/OpenAPS.swift
+++ b/Trio/Sources/APS/OpenAPS/OpenAPS.swift
@@ -534,7 +534,7 @@ final class OpenAPS {
 
         // Check for active Temp Targets and adjust HBT if necessary
         try await context.perform {
-            // Check if a Temp Target is active
+            // Check if a Temp Target is active and check HBT differs from setting and adjust
             if let activeTempTarget = try self.fetchActiveTempTargets().first,
                activeTempTarget.enabled,
                let targetValue = activeTempTarget.target?.decimalValue
--- a/Trio/Sources/Helpers/TempTargetCalculations.swift
+++ b/Trio/Sources/Helpers/TempTargetCalculations.swift
@@ -64,38 +64,8 @@ enum TempTargetCalculations {
         return round(Double(halfBasalTargetValue))
     }
 
-    /// Checks if the settings HBT would result in a percentage at or below the minimum,
-    /// and if so, returns an adjusted HBT that yields exactly the minimum percentage.
-    /// - Parameters:
-    ///   - settingHalfBasalTarget: The HBT from user settings
-    ///   - target: The target glucose value
-    ///   - autosensMax: The maximum autosens multiplier from settings
-    /// - Returns: A tuple containing (percentage, halfBasalTarget) where halfBasalTarget is nil if settings HBT is OK
-    static func computeStandardPercentageAndHBT(
-        settingHalfBasalTarget: Decimal,
-        target: Decimal,
-        autosensMax: Decimal
-    ) -> (percentage: Double, halfBasalTarget: Decimal?) {
-        let rawPercentage = computeRawPercentage(
-            halfBasalTarget: settingHalfBasalTarget,
-            target: target,
-            autosensMax: autosensMax
-        )
-
-        // If raw percentage is at or below minimum, we need to calculate an adjusted HBT
-        if rawPercentage <= minSensitivityRatioTT {
-            let adjustedHBT = Decimal(computeHalfBasalTarget(
-                target: target,
-                percentage: minSensitivityRatioTT
-            ))
-            return (minSensitivityRatioTT, adjustedHBT)
-        } else {
-            return (rawPercentage, nil)
-        }
-    }
-
     /// Determines the effective HBT to use for a TempTarget.
-    /// If the stored HBT is nil (standard TT) and using settings HBT would result in < 15%,
+    /// If the stored HBT is nil (standard TT) and using settings HBT would result in <= 15%,
     /// calculates an adjusted HBT. Otherwise returns the stored HBT or nil.
     /// - Parameters:
     ///   - tempTargetHalfBasalTarget: The HBT stored with the TempTarget (nil for standard TT)
@@ -114,13 +84,21 @@ enum TempTargetCalculations {
             return tempTargetHalfBasalTarget
         }
 
-        // For standard TT (no stored HBT), check if we need to adjust
-        let (_, adjustedHBT) = computeStandardPercentageAndHBT(
-            settingHalfBasalTarget: settingHalfBasalTarget,
+        // For standard TT (no stored HBT), check if settings HBT would result in <= minimum
+        let rawPercentage = computeRawPercentage(
+            halfBasalTarget: settingHalfBasalTarget,
             target: target,
             autosensMax: autosensMax
         )
 
-        return adjustedHBT
+        // If raw percentage is at or below minimum, calculate an adjusted HBT
+        if rawPercentage <= minSensitivityRatioTT {
+            return Decimal(computeHalfBasalTarget(
+                target: target,
+                percentage: minSensitivityRatioTT
+            ))
+        }
+
+        return nil
     }
 }
--- a/Trio/Sources/Helpers/TempTargetCalculations.swift
+++ b/Trio/Sources/Helpers/TempTargetCalculations.swift
@@ -104,20 +104,20 @@ enum TempTargetCalculations {
     /// If the stored HBT is nil (standard TT) and using settings HBT would result in < 15%,
     /// calculates an adjusted HBT. Otherwise returns the stored HBT or nil.
     /// - Parameters:
-    ///   - storedHBT: The HBT stored with the TempTarget (nil for standard TT)
+    ///   - tempTargetHalfBasalTarget: The HBT stored with the TempTarget (nil for standard TT)
     ///   - settingHalfBasalTarget: The HBT from user settings
     ///   - target: The target glucose value
     ///   - autosensMax: The maximum autosens multiplier from settings
     /// - Returns: The effective HBT to use, or nil if settings HBT should be used as-is
     static func computeEffectiveHBT(
-        storedHBT: Decimal?,
+        tempTargetHalfBasalTarget: Decimal?,
         settingHalfBasalTarget: Decimal,
         target: Decimal,
         autosensMax: Decimal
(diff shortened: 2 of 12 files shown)


mountrcg:
min HBT input in settings guard rails are set to 105. Currently 100 is allowed, which is also not correct and gives inplausible results if chosen.

mountrcg:
We need another general computation guardrail for low HBT in settings, which can leed to sensitivity <15%. Eg HBT set to 105 and then a TT is set at 195. Resulting sensRatio would be 5%, which is currently still set for standard and custom highTT's.
In those cases the used hbt needs to deviate from setting HBT in order to raise sensRatio to 15%. Seem complicated especially for a standard highTT that raises sens in this case.

kingst:
@mountrcg looks like this is still a draft, let me know when you're ready for a review and I can take a look.

mountrcg:
Honestly, I haven’t figured out how to change the HBT for a standard temp target, which would be a required for a settings HBT of 105 and the target value above 135. @marv-out  did these parts. Normaly a standard TT does not get a a custom HBT, but in order not to get below 15% some edge cases would requiere it - messy. May be we accept that as is?

dnzxy:
@mountrcg , the logic you are talking about can be found here https://github.com/nightscout/Trio/blob/dev/Trio/Sources/APS/OpenAPS/OpenAPS.swift#L536-L552

I'm not 100% certain that I understand the edge case you have found with a setting-HBT of 105, but this is what I understood:

Given a normal target and a half-basal target, we would normally compute the adjustment sensitivity and based on this new sensitivity, manipulate the setting-HBT with a custom-HBT. But if the computed sensitivity would be < 15%, then the sensitivity must be capped at 15%, and in that capped case, we want to "recompute" the HBT value for that same target, such that the resulting sensitivity is exactly 15%.

If that's correct, we could do something like the following:
1. Add a new helper function.
2. Recompute the HBT.
3. Check if the recomputed HBT differs from the temp target HBT, if so, adjust HBT _again_.
4. Send the twice-adjusted HBT to oref, where the actual sensitivity adjustment takes place.

Add this to `OpenAPS.swift`
```swift
private func computeCappedHalfBasalTarget(
    normalTarget: Decimal,
    tempTarget: Decimal,
    originalHBT: Decimal
) -> Double {
    // Step 1: compute P from existing HBT
    // P = (HBT - Normal) / (HBT - Temp + Normal)
    let numerator = originalHBT - normalTarget
    let denominator = originalHBT - tempTarget + normalTarget
    
    guard denominator != 0 else {
        return round(Double(originalHBT))  // just be safe
    }

    let computedRatio = numerator / denominator
    let computedPercent = Double(truncating: computedRatio as NSNumber) * 100
    
    // Step 2: check for <15%
    if computedPercent >= 15 {
        return round(Double(originalHBT))
    }

    // Step 3: cap at 15% and recompute HBT
    let cappedRatio = Decimal(0.15)

    let cappedHBT =
        ((2 * cappedRatio * normalTarget) - normalTarget - (cappedRatio * tempTarget)) /
        (cappedRatio - 1)

    return round(Double(cappedHBT))
}
```

The computation logic was taken from `AdjustmentsStateModel%2BExtensions/AdjustmentsStateModel%2BTempTargets.swift`, reversed and adds capping.

Below can be found in `OpenAPS.swift`.
```swift
// Check if a Temp Target is active and if its HBT differs from user preferences
if let activeTempTarget = try self.fetchActiveTempTargets().first,
   activeTempTarget.enabled,
   let activeHBT = activeTempTarget.halfBasalTarget?.decimalValue,
   activeHBT != defaultHalfBasalTarget
{
   const recomputedHBT = self.computeCappedHalfBasalTarget(
            normalTarget: normalTarget, // grab via profile
            tempTarget: activeTempTarget.target,
            halfBasalTarget: activeHBT
        )

    // Overwrite the HBT in preferences
    adjustedPreferences.halfBasalExerciseTarget = activeHBT != recomputedHBT ? recomputedHBT : activeHBT
    debug(.openAPS, "Updated halfBasalExerciseTarget to active Temp Target value: \(activeHBT)")
}
```

This is rough code, just written down, while looking at the few files that are involved here. Not yet tested, just roughly outlined.

mountrcg:
> I'm not 100% certain that I understand the edge case you have found with a setting-HBT of 105, but this is what I understood:
> 
> Given a normal target and a half-basal target, we would normally compute the adjustment sensitivity and based on this new sensitivity, manipulate the setting-HBT with a custom-HBT. But if the computed sensitivity would be < 15%, then the sensitivity must be capped at 15%, and in that capped case, we want to "recompute" the HBT value for that same target, such that the resulting sensitivity is exactly 15%.

Not quite. We have 2 kinds of TT with adjusted sensitivity.
1. standard
2. customized

For (2) we do not calculate adjustedSensitivity in this PR, as per @kingst suggestion we always take that value as input and calculate the HBT that will realize this. So problem solved as the input is guarded.

For (1) we take the target value and the HBT from settings and calculate adjusted sensitivity. So here is the issue
a. There is no check against the 15%. 
b. If there would be, we then need to calculate a new HBT to use instead of the one from settings. Also fine and now in PR, but as far as I know the standard TT does not carry over a custom HBT to oref, only if TT is "Custom".

So when I test the the standard TT with that low HBT all looks fine in the Add/Eddit TT but oref will calculate the 11% sens adjustment.

Simulator Screenshot - Dev-16Pro - 2025-12-08 at 18 13 20
Simulator Screenshot - Dev-16Pro - 2025-12-08 at 18 13 02

dnzxy:
@mountrcg I‘m aware how it works. My outlining addressed both cases. Have a read through the code.

mountrcg:
here are the 2 functions that calc

1. adjustedSens https://github.com/nightscout/Trio/blob/3328da60515af9f3c06b5599ecacfb19fec27980/Trio/Sources/Modules/Adjustments/AdjustmentsStateModel%2BExtensions/AdjustmentsStateModel%2BTempTargets.swift#L429
2 adjustedHBT https://github.com/nightscout/Trio/blob/3328da60515af9f3c06b5599ecacfb19fec27980/Trio/Sources/Modules/Adjustments/AdjustmentsStateModel%2BExtensions/AdjustmentsStateModel%2BTempTargets.swift#L388

So I move those to helpers and use in OpenAPS?

dnzxy:
My suggestion was to add a new computation function as a helper util for OpenAPS. You could probably use those two functions and add some logic around them in OpenAPS as a different way to solving this, either or :)

mountrcg:
I have reduced the PR and only added the guarding of min Insulin% of 15% in OpenAPS and the adjusted Insulin% in HomeView. Also moved the HBT and Sens calculations to a helper file to use in:

- Add / Edit TT Forms and the States
- OpenAPS
- HomeView

Ready for testing and Review @dnzxy & @kingst

mountrcg:
Test cases:
Try to get Insulin% substantially below 15% by 

1. changing the target values of new or edited high TT's (custom or standard)
2. setting the HBT in settings to it's minimum of 105 and adding new high TT's > 140, or testing existing TT's > 140

Trio/Sources/APS/OpenAPS/OpenAPS.swift
```diff
@@ -534,15 +534,26 @@ final class OpenAPS {
 
         // Check for active Temp Targets and adjust HBT if necessary
         try await context.perform {
-            // Check if a Temp Target is active and if its HBT differs from user preferences
+            // Check if a Temp Target is active
```
dnzxy: We do still check if the HBT differs (just differently now) and we technically still adjust it here, if need be.

Trio/Sources/APS/OpenAPS/OpenAPS.swift
```diff
@@ -534,15 +534,26 @@ final class OpenAPS {
...
-               let activeHBT = activeTempTarget.halfBasalTarget?.decimalValue,
-               activeHBT != defaultHalfBasalTarget
+               let targetValue = activeTempTarget.target?.decimalValue
             {
-                // Overwrite the HBT in preferences
-                adjustedPreferences.halfBasalExerciseTarget = activeHBT
-                debug(.openAPS, "Updated halfBasalExerciseTarget to active Temp Target value: \(activeHBT)")
+                // Compute effective HBT - handles both custom HBT and standard TT (where HBT might need adjustment)
+                let effectiveHBT = TempTargetCalculations.computeEffectiveHBT(
+                    storedHBT: activeTempTarget.halfBasalTarget?.decimalValue,
```
dnzxy: Please rename `storedHalfBasalTarget` to `tempTargetHalfBasalTarget` for all instances + comments, because _stored_ could be misread as the actual setting-stored HBT.

Trio/Sources/Helpers/TempTargetCalculations.swift
```diff
@@ -0,0 +1,132 @@
...
+/// Helper functions for TempTarget sensitivity calculations.
+/// These are used across the app (UI, OpenAPS) to ensure consistent behavior.
+enum TempTargetCalculations {
+    /// The minimum allowed sensitivity ratio for TempTargets (15%)
+    static let minSensitivityRatioTT: Double = 15
+
+    /// The normal target glucose value used as reference (100 mg/dL)
+    static let normalTarget: Decimal = 100
+
+    /// Computes the raw (unclamped) adjusted percentage for a given HBT and target.
```
dnzxy: Do we gain anything by having these two functions present separately and not just always having the low-end clamp present? This guard / checking condition could be part of `computeRawAdjustedPercentage`, could it not?

Trio/Sources/Helpers/TempTargetCalculations.swift
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /// Checks if the settings HBT would result in a percentage at or below the minimum,
+    /// and if so, returns an adjusted HBT that yields exactly the minimum percentage.
+    /// - Parameters:
+    ///   - settingHalfBasalTarget: The HBT from user settings
+    ///   - target: The target glucose value
+    ///   - autosensMax: The maximum autosens multiplier from settings
+    /// - Returns: A tuple containing (percentage, halfBasalTarget) where halfBasalTarget is nil if settings HBT is OK
+    static func computeStandardPercentageAndHBT(
```
dnzxy: If we combine the two functions, this function here because much easier &/or even because a no longer needed wrapper-util-function, because instead of doing what this function here encapsulates, you might as well just add an invocation of those 2 function that are being called within this one.

Trio/Sources/Helpers/TempTargetCalculations.swift
```diff
@@ -0,0 +1,132 @@
...
+    /// Determines the effective HBT to use for a TempTarget.
+    /// If the stored HBT is nil (standard TT) and using settings HBT would result in < 15%,
+    /// calculates an adjusted HBT. Otherwise returns the stored HBT or nil.
+    /// - Parameters:
+    ///   - storedHBT: The HBT stored with the TempTarget (nil for standard TT)
+    ///   - settingHalfBasalTarget: The HBT from user settings
+    ///   - target: The target glucose value
+    ///   - autosensMax: The maximum autosens multiplier from settings
+    /// - Returns: The effective HBT to use, or nil if settings HBT should be used as-is
+    static func computeEffectiveHBT(
```
dnzxy, line 76: Again, a wrapper, for a wrapper, that just encapsulates a lot of stuff.
Do we need 1-2 layers of abstraction for something as simple as getting a % value and a target value?

Trio/Sources/Models/DecimalPickerSettings.swift
```diff
@@ -72,7 +72,7 @@ struct DecimalPickerSettings {
     var halfBasalExerciseTarget = PickerSetting(
         value: 160,
         step: 5,
-        min: 100,
+        min: 105,
```
dnzxy: Let's please add a short inline-comment here referencing/explaining, that 105 is a hard-coded oref limit and shall not be adjusted (because it will/can lead to inconsistent behavior when calculating TTs).

Trio/Sources/Modules/Adjustments/AdjustmentsStateModel+Extensions/AdjustmentsStateModel+TempTargets.swift
```diff
@@ -421,24 +418,74 @@ extension Adjustments.StateModel {
...
         guard calcTarget != 0
         else { return Double(autosensMax * 100) } // oref defined limit for increased insulin delivery
-        let maxSens = calcTarget > normalTarget ? 95 : Double(autosensMax * 100)
+        let maxSens = calcTarget > TempTargetCalculations.normalTarget ? 95 : Double(autosensMax * 100)
         return maxSens
     }
 
+    /// Computes the raw (unclamped) adjusted percentage for a given HBT and target.
+    /// This is used to check if the percentage would be below minSensitivityRatioTT.
+    func computeRawAdjustedPercentage(
```
dnzxy: Why are you abstracting this by this wrapper instead of just directly calling `TempTargetCalculations.computeRawAdjustedPercentage(…)`? This seems unnecessary.

Trio/Sources/Modules/Adjustments/AdjustmentsStateModel+Extensions/AdjustmentsStateModel+TempTargets.swift
```diff
@@ -421,24 +418,74 @@ extension Adjustments.StateModel {
...
+    ) -> Double {
+        TempTargetCalculations.computeRawAdjustedPercentage(
+            halfBasalTarget: halfBasalTargetValue,
+            target: calcTarget,
+            autosensMax: autosensMax
+        )
+    }
+
     /// Computes the adjusted percentage for the slider.
     func computeAdjustedPercentage(
```
dnzxy: Same question here? Why abstract this instead of just directly calling TempTargetCalculations util?

Trio/Sources/Modules/Adjustments/AdjustmentsStateModel+Extensions/AdjustmentsStateModel+TempTargets.swift
```diff
@@ -421,24 +418,74 @@ extension Adjustments.StateModel {
...
+            autosensMax: autosensMax
+        )
+    }
+
+    /// Computes the standard percentage and adjusted HBT for a given target.
+    /// If the raw percentage (using settingHalfBasalTarget) would be at or below minSensitivityRatioTT (15%),
+    /// returns an adjusted HBT that yields the minimum percentage instead.
+    /// - Parameter target: The target glucose value
+    /// - Returns: A tuple containing (percentage, halfBasalTarget) where halfBasalTarget is nil if standard HBT can be used
+    func computeStandardPercentageAndHBT(usingTarget target: Decimal) -> (percentage: Double, halfBasalTarget: Decimal?) {
```
dnzxy: Same question here? Either, just call the function in `TempTargetCalculations` that does _exactly this_ and returns a tuple of percentage and HBT value, or (if you are removing that layer of abstraction, because I asked for it in my comments in `TempTargetCalculations`), why not call the 2 util functions here directly? Why abstract this instead of just directly calling it? Do we gain anything by having a tuple instead of two "unconnected" values?

dnzxy (commented):
Reviewed it and ran it briefly. It seems okay functionality wise, but I did have a lot of questions. The coding choices done here seem to add a lot of complexity via introducing abstraction layers / encapsulation through wrapper functions, where your initial aim (as per @kingst 's comments on #716) was to massively simplify all this logic. 

Please don't understand this as criticism, just trying to get to the easiest result, hence my questions :) 

Thanks for tackling this and looking into fixing this edge case.

PS: Maybe we should a add a semi-long + descriptive docstring _somewhere_ explaining why 
- 15% is a hard-coded minimum limit for sensitivity adjustments
- why ASmin and ASmax are not used the same way when clamping TT sensitivity adjustments
- where this stems from (i.e., oref / AndroidAPS)

Trio/Sources/APS/OpenAPS/OpenAPS.swift
mountrcg: done

Trio/Sources/APS/OpenAPS/OpenAPS.swift
mountrcg: done

Trio/Sources/Modules/Adjustments/AdjustmentsStateModel+Extensions/AdjustmentsStateModel+TempTargets.swift
mountrcg: done

Trio/Sources/Modules/Adjustments/AdjustmentsStateModel+Extensions/AdjustmentsStateModel+TempTargets.swift
mountrcg: done

Trio/Sources/Models/DecimalPickerSettings.swift
mountrcg: done

Trio/Sources/Helpers/TempTargetCalculations.swift
mountrcg: done

mountrcg:
tackled all the valueble comments and reduced the logging now

Trio/Sources/Modules/Adjustments/AdjustmentsStateModel+Extensions/AdjustmentsStateModel+TempTargets.swift
```diff
@@ -138,6 +138,11 @@ extension Adjustments.StateModel {
         let date = self.date
         guard date > Date() else { return }
 
+        let adjustmentType = halfBasalTarget == settingHalfBasalTarget ? "Standard" : "Custom"
```
dnzxy, line 141: Does a (static) variable for this helper variable make sense vs. adding the same snippet 3 times?

Trio/Sources/Modules/Adjustments/View/TempTargets/TempTargetHelpView.swift
```diff
@@ -7,20 +7,45 @@ struct TempTargetHelpView: View {
...
+                            "A pre-condition to have Temp Targets adjust Sensitivity is that the respective Target Behavior settings High Temp Target Raises Sensitivity or Low Temp Target Lowers Sensitivity are set to enabled."
+                        )
+                    }
+                } header: {
+                    Text("Overview")
+                }
+                .listRowBackground(Color.gray.opacity(0.1))
+
+                Section {
+                    VStack(alignment: .leading, spacing: 10) {
```
dnzxy, line 31: Following text suggestions (via GPT, adjusted for high school student literacy).
Feel free to use/adjust.

```swift
VStack(alignment: .leading, spacing: 10) {
    Text(
        "Sensitivity changes from Temp Targets have a built-in minimum of 15%. Even very high Temp Targets cannot reduce insulin delivery below 15% of normal."
    )
    Text(
        "This 15% minimum is a safety limit taken from oref (OpenAPS reference design) and AndroidAPS. It helps prevent insulin delivery from dropping to unsafe levels."
    )
    Text(
        "Important: Autosens Min and Autosens Max do not affect Temp Targets in the same way. Autosens Max limits how much insulin can be increased, but Autosens Min does not remove the 15% minimum when insulin is reduced."
    )
    Text(
        "This difference exists because situations like exercise often need a much larger insulin reduction than Autosens would detect during a normal daily routine."
    )
}
```

dnzxy (approved):
Two small comments, but overall I'm okay with this.
@kingst please give this a review. I think we're at a good point and have simplified things.

kingst (approved):
LGTM

Trio/Sources/Modules/Adjustments/AdjustmentsStateModel+Extensions/AdjustmentsStateModel+TempTargets.swift
mountrcg: I would like to leave it, used it often searching in code

dnzxy (approved):

PR #880, opened 2025-12-10
fix: LoopKit.startDateToFilterNewData

bastiaanv:
While developing the EversenseKit, I found an issue with the current `startDateToFilterNewData` method in PluginSource. It always returns nil, since the task is performed on a different thread. While no CGMManager use this method, I do believe we need to fix this just to be sure.

This implementation was made by @dnzxy and I am running this code with no problems via EversenseKit.
Loop's implementation: https://github.com/LoopKit/Loop/blob/dev/Loop/Managers/DeviceDataManager.swift#L1013

The other deleted files are unused and not registered in Xcode


dnzxy (approved):

dnzxy:
I’d go ahead and merge this unless we have any concerns around adding this. @kingst ?

Trio/Sources/APS/CGM/PluginSource.swift
```diff
@@ -144,17 +144,10 @@ extension PluginSource: CGMManagerDelegate {
...
-            guard let self = self else { return }
-
-            dispatchPrecondition(condition: .onQueue(self.processQueue))
-
-            date = glucoseStorage.lastGlucoseDate()
-        }
-
-        return date
+        let last = glucoseStorage.lastGlucoseDate()
+        return last == .distantPast ? nil : last
```
kingst: Nit: I don't like checking against `.distantPast` should we have a different interface for lastGlucoseData that returns `nil` if there isn't any data?

kingst (approved):
One small and optional change, but generally LGTM.

It might be worth checking to make sure that backfills still work, that logic was pretty wonky (but it sounds like from the comment it isn't getting used anyway).

Trio/Sources/APS/CGM/PluginSource.swift
dnzxy: That’s what Loop does, to be fair. It‘s just that expect this to return something right now so if we make this return nil we just need to handle that elsewhere.

dnzxy: That’s what Loop does, to be fair. It‘s just that expect this to return something right now so if we make this return nil we just need to handle that elsewhere.

bastiaanv: We could update the lastGlucoseDate to the following:

```swift
    func lastGlucoseDate() -> Date? {
        let fr = GlucoseStored.fetchRequest()
        fr.predicate = NSPredicate.predicateForOneDayAgo
        fr.sortDescriptors = [NSSortDescriptor(keyPath: \GlucoseStored.date, ascending: false)]
        fr.fetchLimit = 1

        var date: Date?
        context.performAndWait {
            do {
                let results = try self.context.fetch(fr)
                date = results.first?.date
            } catch let error as NSError {
                debug(.storage, "Fetch error: \(DebuggingIdentifiers.failed) \(error), \(error.userInfo)")
            }
        }

        return date
    }

    func isGlucoseFresh() -> Bool {
        Date().timeIntervalSince(lastGlucoseDate() ?? .distantPast) <= Config.filterTime
    }
```

Making it nullable, and move the distantPast to the `isGlucoseFresh` function

dnzxy: I like this a lot. Would you 👍 this,
@kingst ?

dnzxy: While we are at it, let’s please do `fr` -> `fetchRequest` 👀😅

bastiaanv: Done! @kingst you agree with the latest commit?

bastiaanv:
@dnzxy In the latest dev, this unit test broke

Screenshot 2025-12-20 at 21 12 31

dnzxy:
> @dnzxy In the latest dev, this unit test broke
> 
> Screenshot 2025-12-20 at 21 12 31

Is your local sim set up with mmol/L?

bastiaanv:
That fixed the issue. Thanks 😂😅

Trio/Sources/APS/CGM/PluginSource.swift
dnzxy: Is this good to go from your side, @kingst ?

kingst: LGTM, thanks @bastiaanv

kingst (approved):

PR #885, opened 2025-12-13
Therapy profiles

zgoettsc:
**Introducing Therapy Profiles**

-Allowing users to create multiple therapy setting profiles and automatically switch between them based on the day of week. Ideal for users who need different settings on weekdays vs weekends or other recurring schedules.
Features

**Profile Management**

- Create up to 10 independent therapy profiles
- Each profile contains its own:
- Basal rates
- Insulin sensitivities (ISF)
- Carb ratios
- Glucose targets
- Duplicate existing profiles as a starting point
- Copy settings from another profile or current active pump settings

**Day-of-Week Scheduling**

- Assign profiles to specific days (e.g., "Weekend" profile for Sat/Sun)
- Each day can only belong to one profile
- Conflict detection with reassignment confirmation dialog
- Quick-select buttons for Weekdays, Weekends, All, None

**Automatic Profile Switching**

- App checks for scheduled profile changes at each loop cycle
- Automatic switch at app startup based on current day
- Switch history tracking for troubleshooting
- Banner notification when profile switches occur

**Manual Override**

- Long-press any profile to "Override for Today"
- Temporarily activates selected profile until midnight
- Automatically reverts to scheduled profile the next day

**Seamless Integration**

- When a profile activates, its settings are written to the standard OpenAPS settings files
- Full compatibility with existing algorithm - no changes to loop logic
- Existing users automatically migrated: current settings become "Default" profile

**Inline Editing**

- Edit all therapy settings directly within the profile editor
- Expandable sections for basal rates, ISF, carb ratios, and targets
- Add, edit, and delete time slots without navigating away

IMG_3500
IMG_3501
IMG_3502
IMG_3503

### DIFF
--- a/Documents/TherapyProfiles_Implementation.md
+++ b/Documents/TherapyProfiles_Implementation.md
@@ -1,0 +1,279 @@
+# Therapy Profiles Implementation Plan
+
+## Overview
+
+Implement weekday/weekend therapy profiles for Trio, allowing users to create multiple named therapy profiles with different settings that automatically switch based on the day of week.
+
+## Requirements
+
+1. **Multiple Profiles**: Up to 10 named therapy profiles
+2. **Complete Settings Per Profile**:
+   - Basal rates
+   - Insulin Sensitivity Factors (ISF)
+   - Carb Ratios
+   - Glucose Targets
+3. **Day Assignment**: User-selectable days of the week per profile
+4. **Automatic Switching**: Profile switches at midnight based on day of week
+5. **Manual Override**: Ability to manually activate a different profile for the current day
+6. **Notifications**: In-app banner when profile switches (not push notifications)
+7. **History Logging**: Log all profile switches with timestamps and reasons
+8. **Nightscout Sync**: Sync profile changes to Nightscout
+9. **No Watch App**: Watch integration not required at this time
+
+## Architecture
+
+### Directory Structure (Trio v0.6.0)
+
+```
+Trio/Sources/
+├── Models/
+│   ├── Weekday.swift                    # Days of week enum
+│   ├── TherapyProfile.swift             # Profile model with all settings
+│   └── ProfileSwitchEvent.swift         # Switch history event
+├── Services/
+│   └── ProfileManager/
+│       ├── ProfileManager.swift         # Protocol
+│       └── BaseProfileManager.swift     # Implementation
+├── Modules/
+│   ├── TherapyProfileList/
+│   │   ├── TherapyProfileListDataFlow.swift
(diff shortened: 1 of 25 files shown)


mountrcg:
Great things, that have been missing so far! What I would be looking for in profiles is to adjust main therapy settings by a given percentage to create a new profile.
Also I would like to have all the other therapy parameters be adjustable for a profile e.g. have a different dynamic ISF factor etc. etc.

zgoettsc:
How would you do the sliders? Choose a base profile and then at each time
point in the profile (carbs, isf, targets, etc) have a slider to move up or
down? Still have the ability to remove and add time points?

The other settings being in the profile makes a lot of sense!

Zack

kristinkand:
I really like it! That would be neat to have. Do you think different SMB/UAM settings would be possible too?

dnzxy:
Before this gets out of hand here - we are already in talks and working on a feature such as this one. 

Please reach out to the developing team on Discord, Zach.

dnzxy:
Needless to say, but I really think it needs to be added here: this is a largely AI generated set of code. There‘s no provided testing results or unit tests, it’s unclear whether it has been tested or how, and **it goes without saying this should not be run on an in-vivo device by anyone** at this point in time. 

This PR aims to change metric sets that directly influence subsequent dosing and have side effects to Dynamic ISF that the current state of the PR does not consider, and, as mentioned earlier, we are designing and already discussing this.

If you want to contribute in a meaningful way to Trio, please reach out directly to the team :)

zgoettsc:
Thank you for the reply. My only goal was to develop features to make my
app more usable and present those ideas to the developers for consideration
in the Trio app so others can use similar features if they find them
helpful. It seems maybe this isn’t the correct way to do that. I am new to
GitHub and PRs and how they work. I will refrain from posting here.

dnzxy:
Please reach out to the team on Discord so we can involve you in the ongoing discussion for Therapy Profiles, @zgoettsc  :)

dnzxy:
I‘ll go ahead and close this for now, we can revisit this PR (and possibly reopen it) after a conversation around its implications to dosing and the established UI/UX in Trio and its implementation details. 

Thanks again for putting this forward @zgoettsc . Please get in contact with the team directly in Discord :) You can join it easily via the handy link https://discord.triodocs.org

Please don’t feel discouraged by a more cautious approach by the team to AI generated code and massive PRs being posted here before any prior conversation about implementation details has happened. We’re looking forward to hearing from you. Get in contact, we don’t bite 😅

Thanks, take care :) 
Dan

Issue #375, opened 2024-08-07
Cannot silence Omnipod low reservoir alarm on first pod


g5WS:
## Describe the bug
I'm on my first pod under Trio and have just passed the 10-unit-remaining mark.  The pod is beeping intermittently, and I can't seem to silence it.  I received no app pop-up or lock screen notification.  I am running iOS18 (latest developer build)

This area of functionality does not seem to be described in the help files: https://docs.diy-trio.org/en/latest/search.html?q=%22Low+reservoir%22&check_keywords=yes&area=default

## Attach a Log
*That's a lot of personal information... but here are some excerpts, and I'm happy to search and excerpt more if someone provides strings to search.  Here are the first 3 log entries when the reservoir hits 10, and when it first goes below.

- 2024-08-07T10:59:42-0700 [DeviceManager] DeviceDataManager.swift - pumpManager(_:didReadReservoirValue:at:completion:) - 484 - DEV: Reservoir Value 10.0, at: 2024-08-07 17:59:42 +0000
2024-08-07T10:59:53-0700 [DeviceManager] DeviceDataManager.swift - deviceManager(_:logEventForDeviceIdentifier:type:message:completion:) - 562 - DEV: Device message: 17364df508030e01000073
2024-08-07T10:59:53-0700 [DeviceManager] DeviceDataManager.swift - deviceManager(_:logEventForDeviceIdentifier:type:message:completion:) - 562 - DEV: Device message: 17364df50c0a1d290551500008358cc80281

---
2024-08-07T11:04:24-0700 [DeviceManager] DeviceDataManager.swift - pumpManager(_:didReadReservoirValue:at:completion:) - 484 - DEV: Reservoir Value 9.85, at: 2024-08-07 18:04:24 +0000
2024-08-07T11:05:08-0700 [DeviceManager] PluginSource.swift - deviceManager(_:logEventForDeviceIdentifier:type:message:completion:) - 86 - DEV: device Manager for Optional("83QR80") : Connected
2024-08-07T11:05:09-0700 [DeviceManager] PluginSource.swift - deviceManager(_:logEventForDeviceIdentifier:type:message:completion:) - 86 - DEV: device Manager for Optional("83QR80") : New reading: 2024-08-07 18:04:55 +0000*

## To Reproduce
Unknown reproducibility

## Expected behavior
Acceptable: in-app notification that allows reminder to be silenced
Ideally: lock screen notification that allows reminder to be silenced and in-app notification that allows reminder to be silenced.
[...]

PR #893, opened 2025-12-20
Fix First Pod Alarm + Refactor Alert History Storage

dnzxy:
This PR fixes an edge case where the first stored alert could be dropped, and modernizes alert history persistence by moving it from file-based JSON storage to `UserDefaults`.

#### What changed

* **Fix: initial alert storage (#375)**
  Ensures the first stored alert is kept when the history file is empty (fallback now includes the newly added alert).

* **Refactor: AlertHistoryStorage now persists via `UserDefaults`**
  Alert history is stored as encoded `[AlertEntry]` in `UserDefaults` (using existing `JSONCoding` encoder/decoder for consistency).

  * Keeps existing behavior: **dedupe by `issuedDate`**, **retain last 24h**, **newest-first sorting**.
  * Thread safety preserved via existing `processQueue`.

* **Migration logic (one-time)**
  On first initialization, if `UserDefaults` has no alert history but legacy `monitor/alerthistory.json` exists:

  * migrate + normalize alert history into `UserDefaults`
  * set a migration flag
  * remove the legacy file to prevent drift
    (`FileStorage` is kept temporarily for migration only; marked with FIXME to remove later.)

* **API + naming improvements + docstrings**

  * Renamed for clarity: `storeAlert` → `addAlert`, `recentNotAck` → `unacknowledgedAlertsWithinLast24Hours`, `forceNotification` → `broadcastAlertUpdates`, etc.
  * `alertNotAck` publisher renamed to `unacknowledgedAlertsPublisher`.
  * Updated call sites in `DeviceDataManager` + PumpConfig flow.
  * Added docstrings to clarify behavior and intent.

* **Minor linting cleanup** in `InsulinSensitivityStepView`.

#### Side Note

* Migration runs once per install and is safe to re-enter (guarded by flag + presence of UserDefaults payload).

### DIFF
--- a/Trio/Sources/APS/DeviceDataManager.swift
+++ b/Trio/Sources/APS/DeviceDataManager.swift
@@ -596,7 +596,7 @@ extension BaseDeviceDataManager: PumpManagerDelegate {
 
 extension BaseDeviceDataManager: DeviceManagerDelegate {
     func issueAlert(_ alert: Alert) {
-        alertHistoryStorage.storeAlert(
+        alertHistoryStorage.addAlert(
             AlertEntry(
                 alertIdentifier: alert.identifier.alertIdentifier,
                 primitiveInterruptionLevel: alert.interruptionLevel.storedValue as? Decimal,
@@ -611,7 +611,7 @@ extension BaseDeviceDataManager: DeviceManagerDelegate {
     }
 
     func retractAlert(identifier: Alert.Identifier) {
-        alertHistoryStorage.deleteAlert(identifier: identifier.alertIdentifier)
+        alertHistoryStorage.removeAlert(identifier: identifier.alertIdentifier)
     }
 
     func doesIssuedAlertExist(identifier _: Alert.Identifier, completion _: @escaping (Result<Bool, Error>) -> Void) {
@@ -685,23 +685,23 @@ extension BaseDeviceDataManager: AlertObserver {
             if let omnipodBLE = self.pumpManager as? OmniBLEPumpManager {
                 if omnipodBLE.state.activeAlerts.isEmpty {
                     // force to ack alert in the alertStorage
-                    self.alertHistoryStorage.ackAlert(alertIssueDate, nil)
+                    self.alertHistoryStorage.acknowledgeAlert(alertIssueDate, nil)
                 }
             }
 
             if let omniPod = self.pumpManager as? OmnipodPumpManager {
                 if omniPod.state.activeAlerts.isEmpty {
                     // force to ack alert in the alertStorage
-                    self.alertHistoryStorage.ackAlert(alertIssueDate, nil)
+                    self.alertHistoryStorage.acknowledgeAlert(alertIssueDate, nil)
                 }
             }
 
             self.pumpManager?.acknowledgeAlert(alertIdentifier: alert.alertIdentifier) { error in
                 if let error = error {
-                    self.alertHistoryStorage.ackAlert(alertIssueDate, error.localizedDescription)
+                    self.alertHistoryStorage.acknowledgeAlert(alertIssueDate, error.localizedDescription)
                     debug(.deviceManager, "acknowledge not succeeded with error \(error)")
                 } else {
-                    self.alertHistoryStorage.ackAlert(alertIssueDate, nil)
+                    self.alertHistoryStorage.acknowledgeAlert(alertIssueDate, nil)
                 }
             }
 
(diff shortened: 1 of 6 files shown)


marionbarker:
## Test

Summary: This fixes the first alert issue, #375.

However, it also modifies alert storage method. That storage update portion of the PR is not yet tested.

### Test Method

Thanks to @itsmojo for his detailed notes - summarized below:

Method to test if Trio continued beeping on initial pod bug will occur using the rPi pod simulator and Pod Simulator Front end.

1. Delete the Trio app and install fresh, onboard accepting defaults - no need for Nightscout
2. Select Omnipod DASH pump and pair with rPi DASH simulator
3. On Trio, go to Omnipod DASH menu, scroll down to “Pod Diagnostics” and tap “Read Pod Status” and stay on this screen
4. Make the following change to the Pod Simulator Front end (actual value of the reservoir is irrelevant)
   * Select “Low Reservoir(4)” alert
5. In the Trio app, tap on “Read Pod Status” and observe Alerts value. 
   * ✅ If it shows the pop-up alert for “Low reservoir”, then acknowledge it and tap “Read Pod Status” see if Alert clears. 
      * If alert was cleared, you will see “No alerts” 
   * ❌ If app doesn’t display a “Low Reservoir” pop up, then the Trio first beep never clears issue was reproduced
      * The pod status screen will continue to show Low Reservoir
      * Send a different alert such as Expiration Reminder
      * You will see the Trio app pop-up for Expiration Reminder, once you acknowledge that, the pod status will change from Low Reservoir, Expiration Reminder to just Low Reservoir

### Reproduce the error

Use SE 2nd gen phone running iOS 18.7.2
* Delete Trio app
* Build Trio `dev`  0.6.0.29
* Follow test plan above
* ❌ Step 5 did not show the "Low Reservoir" pop up, alert was not cleared, next alert popped up and was cleared

### Test the fix

Use SE 2nd gen phone running iOS 18.7.2
* Deactivate the pod and delete Trio app
* Build `fix-first-pod-alarm`, commit b5634bfa0
   * Because I am building with Xcode 26.2, need to close Xcode, `git reset --hard` to get rid of linting changes with Xcode 26.2 so I can `git switch fix-first-pod-alarm`
* Follow test plan above
* ✅  Step 5 did show the "Low Reservoir" pop up, alert was cleared

### Test transition of alert storage

The `dev` branch uses json for storage, this PR transitions from legacy on-disk JSON file into UserDefaults

Testing for how this migration works is saved for the next comment.

marionbarker:
## Test Migrations

✅ The Migration tests were successful.

### Test Details

The phone configuration is continued from the previous test and currently uses UserDefaults for alert storage.
* build over the existing app with dev branch, which uses JSON file
* I expect the first alert to be lost again and this was confirmed
   * ❌ Step 5 did not show the "Low Reservoir" pop up, alert was not cleared, next alert popped up and was cleared

Build `fix-first-pod-alarm` over `dev`: I expect the next alert to be properly handled.
* build over the existing app with  `fix-first-pod-alarm` branch, which should transition JSON file to UserDefault storage
* ✅ Step 5 initially showed the "Low Reservoir" alert from the `dev` build
   * Set another alert, Expiration Reminder and both alerts appeared on the Read Pod Status screen
   * Both alerts popped up in Trio and once acknowledged, all alerts were cleared

marionbarker (approved):
Approved from test. I need someone else to do the code review.

kingst (commented):
The code LGTM, very nice refactor. My only question is around unacknowledgedAlertsWithinLast24Hours. This method is both a public method on the protocol and used internally. When it's called externally should it use the dispatch queue? For the internal invocations what you have looks right, but if it's called externally there might be a race condition.

dnzxy:
Oh boy, very good catch. I just realized we have always had this potential for race conditions present (also for the former `recentNotAck`). I’ll serialize `unacknowledgedAlertsWithinLast24Hours()` using a re-entrant processQueue sync (via a dispatch-specific key) to avoid races with writes. That should prevent deadlocks during internal calls, and allow external calls to be thread-safe.

marionbarker:
Update my local clone that has both PR #893 and #894 merged together.
Build on to the test phone currently running a DASH pod that was paired after clearing Trio off the phone and building fresh.

dnzxy:
After some private discussion with @kingst , simplifying the thread safety with another small code change. I wrapped `unacknowledgedAlertsWithinLast24Hours()` in `processQueue.sync` to serialize external reads with writes. Internal call sites use a private `…OnQueue()` helper to avoid nested queue sync/deadlocks/race conditions.

kingst (approved):
LGTM

Issue #574, opened 2025-05-24
Autosense not working after onboarding v0.5, perhaps up to the 7 day limit for dynamic start


kab2000:
Expectation is that Autosense should be working ~~immediately~~ after 24 hrs. I am on day 6

* Hardware: iPhone 16 pro
* iOS Version: 18.5

### 

![Image](https://github.com/user-attachments/assets/52780bfd-6893-4ea1-bced-b5ac4aa96543)
[log-1.txt](https://github.com/user-attachments/files/20428094/log-1.txt)

Pump:
* Manufacturer: Insulet
* Model: Omnipod Dash

### CGM:
* Device: Dexcom G6
* Manager app: Dexcom App

### Trio Version:
* Version Number: 0.5.15

kingst:
Hi @kab2000 thanks for filing the report. I took a look at your logs and I can see some runs where autosens is making adjustments, but I can also see a number of runs where the algorithm is choosing to not make any adjustments.

Just to clarify, do the logs cover the period of time when you were expecting to see autosens but were not? For example, I can see an adjustment at 5:52am East Coast time today.

yeagerm:
> Hi @kab2000 
> thanks for creating this issue. I presume you are either Michael or TI from Facebook? Is that right? 
> Could you please also attach the `preferences.json` file of your Trio build. You ca find it under iOS Files > Trio. 
> Thank you! 

{
  "sigmoid" : false,
  "suspend_zeros_iob" : true,
  "enableSMB_with_temptarget" : false,
  "useNewFormula" : false,
  "maxSMBBasalMinutes" : 90,
  "high_temptarget_raises_sensitivity" : false,
  "smb_delivery_ratio" : 0.5,
  "enableSMB_high_bg" : false,
  "rewind_resets_autosens" : true,
  "low_temptarget_lowers_sensitivity" : false,
  "skip_neutral_temps" : false,
  "unsuspend_if_no_temp" : false,
  "maxDelta_bg_threshold" : 0.2,
  "updateInterval" : 20,
  "useWeightedAverage" : false,
  "weightPercentage" : 0.35,
  "remainingCarbsCap" : 90,
  "maxCOB" : 120,
  "allowSMB_with_high_temptarget" : false,
  "half_basal_exercise_target" : 160,
  "current_basal_safety_multiplier" : 4,
  "autosens_min" : 0.7,
  "insulinPeakTime" : 75,
  "bolus_increment" : 0.05,
[...]

PR #894, opened 2025-12-20
Update autosens to fetch 1.5 days worth of pump records + some lint updates

kingst:
Currently autosens pulls 24 hours of glucose data and 24 hours of pump history. But the autosens algorithm looks for insulin activity across the entire 24 hours of glucose history, so we need to pull more pump history to make sure that the old glucose deviations have accurate insulin activity (and IoB).

I picked 1.5 days as the time span has this will give us the max dia settable via the UI (10 hours) + 2 hours to get any temp basal events that started before the oldest DIA but end after it.

I think that this PR will fix: https://github.com/nightscout/Trio/issues/574

I wrote up a more detailed issue to track the technical details: https://github.com/nightscout/Trio/issues/897

### DIFF
--- a/Model/Helper/PumpEvent+helper.swift
+++ b/Model/Helper/PumpEvent+helper.swift
@@ -79,8 +79,15 @@ public extension PumpEventStored {
 }
 
 extension NSPredicate {
+    /// Used to fetch the last 24 hours of pump history
     static var pumpHistoryLast1440Minutes: NSPredicate {
         let date = Date.oneDayAgoInMinutes
+        return NSPredicate(format: "timestamp >= %@", date as NSDate)
+    }
+
+    /// Used to fetch the last 1.5 days of pump history
+    static var pumpHistoryLast2160Minutes: NSPredicate {
+        let date = Date.oneDayAgoInMinutes - TimeInterval(hours: 12)
         return NSPredicate(format: "timestamp >= %@", date as NSDate)
     }
 
--- a/Trio/Sources/APS/OpenAPS/OpenAPS.swift
+++ b/Trio/Sources/APS/OpenAPS/OpenAPS.swift
@@ -171,11 +171,11 @@ final class OpenAPS {
         return json
     }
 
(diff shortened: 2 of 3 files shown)


kingst:
I'm marking this one as a draft. After some discussions with @dnzxy, he did some analysis and it looks like this bug might be pretty old. Since it's not a simple bug fix I'll do a more rigorous analysis to help reason about the impact of fixing this bug.

marionbarker:
## Update

I got busy doing other things and did not complete the configuration. I just noticed, 2025-12-21 at 12:30 PST that the Trio phone had closed loop disabled. I switched it to closed loop when I noticed. So the time for this test should start from that timestamp.

About that same time, switch the Nightscout site to a different phone and switch this phone to use the on-app CGM simulator.

## Test

Seems like I might be testing code that's going to change. But the pod is connected to the phone already.

I merged kingst:fix-autosens-pump-history into fix-first-pod-alarm in a new branch in my local clone called dev_29_pr893_pr894. (The 29 is because dev was at 0.6.0.29 when I did the merge).

I deleted the app from the phone and then built this version.
* This will confirm that first pod beep truly is fixed.

If this PR gets updated before my pod is done, then I'll build an updated version.

In the meantime, the build has an actual DASH pod connected.
* I only filled it with 100 U of water
* I set the scheduled reminder to earliest possible alert for pod expiration and low reservoir to 50 U
* I also modified the settings from the default values, e.g., 0.1 U/hr, to 1 U/hr, CR to 15 g/U, ISF 50 mg/dL/U, SMB on always, UAM on and set minutes to 90 for both
* Closed Loop is on with max IOB of 10 U
* CGM uses Nightscout which has a glucose pattern ranging from a high event to a low event and repeating

kingst:
I did some [analysis](https://github.com/nightscout/Trio/issues/897#issuecomment-3680903646) on this change and based on my n==1 data (1) there can be significant changes and (2) the fix looks like an improvement. However, because the changes are significant, I'm going to close this PR and move it over to `oref-swift` so that we can evaluate it along with our other bug fixes that we've made as a part of the port.

PR #900, opened 2025-12-21
fix clipping on top of list in editor

EJM0:
fixes overlapping items in editor view
image

### DIFF
--- a/Trio/Sources/Modules/Onboarding/View/TherapySettingEditorView.swift
+++ b/Trio/Sources/Modules/Onboarding/View/TherapySettingEditorView.swift
@@ -52,104 +52,109 @@ struct TherapySettingEditorView: View {
                 .background(Color.chart.opacity(0.65))
                 .padding(.bottom, -8)
 
-                // Subtle separator at the head of the list
-                Rectangle()
-                    .fill(Color.secondary.opacity(0.3))
-                    .frame(height: 1)
-                    .padding(.horizontal, 20)
-
-                List {
-                    ForEach($items) { $item in
-                        VStack(spacing: 0) {
-                            Button {
-                                selectedItemID = selectedItemID == item.id ? nil : item.id
-                                sortTherapyItems()
-                            } label: {
-                                HStack {
-                                    HStack {
-                                        Text(displayText(for: unit, decimalValue: item.value))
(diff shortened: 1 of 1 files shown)


EJM0:
image

dnzxy:
Sorry, what's the issue here? What's getting clipped?

EJM0:
the custom list view and the chart above are overlapping

dnzxy:
Intentionally so, yes 😅

EJM0:
oh okay than neverminde

EJM0:
@dnzxy sorry im confused why are there screenshots of people on the discord where the layout is the same as in my pr.
Was this previously pathed out, if so why was the overlapping approach used?
[such a screen](https://cdn.discordapp.com/attachments/1239982177190678549/1402059233708736704/image0.jpg?ex=6949c96a&is=694877ea&hm=f6693614a4fc1f6c4f6b15a8bcab14b04ce2568f1b9fd7198249678cd59a684b&)

dnzxy:
> @EJM0:
> Was this previously pathed out, if so why was the overlapping approach used? [such a screen](https://cdn.discordapp.com/attachments/1239982177190678549/1402059233708736704/image0.jpg?ex=6949c96a&is=694877ea&hm=f6693614a4fc1f6c4f6b15a8bcab14b04ce2568f1b9fd7198249678cd59a684b&)


That is the "old" style of therapy setting editor before we brought over the newer style that was introduced when I wrote the app Onboarding. 

The Onboarding landed before Trio went from closed beta to open beta, but after all therapy editors and the entire settings of the app were redesigned/refactored. At the time, we wanted to bring the app to an open beta state, so postponed an update/refactor of the therapy settings editor to adopt the newer look to a later point in time. 

That later point in time came when @marv-out updated and refactored the therapy editors with #819 .
The new style for therapy settings is 1:1 that of Onboarding, and the clipping here is intentional – it basically hides the 1px grey divider :)

EJM0:
Ahh I now understand. I added a divider.(line 153)
After setting the padding to -8 there was no additional divider. This could also be a difference between iOS 18 and 26.(just a guess since the sdk is the same)
A version check for iOS 26 could be used to set the bottom padding to -8 for a clean solution.

PR #913, opened 2025-12-31
Cleanup and slightly refactor History

dnzxy:
* Rename/move files DataTable -> History
* Ensure strings and labels are properly wrapped with String(localized: …)
* Use checkbox square SF symbol for filters instead of circles

### DIFF
--- a/Trio/Sources/Modules/History/View/HistoryRootView.swift
+++ b/Trio/Sources/Modules/History/View/HistoryRootView.swift
@@ -2,7 +2,7 @@ import CoreData
 import SwiftUI
 import Swinject
 
-extension DataTable {
+extension History {
     struct RootView: BaseView {
         let resolver: Resolver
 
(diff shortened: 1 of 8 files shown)


dnzxy:
This one's for you @MikePlante1 :P

Trio/Sources/Modules/History/View/HistoryRootView.swift
```diff
@@ -252,7 +255,7 @@ extension DataTable {
                 },
                 label: {
                     HStack {
-                        Text(showFutureEntries ? "Hide Future" : "Show Future")
+                        Text(showFutureEntries ? String(localized: "Hide Future") : String(localized: "Show Future"))
```
MikePlante1, line 258: It seems like it was localizing fine without these `String(localized:_)` additions, but they don't hurt anything and I suppose it's probably the right way to do it.

MikePlante1 (approved):
LGTM, thanks! (code review and in-silica testing with Xcode 16.4, iPhone Air, iOS 26.1)

Trio/Sources/Modules/History/View/HistoryRootView.swift
marionbarker: With SwiftUI, Text is automatically localized (IIRC starting with Xcode 15). 

Strings need to have localized added.

Both should have a comment for context to help the translator.

* https://developer.apple.com/documentation/xcode/preparing-your-apps-text-for-translation

dnzxy: The overwhelming majority of our strings do not have a comment, so I kept it this way. 

My understanding is that a direct text in `Text()` is localized fine, but conditional variable based content isn’t. I can revert this change though.

PR #914, opened 2025-12-31
Fix Adjustment Badge to Display SMB &/or UAM minute changes

### LINKED ISSUE #895: Active Overrides Not Indicated on Main Profile Bar (SMB/UAM Minutes changed only)
When an override is activated—especially one that changes SMB/UAM minutes—the Main Screen Profile bar does not indicate that any override is active. The Profile bar appears unchanged, making it look like no override is applied.

The override does correctly show as active in:

Adjustments window

Live Activity

This occurs even though the override is functioning.

When any override is active, the Profile bar should clearly indicate that an override is applied, and ideally reflect overridden parameters such as SMB/UAM minutes.

dnzxy:
Fixes #895.

| Max Adjustments | SMB/UAM min only |
|--------|--------|
| image | image |

### DIFF
--- a/Trio/Sources/Modules/Home/View/HomeRootView.swift
+++ b/Trio/Sources/Modules/Home/View/HomeRootView.swift
@@ -220,6 +220,10 @@ extension Home {
                 return nil
             }
 
+            guard let settingsManager = state.settingsManager else {
+                return nil
+            }
+
             let percent = latestOverride.percentage
             let percentString = percent == 100 ? "" : "\(percent.formatted(.number)) %"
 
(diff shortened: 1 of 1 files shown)


MikePlante1:
Is a target of 100 mg/dL actually used in the screenshot example?

MikePlante1:
> Is a target of 100 mg/dL actually used in the screenshot example?

Tested it in-silica, and looks like this is a bug not related to this PR.

If you start a new or preset Override that doesn't Override Target, then Target is not displayed in the banner, but if you edit a running override (even if you don't enable Override Target) then the banner includes a false Target.

Trio/Sources/Modules/Home/View/HomeRootView.swift
```diff
@@ -264,9 +268,29 @@ extension Home {
...
+                .smbIsScheduledOff ? String(localized: "SMBs Off\(smbScheduleString)") : ""
+
+            var smbMinuteString: String = ""
+            var uamMinuteString: String = ""
+
+            if !latestOverride.smbIsOff || !latestOverride.smbIsScheduledOff {
+                if let smbMinutes = latestOverride.smbMinutes,
+                   smbMinutes.decimalValue != settingsManager.preferences.maxSMBBasalMinutes
+                {
+                    smbMinuteString = "SMB\u{00A0}\(smbMinutes)\u{00A0}" +
```
MikePlante1, line 280: Does `SMB` (and `UAM` in line 287`) have to be made into a `localized:` String?

MikePlante1 (commented):
.

Trio/Sources/Modules/Home/View/HomeRootView.swift
dnzxy: I was unsure and decided not to localize it as I considered it a fixed term. 

Happy to wrap it though.

MikePlante1 (approved):

PR #917, opened 2026-01-01
Claude/add watch complications mnd9b

edwardmsalem:


MikePlante1:
I changed the target branch to dev instead of main, but there is still no description at all for this PR or even a screenshot.

From the title, I assume this is a vibe coded attempt to add a Watch complication that displays current glucose. 

We have specifically chosen _not_ to include glucose Watch complications as Apple has kneecapped Watch complications’ refresh rate so they rarely show the current glucose and are often ~15 m out of date.

Have you tried the Contact Complication feature available in Trio?: Settings > Devices > Smart Watch > Apple Watch > Contact Complications

edwardmsalem:
hey thanks for the reply

this is just for my own use. the contact complication wasnt working great for me, it would show old values with no way to know how stale they were. 15 min delays are fine for me
 i built this with:

∙ staleness indicators (yellow at 10 min, shows “–” in red at 15 min)
∙ background refresh every 5 min
∙ tap to force refresh
totally understand if you want me to close it or make changes, whatever works for you guys. I am in love with what you guys have built and I cannot thank you enough.

Edward Salem

PR #923, opened 2026-01-06
Update submodules to 2026-01-05 version

marionbarker:
## Purpose: 

Bring submodules up to the latest version.

No functional change beyond aligning Trio with upstream repositories.

* The DanaKit submodule was being linted - I fixed this by adding a missing comma in `scripts/swiftformat.sh`
* Most submodule updates were localizations
* The ones with updates in addition to localizations were:
   * LibreTransmitter: stop logging the glucose readings that are not used (reduced Libre logging by 80%)
   * LoopKit: add a purge function that is used only by Loop (old Loop logs were being kept longer than desired)
   * MinimedKit: add a CAGE and IAGE row to the in-app screen. This gets populated after the first rewind and first fixed prime
   * DanaKit (commit range `299331d4e..bad8fad9c`):
      * prevent sync data and loop in parallel
      * improved callback cleanup
   * G7SenorKit updated to support 15.5 day sensors
   * CGMBLEKit and G7SensorKit updated to make upload glucose enabled by default and because the code for trio branch was identical to the upstream repo, the trio branch for each was rebased; each has the same SHA as LoopKit

## Test

Built with Xcode 26.2 and tested with Dana rPi simulator.

### DIFF
--- a/Config.xcconfig
+++ b/Config.xcconfig
@@ -19,7 +19,7 @@ TRIO_APP_GROUP_ID = group.org.nightscout.$(DEVELOPMENT_TEAM).trio.trio-app-group
 
 // The developers set the version numbers, please leave them alone
 APP_VERSION = 0.6.0
-APP_DEV_VERSION = 0.6.0.37
+APP_DEV_VERSION = 0.6.0.38
 APP_BUILD_NUMBER = 1
 COPYRIGHT_NOTICE =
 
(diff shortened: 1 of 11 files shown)


dnzxy (approved):

dnzxy:
Approving PR.
@marionbarker please update the description to include the G6/G7 changes and 15-day support.

marionbarker:
Original comment edited as requested.

PR #925, opened 2026-01-06
Fix faulty target display when editing overrides #924

### LINKED ISSUE #924: Editing Override w/o Target Change Always Displays Target in Badge
Reported when testing #914 :
> >Is a target of 100 mg/dL actually used in the screenshot example?
> 
> Tested it in-silica, and looks like this is a bug not related to this PR.
> 
> If you start a new or preset Override that doesn't Override Target, then Target is not displayed in the banner, but if you edit a running override (even if you don't enable Override Target) then the banner includes a false Target.

dnzxy:
### Summary

Fixes an issue where an override target was displayed incorrectly while editing.
Fixes #924.

### Details

* Default target value is now treated as `0` instead of `100`
* Prevents showing an unintended target value when no explicit override target is set
* Ensures the override summary only displays a target when one is actually defined

### Screenshots from Simulator

| Initial OR | OR Edit1 | OR Edit2 |
|--------|--------|--------|
| image | image | image |

### DIFF
--- a/Trio/Sources/Modules/Home/View/HomeRootView.swift
+++ b/Trio/Sources/Modules/Home/View/HomeRootView.swift
@@ -228,7 +228,7 @@ extension Home {
             let percentString = percent == 100 ? "" : "\(percent.formatted(.number)) %"
 
             let unit = state.units
-            var target = (latestOverride.target ?? 100) as Decimal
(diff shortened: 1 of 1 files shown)


MikePlante1 (approved):
Approved based on code review and testing.

PR #926, opened 2026-01-06
Adjust when SMB/UAM Mins are shown in Adjustment Banner

MikePlante1:
Since 0.6.0.37, the adjustments banner on the main view shows when

`smbIsOff` is false

or when

`smbIsScheduledOff` is false

... but they are never both true, so it always passes that line.

---

This can wrongly happen when using 0.6.0.37 without this PR:

---

I removed the `smbIsScheduledOff` check, because when `smbIsScheduledOff` is true you can still change the SMB/UAM Minutes for the times SMB is not scheduled off.

I added a check for `advancedSettings` which is the boolean that the `Override Max SMB Minutes` toggle controls.

### DIFF
--- a/Trio/Sources/Modules/Home/View/HomeRootView.swift
+++ b/Trio/Sources/Modules/Home/View/HomeRootView.swift
@@ -273,7 +273,7 @@ extension Home {
             var smbMinuteString: String = ""
             var uamMinuteString: String = ""
 
-            if !latestOverride.smbIsOff || !latestOverride.smbIsScheduledOff {
+            if !latestOverride.smbIsOff, latestOverride.advancedSettings {
                 if let smbMinutes = latestOverride.smbMinutes,
                    smbMinutes.decimalValue != settingsManager.preferences.maxSMBBasalMinutes
                 {


dnzxy:
I added that because I felt it makes no sense to display minutes when it’s scheduled to be off, but unfortunately the or condition will always result in true as it seems. 

Since we actually do already check for when the SMBs are scheduled off, wouldn’t it be more precise to check for null/empty of that string, only then show minutes? 

Right now, with the condition in place as it, is would display minutes with a SMBs to be scheduled off (from what I‘m reading), and I didn’t want that.

MikePlante1:
> I added that because I felt it makes no sense to display minutes when it’s scheduled to be off, but unfortunately the or condition will always result in true as it seems.

Scheduled off means there are still times when it's scheduled on, though. And during those times it's scheduled on, UAM/SMB Minutes can be altered via the OR.

> Since we actually do already check for when the SMBs are scheduled off, wouldn’t it be more precise to check for null/empty of that string, only then show minutes?

Not sure I follow.

> Right now, with the condition in place as it, is would display minutes with a SMBs to be scheduled off (from what I‘m reading), and I didn’t want that.

Only if they've also toggled on the "Override Max SMB Minutes" toggle.

dnzxy:
A bit further up we from where you did the change we construct an smbScheduledOff string. My thinking was checking for that. 

You’re totally right though. I don’t feel like making it overly complex to display and then hide changed minutes does not make sense. Let’s go with yours! ☺️

dnzxy (approved):

type1berger:
IMG_0802
IMG_0801

Looks good to me!

Issue #912, opened 2025-12-30
The preset meal are logging wrong carbs


ahmmadqaqa:
## Describe the bug
*A clear and concise description of what the bug is. Describe what you see versus what you expect to see.*

## Attach a Log
*Tap the Trio settings icon at the bottom of the screen, then tap 'Share logs' on the bottom of the list and attach it to this ticket.*

## To Reproduce
*Steps to reproduce the behavior:*
1. *Go to '...'*
2. *Click on '....'*
3. *Scroll down to '....'*
4. *See error*

## Expected behavior
*A clear and concise description of what you expected to happen.*

## Screenshots
*If applicable, add screenshots to help explain your problem.*

### Smartphone:
* Hardware: *[e.g. iPhone 15 Pro]*
* OS Version: *[e.g. iOS 17.5]*

### Pump:
* Manufacturer: *[e.g. Insulet]*
* Model: *[e.g. Omnipod Dash or Eros]*

### CGM:
* Device: *[e.g. Dexcom G7]*
* Manager app: *[e.g. Dexcom App or xDrip4iOS]*

### Trio Version:
* Version Number: *[e.g. 1.9.2]*
* Repo: *nightscout/trio*
* Git Reference: *[e.g. commit hash]*

## Technical Details
*If applicable, provide any technical details that might help in diagnosing the problem. This could include logs, error messages, or relevant configuration details.*

## Additional context
*Add any other context about the problem here.*

dnzxy:
Hi @ahmmadqaqa 
Please follow the bug issue template and fill in all applicable information. 
Without any details, your report will not let us determine whether there is an issue here or what potentially requires a fix. 
Thanks

ahmmadqaqa:
Describe the bug

While logging a preset on treatment I realized the carbs entry were different than the preset. It was like 3x more. The issue is cleared after I restarted the app. 



Setup Information (please complete the following information):

Smartphone: IPhone 17 pro max
OS Version: iOS 26.2
Pump: Medtronic 754
CGM: Dexcom G7 app
Trio Version: 0.2.9

Version Number: 0.6.0.34
Repo: nightscout/trio

![Image](https://github.com/user-attachments/assets/d86f1fd6-22af-4e9d-ac88-184cf6dc18b2)

dnzxy:
What are the individual presets looking like (carb, fat, protein)? 
Walk us through the steps (in the UI) and buttons / items you taped and in which order to get to the issue, please.

[...]

PR #929, opened 2026-01-07
Prevent "New Preset" from resetting list

### LINKED ISSUE #928: Meal presets UI issue can lead to over-delivery
v0.6.0.37

This issue is reproducible:

1) From the Main screen, tap "+" to get to Treatments.
2) Tap on "Presets +".
3) Select one of your preset food items from the Saved Food list.
4) Note the carb/protein/fat counts in the summary.  Also note that the quantity is "1 x [your food].
5) Click on the "+" in "New Preset +" _because you hadn't been aware that you could add multiple food items through this one screen, but you saw an image in someone else's bug report that had multiple foods, and you thought this was how you could add another food item._ ;-)
6) Observe that the Add Meal Preset dialogue opens.  Click cancel _because now you know this is not the way to do this_.
7) Observe that the name of the food you selected and nutrient summary have both disappeared.
8) Repeat steps 2-4 to re-add the food.
9) Notice that while the summary box still says "1 x [your food]", the nutrients have doubled.
10) Click "Add to Treatments" and observe that the insulin recommendation is using the doubled nutrients for its calculation.

MikePlante1:
Addresses Issue:
  * #928
  
  and possibly:
  *  #912 
  
  ---
  
  1. Removed resetValues() from "New Preset" button

  This was the root cause of the bug. When tapping "New Preset", it immediately cleared
  state.summation and all accumulated values before the sheet opened. If you then
  cancelled, the data was already gone.

  2. Changed onCancel to use resetNewPresetForm() instead of resetValues()

  When cancelling the "New Preset" sheet, only clear the form fields for the new preset
  being created, not the presets already added to the meal.

  3. Added resetNewPresetForm() function

  A targeted reset that only clears the 4 form fields (dish, presetCarbs, presetFat,
  presetProtein).

  4. Simplified resetValues() to only clear state model properties

  Removed clearing of @State variables that are destroyed when the view disappears anyway.
   Now only clears state.selection and state.summation.

  5. Removed redundant resetValues() from "Close" button

  The .onDisappear modifier already handles this when the view dismisses.

  6. Removed resetValues() from savePreset()

  The form fields reset automatically when the sheet dismisses, and we don't want to clear
   the accumulated meal values after saving a new preset definition.

### DIFF
--- a/Trio/Sources/Modules/Treatments/View/MealPreset/MealPresetView.swift
+++ b/Trio/Sources/Modules/Treatments/View/MealPreset/MealPresetView.swift
@@ -66,7 +66,6 @@ struct MealPresetView: View {
                 ToolbarItem(placement: .topBarLeading) {
                     Button {
                         dismiss()
-                        resetValues()
                     } label: {
                         Text("Close")
                     }
@@ -74,7 +73,6 @@ struct MealPresetView: View {
                 ToolbarItem(placement: .topBarTrailing) {
                     Button(action: {
                         showAddNewPresetSheet.toggle()
-                        resetValues()
                     }, label: {
                         HStack {
                             Text("New Preset")
@@ -93,7 +91,7 @@ struct MealPresetView: View {
                     onSave: savePreset,
                     onCancel: {
                         showAddNewPresetSheet.toggle()
-                        resetValues()
+                        resetNewPresetForm()
                     }
                 )
             }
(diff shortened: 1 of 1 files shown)


dnzxy:
@g5WS would you care to spin this up for a test? :)

dnzxy (approved):
Haven’t test run this. Please merge after pending test.

g5WS:
> @g5WS would you care to spin this up for a test? :)

Worked for me!  Nicely done!

PR #930, opened 2026-01-07
Add "Color Mode" to Contact Complication Image options

MikePlante1:
This option was requested by a user on Discord because the dynamic color made the glucose hard to read when used on a monochrome watch face.

User's screenshot before this PR:

Color Mode option this PR adds:

### DIFF
--- a/Model/Classes+Properties/ContactTrickEntryStored+CoreDataProperties.swift
+++ b/Model/Classes+Properties/ContactTrickEntryStored+CoreDataProperties.swift
@@ -17,6 +17,7 @@ public extension ContactImageEntryStored {
     @NSManaged var ringWidth: Int16
     @NSManaged var ringGap: Int16
     @NSManaged var id: UUID?
+    @NSManaged var colorMode: String?
     @NSManaged var fontSize: Int16
     @NSManaged var fontSizeSecondary: Int16
     @NSManaged var fontWidth: String?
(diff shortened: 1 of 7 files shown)


dnzxy (approved):
I like it! Approved based on code review, I haven’t test-run this.

pontuslandgren:
I am testing this now, works well for me and I like it!

tswire:
incoming-7060B4D3-09A1-4870-B3E9-D572496AF0A5
incoming-018664E6-292C-4DDA-9C4B-197110064219
Tested as well and looks PERFECT. The only tiny issue I see is when the watch goes from AOD to on or on to AOD there is a small dimming (like back to a grey color) effect that immediately goes back to white once in AOD (or ON).  First image above is before the change and second is after this change.

MikePlante1:
> The only tiny issue I see is when the watch goes from AOD to on or on to AOD there is a small dimming (like back to a grey color) effect that immediately goes back to white once in AOD (or ON).

The weird transition from wrist down to wrist up (and wrist up to wrist down) with Always On Display seems to be an Apple thing outside of our control, unfortunately.

MikePlante1:
Also to note, in researching this PR, I found that the `UIGraphicsBeginImageContext` we use depreciated in iOS 17 so we should replace it with `UIGraphicsImageRenderer`... but it works fine as-is, so I'm just gonna ignore that for now to reduce changes.

Issue #723, opened 2025-07-21
HUD: mmol/L rounding issue


g5WS:
v0.5.1.5

Observe: 
- Trio recorded the most recent  two BGs as 5.7 (10:52pm) and 5.2 (10:57pm)
- 5.2 - 5.7 = -0.5
- Trio displayed a delta of -0.4 instead of -0.5 (xDrip also calculated -0.5, as seen in Dynamic Island)

![Image](https://github.com/user-attachments/assets/3a1a11d2-1e13-4c9d-97f6-21bb21434e9f)
![Image](https://github.com/user-attachments/assets/8eba530f-8b77-4c2a-a504-aa360483499f)

bastiaanv:
I've noticed this too. Issue is related to the second decimal character. In the Health app you can see the actual value being used.

g5WS:
Can someone talk me through why this is closed?

g5WS:
@MikePlante1 , should this be reopened again?

bastiaanv:
@g5WS Would you be willing to test this PR with me?
https://github.com/nightscout/Trio/pull/935

g5WS:
> [@g5WS](https://github.com/g5WS) Would you be willing to test this PR with me? [#935](https://github.com/nightscout/Trio/pull/935)

Tried!  Not fully confident in how I am setting things up, but the build failed. 

https://github.com/g5WS/Trio/actions/runs/21008226398/job/60396283100

I will retry the build. The error messages suggest (to me anyway!) that the actual build was successful, but that they upload to TestFlight was not.

MikePlante1:
> > [@g5WS](https://github.com/g5WS) Would you be willing to test this PR with me? [#935](https://github.com/nightscout/Trio/pull/935)
> 
> Tried! Not fully confident in how I am setting things up, but the build failed.
> 
> https://github.com/g5WS/Trio/actions/runs/21008226398/job/60396283100
> 
> I will retry the build. The error messages suggest (to me anyway!) that the actual build was successful, but that they upload to TestFlight was not.

That error looks unrelated. This is how to address it from LoopDocs, but the same thing for Trio. So for your case, you should use `126`: https://loopkit.github.io/loopdocs/browser/bb-errors/#intermittent-testflight-upload-number-error

g5WS:
It seems to have built correctly just by stubbornly asking it to rebuild three times. :-)

[...]

PR #935, opened 2026-01-14
Fix mmol/L delta calculation 

bastiaanv:
As described in #723, the glucose change might be incorrect by 0.1 mmol/L in some specific cases. This might be because there is a change in mg/dl, but not when rounded to mmol/L.

This PR fixes this issue by calculating the rate in mmol/L instead of in mg/dl

I am running this PR currently on my device, want to test this for a week before making this ready.


dnzxy:
Is this ready for review now?

bastiaanv:
No, not yet. It is much better, but not perfect yet. The second decimal is being difficult atm :)

bastiaanv:
I don't think I can get this more precise than this.
The asMmolL calculations look good, etc

dnzxy (approved):

PR #939, opened 2026-01-19
Reduce complexity in GarminManager

mountrcg:
### GarminMananger refactoring

I had a lot of checking IOB, Determination and Glucose updates to reduce message volume to Garmin - many were unnecessary or low yield but adde complexity. I simplified that.

It was necessary to make it easier to identify some issues that have come up with Garmin updates starting in the beginning of December. The so far not never logged errors `Garmin: Device status -> bluetoothNotReady` started creeping up and can lead to stalled Garmin messaging execution, that required a Trio app restart. So with additional logging I hope to get more insights into this.


mountrcg:
I rebased earlier instead of merging dev, re-oping it with vlean merge of dev

PR #944, opened 2026-01-20
Refactor Garmin: simplify manager, add documentation, and improve simulator testing

mountrcg:
I had a lot of checking IOB, Determination and Glucose updates in order to reduce message volume to Garmin devices - many were unnecessary or low yield but added complexity. I simplified that.

It was necessary to make it easier to identify some issues that have come up with Garmin updates starting in the beginning of December. The so far not never logged errors Garmin: Device status -> bluetoothNotReady started creeping up and can lead to stalled Garmin messaging execution, that required a Trio app restart. So with additional logging I hope to get more insights into this.

### Removed Complexity:
- Eliminated complex caching mechanisms (app installation status, watchface change tracking)
- Removed connection state tracking and related caching logic  
- Simplified throttle/debounce implementation
- Removed @unchecked Sendable conformance and NSLock instances
- Cleaned up device connection state tracking

### Simplified State Management:
- More direct dispatch queue usage for timer management
- Streamlined glucose/determination coordination for watch updates
- Documented fallback delay configuration with reasoning
- improved logging for connectIQ states

### Documentation Restoration
- Restored all documentation comments that were present in dev branch
- Added doc strings for the new helper methods (debugGarmin, formatIOB, etc.)

### Simulator Testing Improvements
Enhanced simulator testing workflow:
- Removed auto-creation of simulated devices
- Added manual mock device creation via UI
- Created MockIQDevice class with factory method createSimulated()
- Improved testing workflow to mimic real user experience

### DIFF
--- a/Trio/Sources/Services/WatchManager/GarminManager.swift
+++ b/Trio/Sources/Services/WatchManager/GarminManager.swift
@@ -103,6 +103,13 @@ final class BaseGarminManager: NSObject, GarminManager, Injectable {
         debug(.watchManager, message)
     }
 
+    // MARK: - Device Ready State (SDK 1.8+)
+
+    /// Tracks which devices have completed characteristic discovery and are ready for communication.
+    /// In SDK 1.8+, `deviceStatusChanged: connected` does NOT mean the device is ready.
+    /// We must wait for `deviceCharacteristicsDiscovered:` before sending messages.
+    private var readyDevices: Set<UUID> = []
+
     // MARK: - Deduplication
 
     /// Hash of last sent data to prevent duplicate broadcasts
@@ -712,6 +719,10 @@ final class BaseGarminManager: NSObject, GarminManager, Injectable {
         // Without this, hash deduplication could skip sending to new apps if data unchanged
         lastSentDataHash = nil
 
+        // Clear ready state - devices need to complete characteristic discovery again
+        // after re-registration (SDK 1.8+ requirement)
+        readyDevices.removeAll()
+
         for device in devices {
             connectIQ?.register(forDeviceEvents: device, delegate: self)
 
@@ -845,6 +856,13 @@ final class BaseGarminManager: NSObject, GarminManager, Injectable {
 
         watchApps.forEach { app in
             let appName = self.appDisplayName(for: app.uuid!)
+
+            // Check if device is ready (SDK 1.8+ requirement)
+            guard let device = app.device, readyDevices.contains(device.uuid) else {
+                debugGarmin("Garmin: Skipping \(appName) - device not ready")
+                return
+            }
+
             connectIQ?.getAppStatus(app) { [weak self] status in
                 guard status?.isInstalled == true else {
                     debug(.watchManager, "Garmin: App not installed: \(appName)")
@@ -946,25 +964,43 @@ extension BaseGarminManager: IQUIOverrideDelegate, IQDeviceEventDelegate, IQAppM
     // MARK: - IQDeviceEventDelegate
 
     /// Called whenever the status of a registered Garmin device changes (e.g., connected, not found, etc.).
+    /// Note: In SDK 1.8+, `connected` does NOT mean ready for communication.
+    /// Wait for `deviceCharacteristicsDiscovered:` before sending messages.
     /// - Parameters:
     ///   - device: The device whose status has changed.
     ///   - status: The new status for the device.
-    func deviceStatusChanged(_: IQDevice, status: IQDeviceStatus) {
+    func deviceStatusChanged(_ device: IQDevice, status: IQDeviceStatus) {
         // Always log connection state changes - critical for diagnosing SDK issues
         switch status {
         case .invalidDevice:
             debug(.watchManager, "Garmin: Device status -> invalidDevice")
+            readyDevices.remove(device.uuid)
         case .bluetoothNotReady:
             debug(.watchManager, "Garmin: Device status -> bluetoothNotReady")
+            readyDevices.remove(device.uuid)
         case .notFound:
             debug(.watchManager, "Garmin: Device status -> notFound")
+            readyDevices.remove(device.uuid)
         case .notConnected:
             debug(.watchManager, "Garmin: Device status -> notConnected")
+            readyDevices.remove(device.uuid)
         case .connected:
-            debug(.watchManager, "Garmin: Device status -> connected")
+            debug(.watchManager, "Garmin: Device status -> connected (waiting for characteristics)")
         @unknown default:
             debug(.watchManager, "Garmin: Device status -> unknown(\(status.rawValue))")
         }
+    }
+
+    /// Called when device characteristics are discovered and the device is ready for communication.
+    /// This is required in SDK 1.8+ - sending before this callback may fail.
+    /// - Parameter device: The device whose characteristics have been discovered.
+    func deviceCharacteristicsDiscovered(_ device: IQDevice) {
+        debug(.watchManager, "Garmin: Device characteristics discovered - ready for communication")
+        readyDevices.insert(device.uuid)
+
+        // Trigger a data send now that device is ready
+        // This ensures newly connected devices get data promptly
+        triggerWatchStateUpdate(triggeredBy: "DeviceReady")
     }
 
     // MARK: - IQAppMessageDelegate
--- a/Trio/Sources/Services/WatchManager/GarminManager.swift
+++ b/Trio/Sources/Services/WatchManager/GarminManager.swift
@@ -87,7 +87,9 @@ final class BaseGarminManager: NSObject, GarminManager, Injectable {
     /// Enable/disable general Garmin debug logging (connections, sends, etc.)
     private let debugGarminEnabled = true
 
-    /// Helper method for conditional Garmin debug logging
+    /// Helper method for conditional Garmin debug logging.
+    /// Logs messages only if debugGarminEnabled is true.
+    /// - Parameter message: The debug message to log.
     private func debugGarmin(_ message: String) {
         guard debugGarminEnabled else { return }
         debug(.watchManager, message)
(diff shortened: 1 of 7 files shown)


mountrcg:
I cleaned up the logging and switched it off for the stable WatchState creation and debouncing/throttling the different triggers as that works fine. 
ConnectIQ package update 1.8 addressed bluetooth reconnection issue. I hope this is what I was seeing starting early december - so I manually updated the package from 1.7 to 1.8 - do not know why it was not automatically.

Would be great to get that into the feature branch.

Trio/Sources/Services/WatchManager/GarminManager.swift
```diff
@@ -79,97 +73,83 @@ final class BaseGarminManager: NSObject, GarminManager, Injectable, @unchecked S
...
 
-    /// Helper method for conditional Garmin debug logging
+    // MARK: - Debug Logging
+
+    /// Enable/disable watch state preparation and throttling logs:
+    /// - Glucose/IOB received, waiting for determination
+    /// - Determination arrived, cancelled timer
+    /// - Preparing/Prepared watch state, Skipping - data unchanged
+    /// - Settings throttle timer started/running/fired
+    private let debugWatchState = true
```
kingst, line 89: Should this be `false` for prod or do we want this on by default?

Trio/Sources/Services/WatchManager/GarminManager.swift
```diff
@@ -79,97 +73,83 @@ final class BaseGarminManager: NSObject, GarminManager, Injectable, @unchecked S
...
+
+    /// Enable/disable watch status and communication logs:
+    /// - Device status changes (connected, notConnected, etc.)
+    /// - Registered watchface/datafield
+    /// - Sending to / Successfully sent
+    /// - Watchface/datafield config changed
+    private let debugGarminEnabled = true
```
kingst, line 96: Same question here, should this be `false`?

Trio/Sources/Services/WatchManager/GarminManager.swift
```diff
@@ -519,500 +497,254 @@ final class BaseGarminManager: NSObject, GarminManager, Injectable, @unchecked S
...
-                    .replacingOccurrences(of: "  ", with: " ")
-
-                // Show which apps will actually receive data
-                let destinations: String
-                if !isWatchfaceDataEnabled {
-                    destinations = "datafield \(datafieldUUID) only (watchface disabled)"
-                } else {
-                    destinations = "watchface \(watchfaceUUID) / datafield \(datafieldUUID)"
+            // Deduplicate: Check if data is unchanged from last preparation
+            let currentHash = watchStates.hashValue
```
kingst: is the `.hashValue` stable? I don't think it is, but this is just from memory so I could be wrong.

Trio/Sources/Services/WatchManager/GarminManager.swift
```diff
@@ -1096,215 +825,56 @@ final class BaseGarminManager: NSObject, GarminManager, Injectable, @unchecked S
...
+    private func broadcastWatchStateData(_ data: Data) {
+        // Deduplicate: Use stable content-based hash (sorted JSON bytes)
         let currentHash: Int
-        if let jsonData = try? JSONSerialization.data(withJSONObject: state, options: [.sortedKeys]) {
-            currentHash = jsonData.hashValue
+        if let sortedData = try? JSONSerialization.data(
+            withJSONObject: JSONSerialization.jsonObject(with: data, options: []),
+            options: [.sortedKeys]
+        ) {
+            currentHash = sortedData.base64EncodedString().hashValue
```
kingst, line 841: Same question here about `.hashValue`

Trio/Sources/Services/WatchManager/GarminManager.swift
```diff
@@ -1096,215 +825,56 @@ final class BaseGarminManager: NSObject, GarminManager, Injectable, @unchecked S
...
-
-            // 1. If it's a datafield, ALWAYS send (no status check)
-            if isDatafieldApp {
-                debug(.watchManager, "[\(formatTimeForLog())] Garmin: Sending to datafield \(app.uuid!) (no status check)")
-                // Store hash to mark as sent on successful send
-                currentSendHash = hashToSend
-                sendMessage(updatedState, to: app)
-                return
-            }
+            let appName = self.appDisplayName(for: app.uuid!)
```
kingst: We can't force unwrap. I know that it exists but I'm opposed to adding new instances of this outside of test cases

Trio/Sources/Services/WatchManager/GarminManager.swift
```diff
@@ -1487,530 +964,117 @@ extension BaseGarminManager: IQUIOverrideDelegate, IQDeviceEventDelegate, IQAppM
...
     /// - Parameters:
     ///   - message: The message content from the watch app.
     ///   - app: The watch app sending the message.
-    /// Handle messages from watch apps
-    /// Always processes datafield messages, checks settings for watchface
     func receivedMessage(_ message: Any, from app: IQApp) {
-        debugGarmin("[\(formatTimeForLog())] Garmin: Received message \(message) from app \(app.uuid!)")
-
-        let validUUIDs = Set([currentWatchface.watchfaceUUID, currentGarminSettings.datafield.datafieldUUID].compactMap { $0 })
+        let appName = appDisplayName(for: app.uuid!)
```
kingst: Same here, no forced unwraps

kingst (changes requested):
A few high level comment, but generally looks like an improvement. My one required change is getting rid of the forced unwrap but generally I just have a few questions.

I have no way to test this code, so we really need some testing help from our garmin users and it would be useful to have someone who is familiar with this code to also take a look because this is a fairly big change.

dnzxy:
I concur with Sam on his comments. Let's please address them and then have this move in, @mountrcg :)

Trio/Sources/Services/WatchManager/GarminManager.swift
mountrcg: The PR's right now are for the feature/garmin branch and more extensive testing by Garmin users with the watchfaces and datfields available on Garmin connectIQ store. So even after PR we are still in test mode. For this time the dubugging shall be true, later once in dev it can be off

Trio/Sources/Services/WatchManager/GarminManager.swift
mountrcg: this one I thought to have it kept on true, also later on when moving feature into dev.

Trio/Sources/Services/WatchManager/GarminManager.swift
mountrcg: I can only say from tetsing that this condition often prevents the generation of a new watchstate, if nothing changes - fires quite often

Trio/Sources/Services/WatchManager/GarminManager.swift
mountrcg: done

Trio/Sources/Services/WatchManager/GarminManager.swift
mountrcg: done

mountrcg:
> A few high level comment, but generally looks like an improvement. My one required change is getting rid of the forced unwrap but generally I just have a few questions.
> 
> I have no way to test this code, so we really need some testing help from our garmin users and it would be useful to have someone who is familiar with this code to also take a look because this is a fairly big change.

Thank you for the review. Did the minor changes. You can test the UI elements and general flow for Garmin watch configuration in simulator - which was already moved to feature branch before, Settings > Devices > Smart watch ...

The critical stuff can only be tested with a live connection to watch, training device.

I will revert the changes coming in with SDK 1.8 as a more critical bug presents itself with it. No while doing the additional step reconnecting it seems that meesages will be delivered to watch but not acknoledget or processed at that time, so the queing builds up and with many reconnects the watch could get into a state where it only shows data 1 or more loops delayed. that is a watch/garmin framework issue. I have been chasing this goose around the block so much. Need to test it more outside the feature branch. In 1.7 one would just have to restart Trio or toggle BT - re-installing watchface is required at the moment if that fault creeps up. So one more commit coming nailing the SDK to 1.7 and reducing the re-registering steps.

--edit--
reverted back to 1.8 as I found the issues I was tackling in a long standing issue introduced in watchface background handling from 2023.

Trio/Sources/Services/WatchManager/GarminManager.swift
mountrcg: As I really have no clue about the question I asked Claude about the hash subject:

> In Swift, .hashValue is not stable across process launches — Hashable uses random seeding per process for security. But within a single process run, it is consistent: the same value will produce the same hash for the lifetime of the app.
So for this in-memory deduplication purpose (comparing lastPreparedDataHash which is also set in the same process), it works correctly. The hash resets on app restart anyway since lastPreparedDataHash is just a private var, not persisted.
The fact that it fires often and prevents redundant preparation is the intended behavior — if glucose/IOB/determination data hasn't changed between two triggers, there's no need to rebuild the watch state array.
The broadcast-level dedup at line ~837 uses a different approach (sorted JSON bytes → base64 → hashValue), which is also stable within a process but more expensive. Both are fine for their purpose.

mountrcg:
Added more logic to prevent redundant or unnecessary sends to watch. Sry testing the other stuff and reading the logs, toggling changes just made me discover more to optimize. On 4 weeks trekking hikes that might save some relevant battary capacity - well I guess not. But the whole optimizing since I started in Oct yielded around 2-3% per day. Hope that someone can confirm that.

tswire:
I have been testing this PR for a week and now testing a couple of those recent changes.  I am not a developer and can't speak to specific code recommendations but this has been working flawlessly for 10 days now.  I have not had any stale data issues, battery issues or anything to speak of.  Robert has provided clear instructions on setting this up and most important is if you want to switch watchfaces from Trio to another one.  Garmin connection can be finicky and as long as you properly disable data in Garmin device options in Trio, switching from and to Trio specific watchfaces works.

mountrcg:
@dnzxy and @kingst this now only includes the code reviewed and the requested changes. I will put new stuff in a new PR so that this can move forward as Dan suggested.

PR #951, opened 2026-01-26
Refactor FPU handling

dnzxy:
* Minimum carb equivalents size of 10g
* Maximum carb equivalents size of 33g
* Up to 3 carb equivalent entries, 99g total, per single meal announcement
* Start adding carb equivalent entries after delay-setting minutes after carb entry
* Space out >1 carb equivalent entries at +30m of initial carb equivalent entry
* Default handling: carb entry, 1st equiv +60, 2nd equiv +60+30, 3rd equiv +60+60 (each calc'd from initial timestamp)
* Evenly split calculated carb equivalents, e.g.
   * 33g becomes single entry of 33g
   * 60g becomes 30 + 30
   * 90g becomes 30 + 30 + 30
   * 150g becomes 33 + 33 + 33

This will be PR part 1 of 2. Part 2 will refactor the main chart UI and how we display carb equivalents to align their representation closer to how carbs are currently rendered, since we are moving away from a senseless spread to more sensible delayed equivalent entry(ies).

### DIFF
--- a/Trio/Sources/APS/Storage/CarbsStorage.swift
+++ b/Trio/Sources/APS/Storage/CarbsStorage.swift
@@ -127,83 +127,46 @@ final class BaseCarbsStorage: CarbsStorage, Injectable {
     }
 
     /**
-     Processes fat and protein entries to generate future carb equivalents, ensuring each equivalent is at least 1.0 grams.
+     Converts fat and protein into delayed carb-equivalent entries (FPU handling).
 
-     - The function calculates the equivalent carb dosage size and adjusts the interval to ensure each equivalent is at least 1.0 grams.
-     - Creates future carb entries based on the adjusted carb equivalent size and interval.
+     Behavior:
+
+     - Calculates carb equivalents from fat and protein
+       ((fat × 9 + protein × 4) / 10 × adjustment factor).
+     - Rounds down to whole grams.
+     - Drops values below 10 g.
+     - Caps total equivalents at 99 g.
+     - Splits into up to 3 entries.
+     - Caps each entry at 33 g.
+     - Distributes grams as evenly as possible.
+
+     Timing:
+
+     - First entry is scheduled after the configured delay
+       (default: 60 minutes) from the carb entry timestamp.
+     - Additional entries are spaced 30 minutes apart.
+
+     Example (default):
+
+     - Carb entry at T
+     - 1st equivalent at T + 60 min
+     - 2nd equivalent at T + 90 min
+     - 3rd equivalent at T + 120 min
+
+     Generated entries:
+
+     - Are marked with `isFPU = true`
+     - Contain only carbs (fat and protein set to 0)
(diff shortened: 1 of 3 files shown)


kingst:
High level review for now. In general, I really like the direction this is taking. It's simplifying an unnecessarily complex calculation and providing more predictable result, we should move forward with this.

dnzxy:
Thanks for the initial "gut-feel" review @kingst. I've added unit tests and updated docstrings; should be ready for review and testing now. I'll coordinate with folks on Discord for end-user tests.

dnzxy:
Updated the unit test again, to force testing the split of entries (currently evenly weighted entries).
Big question for testing will be whether to keep the even splits, or to create the up to 3 entries "front-heavy", e.g. total carb equivalents = 50g, split up as 1x33g and 1x 17g. Current behavior would create 2x 25g.

Stampertina:
Should 150 be 50 + 50 + 50?

dnzxy:
@Stampertina 
> Should 150 be 50 + 50 + 50?

The way this currently works, we are limiting `carb equivalents` to 99g, so 3x33g. 

Getting to this value, given the default fat-protein adjustment-factor setting, you would need a carb entry that has fat and protein to an extent of 1980 kcal, so an entry that has _more than_ 150g fat and 150g protein. 

Carb equivalents are calculated as follows:
* `(grams of fat x 9 + grams of protein x 4) / 10` to get to kcal
* then `kcal amount x fat-protein adjustment-factor`
* this value is the "final" `carb equivalent value in grams`
* which is then further processed

Trio/Sources/APS/Storage/CarbsStorage.swift
kingst: [Minor] Can this be deleted? I don't think it's being used anymore.

kingst (approved):
LGTM, one minor and optional comment but let's ship this.

Trio/Sources/APS/Storage/CarbsStorage.swift
dnzxy: Removed, thanks.

antoinekh:
Hello @dnzxy 
Is it possible to consider the option of adding protein and fat via the add carbs shortcut also?

MikePlante1:
> Hello @dnzxy Is it possible to consider the option of adding protein and fat via the add carbs shortcut also?

@antoinekh You already can:


https://github.com/user-attachments/assets/4b94bdce-747d-4a2e-8a8b-98392aaaddd0

antoinekh:
> > Hello @dnzxy Is it possible to consider the option of adding protein and fat via the add carbs shortcut also?
> 
> @antoinekh You already can:
> 
>  ScreenRecording_03-03-2026.06-13-09_1-compressed.mp4

Thanks for the quick answer

I made the PR to update the Trio docs to include them.
https://github.com/nightscout/trio-docs/pull/221

kingst (approved):
Approving after testing

Issue #948, opened 2026-01-25
Carbs entered on Apple Watch missing from Tidepool and AH


snickerdoodleparent:
Carbs entered on Apple Watch missing from Tidepool and Apple Health. 

In Nightscout total carbs are 180g for this particular day. 10:30am 25g by watch, 12:00pm 37g by watch, etc smaller amounts given throughout the day.
<img width="1293" height="393" alt="Image" src="https://github.com/user-attachments/assets/3e63dd03-95ff-4228-99f3-ab80849e6f23" />

In Tidepool total carbs for the day are 85g - all the watch remote carbs are missing. Remote carbs sent by Loop Follow at 7:45pm show up.
<img width="1036" height="792" alt="Image" src="https://github.com/user-attachments/assets/8493add6-fc02-4b02-ad60-46e9b5a8ec88" />

Dan asked about checking in Apple Health to see if they show up there as well. They did not.

To reproduce:
1. Enter carbs on watch, everything boluses and calculates correctly.
2. Check AH and Tidepool and only Trio app entered carbs show in total 
3. NS or Trio show all carbs regardless of how they were entered.

Hardware: 3rd Gen SE 
OS Version: 26.2
Apple watch: SE 2nd Gen
Watch OS 26.2
Omnipod Dash & G7
Trio dev:  v0.6.0 63bd24b

dnzxy:
Completed with merge of #957

PR #949, opened 2026-01-25
Apple watch carbs uploaded to Tidepool and Apple Health

snickerdoodleparent:
to solve Carbs entered on Apple Watch missing from Tidepool and AH [#948](https://github.com/nightscout/Trio/issues/948)

Dan suggested these lines, they solve the problem here on our phone

PR #957, opened 2026-01-30
Fix Watch Carb Entry Propagation to Apple Health and Tidepool

dnzxy:
* Adds `isUploadedToHealth` flag
* Adds `isUploadedToTidepool` flag
* Fixes issues where carb entries from Apple Watch were no propagated to third party services
* Addresses  #948


kingst (approved):
Overall LGTM, straightforward change that follows existing patterns (`isUploadedToNS`).

[optional] It looks like there is code code duplication in the code surrounding the fix, it might be a good opportunity to put this duplicate code into a helper function, but totally optional.

I didn't test this change, this review is based on reading code.

dnzxy:
I‘ll note down a refactor in a different issue. Let’s move the fix forward. It‘s been tested in-vivo by the co-author‘s child (see #949 ).

Let’s go ahead and merge this one.

PR #958, opened 2026-01-30
Adjust various setting thresholds

dnzxy:
* Changes various min &/or max bounds
   * MaxIOB from max. 20 U to 30 U
   * Logarithmic Adjustment Factor from max. 150% to 300%
   * Max SMB and SMB UAM Basal Minutes from min. 30min to 15min
   * High threshold from a never-reachable max. 500 mg/dL to 400 mg/dL
 * (Possibly) to be cont'd


kingst (approved):
LGTM

Issue #969, opened 2026-02-11
Fat & Protein order is inconsistent


MikePlante1:
## Describe the bug
Sometimes protein is listed first, other times fat is listed first.

## To Reproduce
Look through the app at all the fat and protein references and notice that the order is not consistent

## Expected behavior
Order should be consistent

## Screenshots

| . | . |
|---|---|
| ![Image](https://github.com/user-attachments/assets/4dd45cac-94e6-4e4b-8d4f-f20aefe7f341) | ![Image](https://github.com/user-attachments/assets/9193435e-34c9-4712-8906-5356542597a9) |
| ![Image](https://github.com/user-attachments/assets/d1ce9f83-d251-450d-aae3-7ab0d040cd91) | ![Image](https://github.com/user-attachments/assets/cebda44e-903b-44a3-93b7-28a6439edce3) |

### Trio Version:
* Version Number: 0.6.0.48
* Repo: nightscout/Trio

## Additional context

I think it should be Carbs > Fat > Protein. This matches the order of the FPU acronym as well as how it's written out in the app everywhere (`Fat and Protein`). Fat also has a higher conversion rate to carbs in the Warsaw method used in Trio.

Fixing this might cause confusion for users that are used to entering protein first and muscle memory can be hard to retrain.

g5WS:
Has there been any discussion about a popup that runs the first time you use Trio after an upgrade?  This ticket’s changes might merit an escalated level of notification.

dnzxy:
> Has there been any discussion about a popup that runs the first time you use Trio after an upgrade? This ticket’s changes might merit an escalated level of notification.

Yes but that is really no small task to implement. Swift has a whole library for that included, https://developer.apple.com/documentation/tipkit/.

PR #972, opened 2026-02-12
Fix and improve the Add Carbs shortcut

### LINKED ISSUE #967: Shortcuts action converts grams of carbs to time units when grams > 59.
v0.6.0.48

Test # 1: Try to bolus for 59 grams of carbs.

Image
Image

It works. 

Test # 2: Try to bolus for 60 grams of carbs.

Image
Image

The 60 is converted to 1:00 hours as if the 60 is minutes.

### LINKED ISSUE #968: UI: Shortcuts action has inconsistencies
v0.6.0.48

Observations: 
- The shortcut asks permission to add carbs, yet adds carbs, protein and fats
- The shortcut rounds down when asking permission ("Add 1 grams of carbs?"), but rounds up when confirming what was done ("Added 2 grams of carbs and...").
- The confirmation message uses the order carbs, fats, protein while Trio uses carbs, protein, fat.
- Units are always stated as grams (plural), even when there is only one.

_[Edit: 1.5 g isn't the clearest example. If you submit 7.8 g of carbs, it will display 7 g on the permission requesting dialogue, and 8 g on the confirmation dialogue.]_

Image
Image
Image

MikePlante1:
Address the following Issues:
* #967 
* #968 

This PR updates the Add Carbs shortcut to use `Int` instead of `Double` for the macronutrients. This addresses both Issues by removing the need for rounding as well as remove the need to format it for a String.

This PR also increases the maximum values of macronutrient entries to 300, which is the top of the range allowed for the Max Carbs/Fat/Protein settings in Trio. The shortcut also now fails with an error message and does not add anything if the users enters over the max amount. This is safer than before this PR where the carb entry is just quietly lowered to the Max Carbs setting and Max Fat/Protein aren't checked at all.

This PR also adds the Fat and Protein entries to the confirmation message.

### DIFF
--- a/Trio/Sources/Shortcuts/Bolus/BolusIntent.swift
+++ b/Trio/Sources/Shortcuts/Bolus/BolusIntent.swift
@@ -3,7 +3,7 @@ import Foundation
 import Intents
 import Swinject
 
-@available(iOS 16.0,*) struct BolusIntent: AppIntent {
+struct BolusIntent: AppIntent {
     // Title of the action in the Shortcuts app
     static var title = LocalizedStringResource("Enact Bolus")
 
--- a/Trio/Sources/Shortcuts/Bolus/BolusIntentRequest.swift
+++ b/Trio/Sources/Shortcuts/Bolus/BolusIntentRequest.swift
@@ -2,7 +2,7 @@ import Combine
 import CoreData
 import Foundation
 
-@available(iOS 16.0,*) final class BolusIntentRequest: BaseIntentsRequest {
+final class BolusIntentRequest: BaseIntentsRequest {
     func bolus(_ bolusAmount: Double) async throws -> String {
         var bolusQuantity: Decimal = 0
         switch settingsManager.settings.bolusShortcut {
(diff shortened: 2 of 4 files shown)


MikePlante1:
@g5WS if you'd like to give this a test to see if it addresses the two issues you opened about the Add Carbs shortcut?

g5WS:
Hey @MikePlante1 , I’m a little uncertain if I have pulled things correctly. (I feel like I stumble across a new method to pull code each time I attempt to test something. 🙃)

https://github.com/g5WS/Trio/pull/6

I see decimals still being accepted, no change in “1 grams of carbs”, only being asked to confirm carbs and not fat/protein… did I fumble the ball here?

MikePlante1:
> [g5WS#6](https://github.com/g5WS/Trio/pull/6)
> 
> I see decimals still being accepted, no change in “1 grams of carbs”, only being asked to confirm carbs and not fat/protein… did I fumble the ball here?

You have to merge your PR before you build.

g5WS:
> You have to merge your PR before you build.

Ah. :-)

Yes, now that I have built correctly, I see the changes. :) Nice job!  Only outstanding issue I see is the fat/protein order issue, but I recognize that will come via #969 and without changing the shortcut action. 

👍👍

dnzxy (approved):

Issue #966, opened 2026-02-05
Rearrange Shortcuts action to enhance safety


g5WS:
v0.6.0.48

(Sorry, @avouspierre!) 

The Trio app organizes the nutrients as: carbs, protein, fat.

![Image](https://github.com/user-attachments/assets/a0d697fc-752a-49c2-80b9-be7f28f1108b)

The Shortcuts action organizes the nutrients as: carbs, fat, protein.

![Image](https://github.com/user-attachments/assets/7f440391-a92c-45c8-af6f-cbd331ae20bf)

As this re-ordering is unexpected/unintuitive, it may be safer to reflect the nutrients in the order Trio uses.

(Also, it might be nice to change the L in "Confirm Before logging" to a capital letter. :) )

PR #973, opened 2026-02-13
Align the order of Fat and Protein throughout the app

MikePlante1:
Addresses this Issue:
- #969

### DIFF
--- a/Trio/Resources/InfoPlist.xcstrings
+++ b/Trio/Resources/InfoPlist.xcstrings
@@ -457,6 +457,18 @@
         }
       }
     },
+    "NSCalendarsFullAccessUsageDescription" : {
+      "comment" : "Privacy - Calendars Full Access Usage Description",
+      "extractionState" : "extracted_with_value",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "new",
+            "value" : "To create events with BG reading values, so that they can be viewed on Apple Watch and CarPlay"
+          }
+        }
+      }
+    },
     "NSCalendarsUsageDescription" : {
       "comment" : "Privacy - Calendars Usage Description",
       "extractionState" : "extracted_with_value",
--- a/Trio/Sources/Helpers/PropertyPersistentFlags.swift
+++ b/Trio/Sources/Helpers/PropertyPersistentFlags.swift
@@ -23,4 +23,7 @@ final class PropertyPersistentFlags {
     @PersistedProperty(key: "diagnosticsSharing") var diagnosticsSharingEnabled: Bool?
 
     @PersistedProperty(key: "lastCleanupDate") var lastCleanupDate: Date?
+
+    // TODO: This flag can be deleted in March 2027. Check the commit for other places to cleanup.
+    @PersistedProperty(key: "hasSeenFatProteinOrderChange") var hasSeenFatProteinOrderChange: Bool?
 }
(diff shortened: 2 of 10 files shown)


dnzxy:
I‘m not sure if muscle memory could potentially cause issues here and cause unwanted dosing. What do others think? @kingst ?

dnzxy:
Should this PR also include a change to address #966 too?

MikePlante1:
> Should this PR also include a change to address #966 too?

How does it not address that issue?

g5WS:
> > Should this PR also include a change to address #966 too?
> 
> How does it not address that issue?

I think the shortcut is already in CFP order, whereas Trio’s main screen is in CPF order, and Mike has proposed adopting CFP throughout, no?

dnzxy:
If it already is in CFP order, why is there an issue #966 ?

dnzxy:
> > Should this PR also include a change to address #966 too?
> 
> How does it not address that issue?

It doesn’t include a change to the shortcut and I haven’t double checked the current order, hence my question.

g5WS:
@dnzxy :  Here’s my understanding:

I proposed #966 when I observed the main screen was in CPF order and the shortcut action was CFP. I proposed updating the shortcut to CPF order to reflect the main screen. 

Then Mike did a deeper dive and observed the app really has no standard order: it uses both CFP and CPF.  Mike proposed standardizing CFP order, which was the goal of this ticket. (This would affect the main screen, which is why I asked about a first-run splash screen.)

The shortcut action no longer needs to be edited because it’s already in CFP order. #966 will be obsolete once Mike’s PR is merged.

dnzxy:
Thanks for adding some context, very much appreciated.
Admittedly, as a user who doesn’t use FPU nor shortcuts these things aren’t super readily available in my brain 😂 

I‘m still a bit unsure whether muscle memory and just moving inputs around that indirectly lead to insulin dosing can / will cause issues for the subset of users that uses FPU, so happy to hear if this is a concern others rate as very-low risk, or whether we need to wait for this PR to land until there is a "better" way to notify users about it in-app? I don’t have a perfect answer to it. 

I guess we can close #966 since it already has the correct order (and the issue calls for shortcuts to be altered), and this PR here will align everything else.

MikePlante1:
How about something like this? https://github.com/MikePlante1/Trio/compare/fat-protein...MikePlante1:Trio:fat-protein-warning?expand=1 (Claude assisted, for transparency)

* It only displays if you have FPU enabled.
* Once you tap the `x` a flag is set and the banner goes away forever.
* Flag is also set in onboarding, so the banner won't be shown to users who never saw the old order.
* Flag and banner code could be removed in an update after a year.

| dark | light |
|---|---|
|  |  |

g5WS:
> How about something like this? 

I think this is great, and I really appreciate the thoughtful appearance conditions!

If it is possible to include images, I think it would enhance clarity and comprehension to show images comparing before and after. I can empathize that this would really bump the complexity, if only for the translation issues.

dnzxy:
I'm also okay with this. Maybe add the word "input" as in "The order of Fat and Protein inputs has changed." and maybe make the text slightly larger, even if it becomes two rows, but if you add this one here, this PR can go through 👍 
Good idea 👍

MikePlante1:
Updated and pushed to this PR'd branch:

dnzxy (approved):

dnzxy:
Approving (and subsequently merging this) with the banner change.
Thank you both (@g5WS, @MikePlante1 ) for bringing this up and handling it.

PR #980, opened 2026-02-18
Fix: Delete Glucose entries in Nightscout API

ryceg:
This PR amends the Nightscout API to correctly target the entries collection, and uses the `_id` for discrimination since Trio's glucose entries have an `_id` instead of an `id`


kristinkand:
Hi, after @dnzxy nudge I cloned @rhysgg fix/delete-glucose-ns branch into mine and built it. But after waiting for two loops, the old bg still shows on NS after deleting it from Trio.

For testing purpose we'd entered a manual bg this morning since we then see a big red dot on NS.
Did I miss smth?

dnzxy:
@kristinkand this PR should fix deletions of manual glucose readings of NS. Tbf, I'm not entirely sure if that was actually broken before.

What is still "missing" is deletion of normal readings (`sgv`s). 

I believe adding the following snippet to `Trio/Sources/Services/Network/Nightscout/NightscoutAPI.swift`
```swift
func deleteGlucose(withId id: String) async throws {
        var components = URLComponents()
        components.scheme = url.scheme
        components.host = url.host
        components.port = url.port
        components.path = "/api/v1/entries/"
        components.queryItems = [
            URLQueryItem(name: "find[_id][$eq]", value: id)
        ]

        guard let url = components.url else {
            throw URLError(.badURL)
        }

        var request = URLRequest(url: url)
        request.allowsConstrainedNetworkAccess = false
        request.timeoutInterval = Config.timeout
        request.httpMethod = "DELETE"

        if let secret = secret {
            request.addValue(secret.sha1(), forHTTPHeaderField: "api-secret")
        }

        let (_, response) = try await URLSession.shared.data(for: request)

        guard let httpResponse = response as? HTTPURLResponse, (200 ... 299).contains(httpResponse.statusCode) else {
            throw URLError(.badServerResponse)
        }

        debugPrint("Delete glucose from NS successful for entry with ID \(id)")
    }
```

and this snippet to `Trio/Sources/Services/Network/Nightscout/NightscoutManager.swift`
```swift
func deleteGlucose(withID id: String) async {
        guard let nightscout = nightscoutAPI, isUploadEnabled else { return }

        do {
            try await nightscout.deleteGlucose(withId: id)
        } catch {
            debug(
                .nightscout,
                "\(DebuggingIdentifiers.failed) Failed to delete CGM Glucose Reading from Nightscout with error: \(error.localizedDescription)"
            )
        }
    }
``` 
plus adding `func deleteGlucose(withID id: String) async` to the protocol definition in the top of the file

as well as these changes to the `History` module:
`Trio/Sources/Modules/History/HistoryProvider.swift`
```swift
func deleteGlucoseFromNightscout(withID id: String) {
            Task.detached { [weak self] in
                guard let self = self else { return }
                await self.nightscoutManager.deleteGlucose(withID: id)
            }
        }
```

`Trio/Sources/Modules/History/HistoryDataFlow.swift`
```swift
// add to protocol pattern
    func deleteGlucoseFromNightscout(withID id: String)
```

In `Trio/Sources/Modules/History/HistoryStateModel.swift` find `func deleteGlucoseFromServices(_ treatmentObjectID: NSManagedObjectID) async` and add/find
```swift
// Delete from Nightscout
                    if let id = glucoseToDelete.id?.uuidString {
                        self.provider.deleteGlucoseFromNightscout(withID: id)
                        self.provider.deleteManualGlucoseFromNightscout(withID: id)
                    }
```

 should probably solve this :) 

If you feel confident giving this a try, please do. 
Otherwise, @ryceg or I can take a look and add it as a commit.

--- 
Edit to fix `DateTable`; module is now called `History` (https://github.com/nightscout/Trio/tree/dev/Trio/Sources/Modules/History)

kristinkand:
I rebuild successfully after editing the different files. However the red dot of the manual bg is still there.
Once again, we added a manual bg in Trio (same as CGM bg) which shows with a red dot in NS. Afterwards we deleted it from Trio, but it's still showing in NS after several loops.

kristinkand:
One more comment: After the helpful hand from @dnzxy , I reverted the first two commits @ryceg , but left in place all the "snippets" as posted by @dnzxy . Now when I delete a manual bg in Trio, it's being deleted from NS too (refresh NS or wait for a loop).
So, only the deletion of "normal" bgs from NS is now missing. How do we fix that?

kristinkand:
I attempted to build after your last commits  @ryceg However every build comes back with two errors: missing url path to submodule..xdrip and something about homebrew.
To make sure I'd cloned the latest version, I nuked my earlier copies of this branch and pulled down a new copy.

kristinkand:
I built again, now it builds successfully after the removal of the submodule. However the normal bg values are still not deleted from NS. If I build without @dnzxy's suggestions, nothing is being deleted neither manual nor normal bg values.

dnzxy:
Has anyone been able to test this since the last changes?

dnzxy:
I'd very much appreciate some testing on this. 
@kristinkand (ideally not in-vivo 😅 ) or maybe @marionbarker or @bjornoleh ?

dnzxy:
I tested this locally with a test NS site and a Trio build.

This now lets you delete both manual glucose readings and sensor readings. 
It also addresses #1099 by changing how we upload manual glucose readings.
Before this PR, we uploaded manual glucose readings into `treatments` with `type: "BG Check"` and a `glucose: xxx` .
Now, we upload manual readings into `entries` with `type: "mbg"` and `mbg: xxx`.

Feel free to test both glucose uploads and deletions.

kristinkand:
It's working! I tested to delete both manual and cgm bg and both went away.
Important for LF users: the main display doesn't show the deletion, but the NS screen shows it.

dnzxy:
I guess for LF users, LF would have to refetch `entries` and then make a check for delta / removed elements, then update its storage and re-render the chart — just pinging @bjorkert so he‘s aware. We should probably be recording this in an issue over in the LF repo.

bjorkert:
LoopFollow always performs a full refresh. Waiting for the next glucose value refresh or pulling down on the glucose value to force a refresh should resolve it.

kristinkand:
I already did and created a PR614

kristinkand:
> refresh

Nope, tried it and waited because of the anticipation that the next loop cycle would update it. But it didn't. Any way, just wanted to point it out!

bjorkert:
Then the value is still in Nightscout. It’s possible that Nightscout receives BG data both from Trio and via the dexcom bridge, and Trio likely only deletes the entry it created. Just a guess though, I haven’t looked into this PR.

dnzxy:
@bjorkert This PR deletes values based on date or id (external sync identifier) from the `entries` collection of NS. If there are duplicates we cannot grab them.

bjorkert:
I have requested a Nightscout token to verify, I'll post my findings here later.

kristinkand:
Fixed! The values are deleted from NS.
Dexcom share values are shown on the main screen of LF and prioritized compared to NS, while NS values are just shown in the NS screen and those are lacking the deleted bg value. 🥳

dnzxy (approved):

bjorkert (approved):

PR #984, opened 2026-02-23
feature: hide SMB label

bastiaanv:
To allow user to clean up the main chart, I believe it is good to give users an extra option to hide the amount label above the SMB's.

Screenshots:

IMG_1411
IMG_1410

### DIFF
--- a/Trio/Sources/Modules/Home/HomeStateModel.swift
+++ b/Trio/Sources/Modules/Home/HomeStateModel.swift
@@ -76,7 +76,7 @@ extension Home {
         var displayXgridLines: Bool = false
         var displayYgridLines: Bool = false
         var thresholdLines: Bool = false
-        var showSmbLabel: Bool = true
+        var bolusDisplayThreshold: BolusDisplayThreshold = .allUnits
         var hours: Int16 = 6
         var totalBolus: Decimal = 0
         var isLoopStatusPresented: Bool = false
@@ -405,7 +405,7 @@ extension Home {
             eA1cDisplayUnit = settingsManager.settings.eA1cDisplayUnit
             displayXgridLines = settingsManager.settings.xGridLines
             displayYgridLines = settingsManager.settings.yGridLines
-            showSmbLabel = settingsManager.settings.showSmbLabelInChart
+            bolusDisplayThreshold = settingsManager.settings.bolusDisplayThreshold
             thresholdLines = settingsManager.settings.rulerMarks
             showCarbsRequiredBadge = settingsManager.settings.showCarbsRequiredBadge
             forecastDisplayType = settingsManager.settings.forecastDisplayType
@@ -670,7 +670,7 @@ extension Home.StateModel:
         displayXgridLines = settingsManager.settings.xGridLines
         displayYgridLines = settingsManager.settings.yGridLines
         thresholdLines = settingsManager.settings.rulerMarks
-        showSmbLabel = settingsManager.settings.showSmbLabelInChart
+        bolusDisplayThreshold = settingsManager.settings.bolusDisplayThreshold
         showCarbsRequiredBadge = settingsManager.settings.showCarbsRequiredBadge
         forecastDisplayType = settingsManager.settings.forecastDisplayType
         cgmAvailable = (fetchGlucoseManager.cgmGlucoseSourceType != CGMType.none)
--- a/Trio/Sources/Modules/Home/HomeStateModel.swift
+++ b/Trio/Sources/Modules/Home/HomeStateModel.swift
@@ -677,7 +677,6 @@ extension Home.StateModel:
         displayPumpStatusHighlightMessage()
         displayPumpStatusBadge()
         setupBatteryArray()
-        setupInsulinArray()
         Task {
             await setupCGMSettings()
         }
--- a/Trio/Sources/Modules/Home/HomeStateModel.swift
+++ b/Trio/Sources/Modules/Home/HomeStateModel.swift
@@ -76,7 +76,7 @@ extension Home {
         var displayXgridLines: Bool = false
         var displayYgridLines: Bool = false
         var thresholdLines: Bool = false
-        var showSmb: Bool = true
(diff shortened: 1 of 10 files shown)


bastiaanv:
We could also go the iAPS route and use a minimum amount before showing the text. For example, only show if bolus/SMB >= 0.3U

dnzxy:
Personally, not a fan of adding more settings and I know of ~3 users out if ~5k that want this type of option, but if the team thinks this is a good addition, happy to add it. 

Edit to add: 
- I have concerns about just hiding _all_ SMB labels. Small boli would/can also obfuscate chart clarity, while SMBs can be very large, too, and hiding that information seems dangerous. 
- If this change is considered by the team, I have some remarks for the code, but will not add a review is this is a non-starter.

g5WS:
Love it, with @dnzxy’s addendum to continue to label large boli. 🥳. My SMB labels often overlap to the point where it’s no longer meaningful information. 

Were you counting me amongst the three people you know, @dnzxy? 🙃

MikePlante1:
I had the same thought as @dnzxy about hiding based on SMB vs Manual and think that would not be good.

I wouldn’t use it, but I’m okay with an option for only displaying a label when over a certain threshold.

Nightscout offers this, for reference:

IMG_9152

MikePlante1:
Another option to improve readability is rotating the label. 90 is best at reducing the overlap but I suppose sideways text probably isn't the easiest to read.

I'm not necessarily suggesting we use this option, just putting it out there. (edit to clarify: obviously the slider would not be part of the option, it's just to make exploring the option easier)

| 0 | 33 |
| :---: | :---: |
|  |  |
| 45 | 90 |
|  |  |

g5WS:
Building on Mike’s ideas, what about adding the SMBs to the data popup?  That would at least let someone clarify directly.  

image

bastiaanv:
I liked the Nightscout variant, I didn't have the inspiration for the helper text

Simulator Screenshot - iPhone 17 Pro - 2026-03-01 at 22 15 13

Trio/Sources/Modules/Home/HomeStateModel.swift
```diff
@@ -668,12 +670,14 @@ extension Home.StateModel:
...
         displayYgridLines = settingsManager.settings.yGridLines
         thresholdLines = settingsManager.settings.rulerMarks
+        bolusDisplayThreshold = settingsManager.settings.bolusDisplayThreshold
         showCarbsRequiredBadge = settingsManager.settings.showCarbsRequiredBadge
         forecastDisplayType = settingsManager.settings.forecastDisplayType
         cgmAvailable = (fetchGlucoseManager.cgmGlucoseSourceType != CGMType.none)
         displayPumpStatusHighlightMessage()
         displayPumpStatusBadge()
         setupBatteryArray()
+        setupInsulinArray()
```
marv-out: This addition shouldn't really be necessary because we are already subscribing to the PumpEventStored events and we are basically only working on the display logic for smbs.

marv-out (commented):
One small comment, besides that LGTM

dnzxy:
Please add a whitespace before the units for the dropdown texts. 
Please capitalize the last selection to "Show All Labels" 
Rest LGTM.

marv-out:
When testing I've found that the switcher doesn't work. I've debugged it and it was actually pretty simple 😂 the condition for the annotation in the Insulin View needs to be the other way round. Before it was ``if bolusDisplayThreshold.rawValue > amount as Decimal`` and it should ofc be ``if amount as Decimal > bolusDisplayThreshold.rawValue ``

bastiaanv:
Thank you for the feedback!

I updated the `Show all labels` to `All Labels`, because the `Show All Labels` didn't fit horizontally on my iPhone 16 Pro.
Also add the hint text & synced with dev

marv-out (approved):

dnzxy:
> Thank you for the feedback!
> 
> I updated the `Show all labels` to `All Labels`, because the `Show All Labels` didn't fit horizontally on my iPhone 16 Pro. Also add the hint text & synced with dev

Then maybe just "Show All" ? 
It will get longer depending on language. 
German would be "Alle anzeigen" for example.

bastiaanv:
> Then maybe just "Show All" ? It will get longer depending on language. German would be "Alle anzeigen" for example.

Done!

Trio/Sources/Modules/UserInterfaceSettings/View/UserInterfaceSettingsRootView.swift
```diff
@@ -191,6 +191,48 @@ extension UserInterfaceSettings {
...
+                            .lineLimit(nil)
+                            Spacer()
+                            Button(
+                                action: {
+                                    hintLabel = String(localized: "Bolus Display Threshold")
+                                    selectedVerboseHint =
+                                        AnyView(
+                                            VStack(alignment: .leading) {
+                                                Text(
+                                                    "Choose whether to display the amount above every bolus indicator or just for a bigger bolus amount"
```
dnzxy: Let's change this to
> This setting controls which bolus amount labels are shown on Trio’s main chart. Boluses appear as blue upside-down triangles, with a number showing the amount. Depending on the option you choose, only boluses at or above that amount will show a label. For example, if you choose ‘0.5 U and over’, only boluses of 0.5 U or more will show a label.

dnzxy (commented):
Minor adjustment for the hint.
Rest LGTM and works well!

Trio/Sources/Modules/UserInterfaceSettings/View/UserInterfaceSettingsRootView.swift
dnzxy: Quickly pushed this myself. Never mind.

dnzxy (approved):

g5WS:
Oh, this feature is _so nice_, @bastiaanv.  I appreciate your efforts! 👍🤩🙏

PR #987, opened 2026-02-24
Chore: Update submodules to latest versions

marionbarker:
## Purpose

* Add some helper scripts to Trio to assist in this chore
* Update submodules to the latest version for Trio submodules using the downstream forks found at loopandlearn

## Method

1. Make sure the loopandlearn forks are all up to date
   * This step is done when updating LoopWorkspace
2. Run the new script: ./scripts/update_submodules_trio.sh
   * The script was copied and modified from the similar script used by LoopWorkspace

### DIFF
--- a/scripts/define_common_trio.sh
+++ b/scripts/define_common_trio.sh
@@ -12,16 +12,19 @@
 # define the TRIO_PROJECTS used by Trio where for Trio the .gitmodules points
 #   to the downstream fork of loopandlearn in all cases
 #
+# The only submodule that needs a special trio branch (at this time) is LoopKit
+#    put that repository first in the list
+#
 # The scrips in LoopWorkspace are used to update translations and make
 #   sure the loopandlearn branches are up to date
 # Even though MedtrumKit and EversenseKit are not yet part of dev
 #   it does no harm to update those repositories using this script
 TRIO_PROJECTS=( \
+    loopandlearn:LoopKit:trio \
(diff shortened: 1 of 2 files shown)


marionbarker:
## More Details

There are modifications for two submodules which affect more than just translations.

### DanaKit

@bastiaanv has been continuing to work on DanaKit
* this update includes a rebase
* the DanaKit dose entries now support the pump details expected by Loop
* the User Interface was updated.
* in addition, the responsibility for translations was returned to the Loop lokalise project, so the crowdin.yml was removed

### LoopKit

There was a formatting change in LoopKit that makes the detailed pump events (only reported in Loop) easier to read.
The other modifications are just translations changes.

marionbarker:
## Test

I built this on a SE 3rd gen test phone running iOS 26.3 and connected to a Dana rPi simulator.
I have a green loop and successfully entered a meal and boluses on the test phone.

I acknowledged the message that the order of fat and proteins was modified.

bastiaanv:
DanaKit SHA looks good! I am currently running this sha with no issues

mountrcg:
@marionbarker could you elaborate on when to use it? I have so many issues handling the submodules across 3 repos. When locally switching between different repo branches this gets all messy. I try to not do it (like being in Trio branch and then locally switching to a Tai branch) but it happens.

dnzxy (approved):

PR #1001, opened 2026-03-06
Refactor Glucose Smoothing

### LINKED ISSUE #985: Refactor & Refine Trio's Glucose Smoothing

Trio currently applies glucose smoothing in a way that isn’t aligned with AAPS and may introduce unintended behavior (e.g., using a processed value beyond purely algorithmic input). The current Savitzky–Golay approach also raises safety/application concerns for an AID context and makes it harder to reason about which value is “truth” in UI/alerts and what’s fed into oref vs. what's displayed by the CGM app (or xDrip4iOS).

* Refine Trio glucose smoothing to match AndroidAPS behavior by switching to _Exponential Smoothing_ and removing the Savitzky–Golay filter.
* Refactor data handling so the raw CGM reading remains the only value used for UI display and alerts, while the smoothed value is stored alongside it and used exclusively as algorithmic input for oref.
* Update the main chart display accordingly.
* Add unit tests to lock in the behavior.

* Keep Savitzky–Golay but restrict its use to algorithm input only — rejected due to safety/application concerns and mismatch with AAPS behavior.
* Disable smoothing entirely. Simpler, but loses a known beneficial input conditioning step that AAPS already supports, especially for Dexcom G7 users.

## Additional context

* Goal is behavioral parity with AAPS smoothing to reduce divergence and ambiguity. Reference implementation: [AndroidAPS ExponentialSmoothingPlugin](https://github.com/nightscout/AndroidAPS/blob/master/plugins/smoothing/src/main/kotlin/app/aaps/plugins/smoothing/ExponentialSmoothingPlugin.kt).
* Active Blocker for releasing Trio `dev` (`v0.6.0.x`) to `main`.

## Technical Details

* Replace smoothing implementation with Exponential Smoothing aligned to AAPS plugin logic/parameters.
* Remove Savitzky–Golay filter path entirely.
* Storage refactor:

  * Persist raw CGM value unchanged.
  * Persist smoothed value as an additional field (not replacing raw).
  * Ensure UI/alerts always consume raw CGM values only.
  * Ensure oref input fetches smoothed values only (explicit codepath).
* Update main chart rendering to reflect “raw-only” display semantics (and avoid accidentally charting smoothed as CGM).
* Add unit tests covering:

  * Parity with AAPS smoothing output for known sequences
  * Storage correctness (raw + smoothed)
  * Correct routing (UI/alerts use raw; oref uses smoothed)

## User Impact

Primarily improves safety/clarity and reduces confusion: users always see and are alerted on real CGM values, while the algorithm gets the intended conditioned signal. Also reduces behavioral drift vs. AAPS and makes future troubleshooting easier.

dnzxy:
## Summary

Refactors Trio's glucose smoothing to align with AndroidAPS by replacing the Savitzky–Golay filter with exponential smoothing and separating raw CGM values from algorithmic input.

Raw CGM glucose remains the source of truth for UI and alerts, while smoothed values are stored alongside it and used only as algorithm input when smoothing is enabled.

## Changes

* Replace Savitzky–Golay smoothing with AAPS-style exponential smoothing
* Add `smoothedGlucose` field to `GlucoseStored`
* Store smoothed values alongside raw CGM readings
* Ensure:
  * UI + alerts use raw CGM values
  * oref uses smoothed values when enabled
* Introduce `AlgorithmGlucose` model for algorithm input serialization
* Update chart to display smoothed values when smoothing is enabled
  * Display trend line for smoothed values while keeping raw CGM points 
  * Add smoothed value to selection popover
* Add unit tests for smoothing behavior and algorithm routing
* Remove obsolete Savitzky–Golay implementation

| Popover | Chart w/ Trend |
|--------|--------|
| image | image |

Addresses #985


kristinkand:
I built it and I really like that it shows what the CGM says and the smoothed value when hovering over a bg value.👍
We have encountered two errors/problems: in history a lot of bgs have "vanished" replaced by two dashes and the main screen showed suddenly to grey vertical lines where had added manual bgs.
Additionally I don't see the grey line as shown in the above screenshots.

Screenshot 2026-03-07 at 09 52 32
Screenshot 2026-03-07 at 09 53 10

dnzxy:
Thanks @kristinkand , I must admit I didn’t check history. 
Let me see about your findings. Will try and push and update today or tomorrow. 

As for the chart, let me do another test, but if memory serves right we are not retroactively smoothing glucose, so I _think_ that’s why you are not seeing the smooth trend line mark for old values (?). Manual glucose readings definitely shouldn’t be showing this weird buggy plot in the chart.

kristinkand:
No rush @dnzxy -not traveling this week, so can build pretty quickly. :D

I don't know what to expect to see in the history chart. The smoothed values or the CGM values to be honest. The two dashes are kind of confusing. Edit: there used to be the arrows. 
Maybe to stay true to the main screen, if the history page shows the smoothed values it could show the stars instead of the -- and the arrow.

I built last night and turned it the smoothing on right away, so the chart should show the trend line mark at least throughout the night.

My T1D installed dev after those findings, but the -- didn't disappear.
Edit: I installed once more the glucose smoothing version and right away the arrow after the number in history was replaced by two dashes. Additionally I've noticed that a dash appears in front of the iob number like -,14U.That seems only the case while smoothing is enabled.

One more note: we need to update the text in settings about the smoothing, but that is easy peasy once it's in place. 

Very cool, thanks for your hard work!

Trio/Sources/APS/FetchGlucoseManager.swift
```diff
@@ -377,3 +351,168 @@ extension CGMManager {
         ]
     }
 }
+
+extension BaseFetchGlucoseManager: SettingsObserver {
+    /// Smooth glucose data when smoothing is turned on
+    func settingsDidChange(_: TrioSettings) {
+        if settingsManager.settings.smoothGlucose, !shouldSmoothGlucose {
+            glucoseStoreAndHeartLock.wait()
```
kingst: Is this getting called on from the `main` thread? If so, consider putting it on a background dispatch queue.

Trio/Sources/APS/FetchGlucoseManager.swift
```diff
@@ -377,3 +351,168 @@ extension CGMManager {
...
+        firstOrderSmoothed.reserveCapacity(validWindowCount + 1)
+
+        // Initialize with the oldest valid point (index validWindowCount - 1)
+        firstOrderSmoothed = [Decimal(Int(data[validWindowCount - 1].glucose))]
+
+        for i in 0 ..< validWindowCount {
+            let raw = Decimal(Int(data[validWindowCount - 1 - i].glucose))
+            let prev = firstOrderSmoothed[0]
+            let next = prev + firstOrderAlpha * (raw - prev)
+            firstOrderSmoothed.insert(next, at: 0)
```
kingst: In Swift this is a O(n) operation, making the algorithm O(n^2) overall https://github.com/swiftlang/swift/blob/2b8da2a4e352358cc4382392e9618fa99b12a275/stdlib/public/core/Array.swift#L1379

Trio/Sources/APS/FetchGlucoseManager.swift
```diff
@@ -377,3 +351,168 @@ extension CGMManager {
...
+        }
+
+        // Apply to the most recent `limit` readings (same behavior as Kotlin)
+        let limit = min(blended.count, data.count)
+        for i in 0 ..< limit {
+            let rounded = blended[i].rounded(toPlaces: 0) // nearest integer, ties away from zero
+            let clamped = max(rounded, minimumSmoothedGlucose)
+
+            data[i].smoothedGlucose = clamped as NSDecimalNumber
+            data[i].direction = .none
```
kingst: Does it make sense to re-calculate `direction` vs just clear it out? I don't know if this gets used in the UI or not but it could be something that people use when dosing or treating lows.

Trio/Sources/APS/FetchGlucoseManager.swift
```diff
@@ -377,3 +351,168 @@ extension CGMManager {
...
+            } catch {
+                debugPrint("Failed to save context after smoothing: \(error)")
+            }
+        }
+
+        let duration = Date().timeIntervalSince(startTime)
+        debugPrint(String(format: "Exponential smoothing duration: %0.04fs", duration))
+    }
+
+    private func applyExponentialSmoothingAndStore(
```
kingst: This function overall is a bit hard to follow since algorithmically it's going in reverse (e.g., `validWindowCount - 1 - i`). Consider reversing arrays and processing forward  then reversing back. I think that will be more efficient as well since your insert(at: 0) calls turn into appends.

Also, we should avoid using array indexing if possible, but I recognize that sometimes the code needed when avoiding array indexing is hard to follow.

Trio/Sources/APS/FetchGlucoseManager.swift
```diff
@@ -234,40 +240,17 @@ final class BaseFetchGlucoseManager: FetchGlucoseManager, Injectable {
         return Manager.init(rawState: rawState)
     }
 
-    private func fetchGlucose() async throws -> [GlucoseStored]? {
+    func fetchGlucose(context: NSManagedObjectContext) async throws -> [GlucoseStored]? {
         try await CoreDataStack.shared.fetchEntitiesAsync(
             ofType: GlucoseStored.self,
             onContext: context,
-            predicate: NSPredicate.predicateFor30MinAgo,
+            predicate: NSPredicate.predicateForOneDayAgoInMinutes,
```
kingst: Can you add a comment here that this needs to be as far back as the algorithm goes? If we change the how far back the algorithm fetches values (e.g., if we fix Autosens to use 36 hours of glucose values) this value will need to change.

kingst (commented):
Left a few comments / question, in general the code looks good. I'll test it out on my device today -- I'm excited for this to happen!

Trio/Sources/APS/FetchGlucoseManager.swift
dnzxy: This actually turned out to be a bug and I will get rid of this.
We only use the smoothed value for the algorithm, this never should have been cleared and was a mistake on my end. 
It's what is causing the issue Kristin reported with the `--` in History.

Trio/Sources/Modules/Home/View/Chart/ChartElements/GlucoseChartView.swift
```diff
@@ -32,46 +32,34 @@ struct GlucoseChartView: ChartContent {
...
                 }
-            } else {
-                PointMark(
+            }
+
+            if isSmoothingEnabled, let smoothedGlucose = item.smoothedGlucose {
+                LineMark(
                     x: .value("Time", item.date ?? Date(), unit: .second),
-                    y: .value("Value", glucoseToDisplay)
+                    y: .value("Value", smoothedGlucose as Decimal),
```
kingst: I believe that we need to convert this to mmol before displaying it

kingst (commented):
One more comment...

Trio/Sources/APS/FetchGlucoseManager.swift
dnzxy: Moved this to the process queue

Trio/Sources/Modules/Home/View/Chart/ChartElements/GlucoseChartView.swift
dnzxy: Fixed.

Trio/Sources/APS/FetchGlucoseManager.swift
dnzxy: Agree! Refactored the entire function to remove all index shenanigans and reduce complexity.

dnzxy:
> I don't know what to expect to see in the history chart. 

I erroneously broke listing the trend arrow. 
Currently, you do not see the smoothed value in history. I can totally put that in there…might actually be a good idea. 
Hold on a moment 😂

dnzxy:
There you go :) 

| Noisy Chart w/ Smoothing | History w/ Smoothing |
|--------|--------|
| image | image |

kingst (approved):
LGTM from a code perspective, I'll test it out today

kingst:
When testing on my phone everything appeared to be working from a algorithm perspective, but the smoothed values weren't being picked up by the UI. When I turned on smoothing I could see the forecasts change location and I could see the duration log so I know that the data was being smoothed, but the smoothing line and history entries weren't being displayed. When I clicked on a glucose value it shows a smoothed value of "0". Then I restated the app and it picked up the values that had been smoothed previously but wasn't updating when new values came in. I have a few screen shots that show the old smoothed values being displayed but not the new ones.

Also, the help string when I enabled smoothing still mentioned the old SG algorithm in the string

Screenshot 2026-03-08 at 10 16 15 AM
Screenshot 2026-03-08 at 10 18 50 AM

dnzxy:
The hint definitely needs changing. Thank you, good finding. Will update that.

The 0 values sound to me like writing the smoothed values (i.e., the update of the glucose entry stored in CoreData) fails, as optional NSDecimalNumber values are put out as 0 values instead of nil IIRC. 

I haven’t had this happen on a simulator when testing this out, I haven’t run this in-vivo though. I‘ll have to try and find the source of that issue / race condition 👀🫣

kristinkand:
" but the smoothing line and history entries weren't being displayed. When I clicked on a glucose value it shows a smoothed value of "0". Then I restated the app and it picked up the values that had been smoothed previously but wasn't updating when new values came in."
I noticed the same last night, but thought it was just me. Therefore I deleted my branch, created a new one based on @dnzxy glucose smoothing branch and build again. Since then the smoothed values come in and the helpline is updated.

Not sure what you mean with "0".!

[image](https://github.com/user-attachments/assets/e0e5507a-edca-4598-8bba-31b2558b298a)

image

Trio/Sources/APS/OpenAPS/OpenAPS.swift
```diff
@@ -98,8 +98,12 @@ final class OpenAPS {
     }
 
     // fetch glucose to pass it to the meal function and to determine basal
-    private func fetchAndProcessGlucose(fetchLimit: Int?) async throws -> String {
-        let results = try await CoreDataStack.shared.fetchEntitiesAsync(
+    func fetchAndProcessGlucose(
```
marv-out, line 101: I would suggest a slightly rewritten version with basically 3 differences:

1. making the CD fetch async and await it to avoid using 'performAndWait'
2. mapping within the context closure, JSON conversion outside to avoid the Sendable error we are getting at the moment
3. extracting handler to only create it 1x

```
func fetchAndProcessGlucose(
    context: NSManagedObjectContext,
    shouldSmoothGlucose: Bool,
    fetchLimit: Int?
) async throws -> String {
    // make it async and await it
    let results = try await CoreDataStack.shared.fetchEntitiesAsync(
        ofType: GlucoseStored.self,
        onContext: context,
        predicate: NSPredicate.predicateForOneDayAgoInMinutes,
        key: "date",
        ascending: false,
        fetchLimit: fetchLimit,
        batchSize: 48
    )

    // mapping within the context closure, JSON conversion outside
    let algorithmGlucose = try await context.perform {
        guard let glucoseResults = results as? [GlucoseStored] else {
            throw CoreDataError.fetchError(function: #function, file: #file)
        }

        // extracting handler to only create it 1x
        let roundingBehavior = NSDecimalNumberHandler(
            roundingMode: .plain,
            scale: 0,
            raiseOnExactness: false,
            raiseOnOverflow: false,
            raiseOnUnderflow: false,
            raiseOnDivideByZero: false
        )

        return glucoseResults.map { glucose -> AlgorithmGlucose in
            let glucoseValue: Int16
            if shouldSmoothGlucose, !glucose.isManual, let smoothedGlucose = glucose.smoothedGlucose {
                glucoseValue = smoothedGlucose.rounding(accordingToBehavior: roundingBehavior).int16Value
            } else {
                glucoseValue = glucose.glucose
            }
            return AlgorithmGlucose(
                date: glucose.date,
                direction: glucose.direction,
                glucose: glucoseValue,
                id: glucose.id,
                isManual: glucose.isManual
            )
        }
    }

    return jsonConverter.convertToJSON(algorithmGlucose)
}
```

Trio/Sources/APS/FetchGlucoseManager.swift
```diff
@@ -234,40 +240,20 @@ final class BaseFetchGlucoseManager: FetchGlucoseManager, Injectable {
...
             onContext: context,
-            predicate: NSPredicate.predicateFor30MinAgo,
+            // Predicate must cover at least the full glucose horizon used by downstream algorithm consumers.
+            // If autosens / oref / smoothing logic ever starts looking back further (e.g. 36h),
+            // this fetch window must be expanded accordingly.
+            predicate: NSPredicate.predicateForOneDayAgoInMinutes,
             key: "date",
-            ascending: false,
-            fetchLimit: 6
+            ascending: true, // fetch newest -> oldest
```
marv-out: I guess you are mistakenly using the ascending fetch order here. What it does right now is going back 24h and then fetching the next up to 350 glucose values due to the fetch limit. So basically this should be ok because we can only have 288 glucose values in the last 24h (+ maybe a few manual glucose values). So it reaches current time. But if you would lower the fetchLimit to e.g. 200 you wouldn't reach the current values. When fetching in an **ascending order** by "date" the **first** element will be **oldest** glucose value contrary to your comment.

marv-out: To add: I am seeing in the exponentialSmoothingGlucose func now that you were aware of the fetch order, so it was probably on purpose. Anyway, just wanna get sure here because it matters

Trio/Sources/APS/FetchGlucoseManager.swift
```diff
@@ -377,3 +354,180 @@ extension CGMManager {
...
+            }
+        }
+    }
+}
+
+extension BaseFetchGlucoseManager {
+    /// CoreData-friendly AAPS exponential smoothing + storage.
+    /// - Important: Only stores `smoothedGlucose`. UI/alerts should still use `glucose`.
+    ///
+    func exponentialSmoothingGlucose(context: NSManagedObjectContext) async {
```
marv-out, line 400: I would suggest the following

```
func exponentialSmoothingGlucose(context: NSManagedObjectContext) async {
        let startTime = Date()
        do {
            // get objectIDs
            let objectIDs = try await fetchGlucose(context: context)

            try await context.perform {
                // Load managed objects from object IDs
                // Filtering (isManual, date) already done at DB level in fetchGlucose
                let glucoseReadings = objectIDs.compactMap {
                    context.object(with: $0) as? GlucoseStored
                }

                guard !glucoseReadings.isEmpty else { return }

                // Static method call to avoid self-capture
                Self.applyExponentialSmoothingAndStore(
                    glucoseReadings: glucoseReadings,
                    minimumWindowSize: 4,
                    maximumAllowedGapMinutes: 12,
                    xDripErrorGlucose: 38,
                    minimumSmoothedGlucose: 39,
                    firstOrderWeight: 0.4,
                    firstOrderAlpha: 0.5,
                    secondOrderAlpha: 0.4,
                    secondOrderBeta: 1.0
                )

                try context.save()
            }

            let duration = Date().timeIntervalSince(startTime)
            debugPrint(String(format: "Exponential smoothing duration: %0.04fs", duration))
        } catch {
            debug(.deviceManager, "Failed to smooth glucose: \(error)")
        }
    }
```

And here`s why:

- it avoids the 2 Sendable warnings we are getting at the moment because an Array of glucoseStored is ofc not sendable, but the NSManagedObjectIDs are indeed, so its not only Core Data Thread safe but also compliant with Swift 6
- filters for "isManual" and "date" on db-level and not in Swift which is best practice for CD
- introduces a more robust error handling instead of using "try?"

fetchGlucose() would basically look like this:

```
func fetchGlucose(context: NSManagedObjectContext) async throws -> [NSManagedObjectID] {
        // Compound predicate: time window + non-manual + valid date
        let timePredicate = NSPredicate.predicateForOneDayAgoInMinutes
        let manualPredicate = NSPredicate(format: "isManual == NO")
        let datePredicate = NSPredicate(format: "date != nil")

        let compoundPredicate = NSCompoundPredicate(andPredicateWithSubpredicates: [
            timePredicate,
            manualPredicate,
            datePredicate
        ])

        let results = try await CoreDataStack.shared.fetchEntitiesAsync(
            ofType: GlucoseStored.self,
            onContext: context,
            // Predicate must cover at least the full glucose horizon used by downstream algorithm consumers.
            // If autosens / oref / smoothing logic ever starts looking back further (e.g. 36h),
            // this fetch window must be expanded accordingly.
            predicate: compoundPredicate,
            key: "date",
            ascending: true, // the first element is the oldest
            fetchLimit: 350
        )

        guard let glucoseArray = results as? [GlucoseStored] else {
            throw CoreDataError.fetchError(function: #function, file: #file)
        }

        return glucoseArray.map(\.objectID)
    }
```


and applyExponentialSmoothingAndStore() would become static which isn't a problem because we are not using any instance properties. Actually it is the correct design decision here

marv-out (dismissed):
Left a few comments and requested some changes. After code review I will test running it now

marv-out:
@kingst @kristinkand Would you mind testing it again with the few changes and improved error handling? Haven't experienced that problem myself yet

kristinkand:
> @kingst @kristinkand Would you mind testing it again with the few changes and improved error handling? Haven't experienced that problem myself yet

I rebuilt last night and all seems to be working fine. History shows values for each row, the main screen shows the cgm value and the smoothed value when hovering over it. Good job @marv-out

kingst:
@marv-out I'm still not seeing the smoothed values in the UI as they come up, but when I kill the app and restart it they do show up. I'm using a test phone, rpi pod simulator + nightscout CGM.

Also, I noticed that "smoothedGlucose" is both optional and has a default value, we probably don't want the default value.

marv-out:
Hm ok I will have another look

kingst:
@marv-out I just pushed some changes that I think address it...

type1berger:
I just want to say I love the new Glucose Smoothing! I love that the CGM reading shows exactly the same as my Dexcom app but in the background when you click on the graph you see the "Smoothing" difference. I think that's correct? LOL.. Well great job on this. So much cleaner!

pepijnjoost:
> I just want to say I love the new Glucose Smoothing! I love that the CGM reading shows exactly the same as my Dexcom app but in the background when you click on the graph you see the "Smoothing" difference. I think that's correct? LOL.. Well great job on this. So much cleaner!

I have to agree with this one!

marv-out:
Refactored the TreatmentsView to also reflect the new UI when smoothing is enabled

kingst (approved):
LGTM tested in vivo and works great

PR #1012, opened 2026-03-16
Crowdin translations and %% character fixes 2026-03-16

t1dude:
Crowdin translations and %% character fixes 2026-03-16


t1dude:
Failing tests:
    LocalizationTests.testNoStrayPercent()

Who set up this test? I'd like to know what it is looking for, because i went through the file looking for %% translated into %, and cannot find any more errors.

@MikePlante1 @dnzxy please have a look :)

MikePlante1:
@t1dude I can take a closer look later, but here’s what GitHub’s built-in CoPilot has to say about the failed test:

> The log reports two strings in the Vietnamese ("vi") localization: "If Half Basal Exercise Target is %@ %@, a Temp Target of %@ %@ uses an Autosens Ratio of 1.09. A Temp Target of %@ %@ use..."

PR #1030, opened 2026-03-25
Use segmented picker instead of radio buttons for Bolus Type in Treatment View

### LINKED ISSUE #977: UI: Represent all bolus types on food/bolus entry screen
v0.6.0.50

There are three different types of bolus: regular, reduced, and super.  However, we only represent two of them on screen, and we do so via radio buttons.  Using two radio buttons to represent three choices is an unexpected design, and could lead to confusion.

Image

Proposal: switch the radio buttons to a segmented to represent the three choices with three options on screen.  This UI would increase consistency as the app's other main screens also use segmented controls.

Could look like this:

1. Reduced bolus disabled, super bolus disabled: don't show segmented controls
2. Reduced bolus disabled, super bolus enabled: show [regular bolus | super bolus]
3. Reduced bolus enabled, super bolus disabled: show [reduced bolus | regular bolus]
4. Reduced bolus enabled, super bolus enabled: there may not be adequate space here. Can a segmented controls render text dynamically?  Could you show:
- [reduced | regular bolus | super] by default
- [reduced bolus | regular | super] when reduced bolus is selected
- [reduced | regular | super bolus] when super bolus is selected?

OR could use replace the text with some appropriate SF Symbols?  Maybe there's a gauge-like _trio_ that would represent less, average, and more? (Sorry, couldn't help the pun.)

MikePlante1:
Addresses:
- #977 

| Bolus Type | Reduced (OFF) | Reduced (ON) |
| :---: | :---: | :---: |
| Super Bolus (OFF) |  |  |
| Super Bolus (ON) |  |  |


Sjoerd-Bo3:
I like it! Good sleek Idea

Sjoerd-Bo3 (approved):
LGTM

dnzxy:
Despite this having been PR'd for a while now, after going over the discussion on the GitHub issue again, I think using `Tabs` for making this very simple selection is a rather odd choice and feels slightly out of place.

I've looked at other proposals and I think the following – I quickly scribbled it, `Standard` should be selected here – is the best out of both worlds:
image

MikePlante1:
> Despite this having been PR'd for a while now, after going over the discussion on the GitHub issue again, I think using `Tabs` for making this very simple selection is a rather odd choice and feels slightly out of place.
> 
> I've looked at other proposals and I think the following – I quickly scribbled it, `Standard` should be selected here – is the best out of both worlds: 

Acceptable, but I still prefer the segmented picker

Sjoerd-Bo3:
Sorry but the radio buttons with a standard option is really odd imo. Looks like it doesnt belong in iOS imo. I liked mikes proposal a lot. But is there bot another iOS native implementation for this?

dnzxy:
It's the exact same thing as using tabs, yet a more natural and common choice of control. 
Can only say I've ran this past another product designer and we agreed on extending the existing design and using radios was the best solution out of all the proposals. 
I'll not die on this hill though :)

Sjoerd-Bo3:
Ok, but then the vote on discord makes the decision for us. @MikePlante1 I think you have a go for this:
 image

dnzxy:
So, since you are now referring to a vote, then let’s please look at what was voted on, which is:
- do you prefer the current status quo?, or
- do you prefer the addition this PR introduces? (all available bolus types as a one tap control)
 
We collectively agree on the latter (see vote), which isn’t the question or point I raised here. The vote on Discord did not include the other "design proposal" Mike had made and which I commented above, plus Mike and I were conversing while it was ongoing. 

However, as I said, I won’t die on this hill, but if we’re trying to be smart about it, let’s do it properly :)

Sjoerd-Bo3:
@MikePlante1 could you donanother cote then please? With all 3 options?

dnzxy:
@MikePlante1 if you can find the time to look into this one again, please update the PR to go along with this approach https://github.com/nightscout/Trio/pull/1030#issuecomment-4177876898 here and then let's get this in. 
If you're crunched on time, let's close this.

g5WS:
Sorry, just to clarify, was this PR closed because @dnzxy prefers radio buttons?

Apple’s guidelines state: “A segmented control offers a single choice from among a set of options,”, which seems to fit this use case. 

I also note that Apple’s Guidelines on selection and input^^ do not list radio buttons, at all.  From what I am reading, radio buttons are a macOS/AppKit control, not an iOS control.

@dnzxy ?

^^ https://developer.apple.com/design/human-interface-guidelines/selection-and-input

dnzxy:
I reckon it was closed because Mike does not have time to see this through, and the segmented picker approach is not something that'll get in. It's predominantly used to let the user switch between subviews or groups, not make distinct options, cf. https://developer.apple.com/design/human-interface-guidelines/segmented-controls.
I personally disagree with the notion that segmented picker works here, but that's not the primary deciding factor, other developers on the team are in agreement with that, and that is.

g5WS:
Radio buttons aren’t native to iOS and aren’t part of Apple’s best practices for iPhone interfaces. 

@dnzxy, which element(s) from Apple’s Human Interface Guidelines on selection and input do you think would work best here? 

https://developer.apple.com/design/human-interface-guidelines/selection-and-input

dnzxy:
Technically speaking, a select-dropdown menu, but it adds taps/clicks, so from a user perspective is a UX downgrade.
Second, radio buttons, as they are a native element to all user interfaces; they don't need to be part of HIGs to be an established element to work well and be known for users. Plus, been present ever since we released Trio. This isn't a pick-and-choose of what Apple lists under https://developer.apple.com/design/human-interface-guidelines 🤷‍♂️ .

g5WS:
Sorry, @dnzxy, can you explain more about why you think radio buttons are native to iOS?  This doesn’t follow for me.

munchyk:
To be fair Apple does clearly indicate throughout their SwiftUI developer docs that segmented **_pickers_** are used for option selection (as seen with the flavor/topping selection in the attached screenshots though it’s actually a horrible example because there is no default option). The most frequently used example of this is the AirPods “Listening Mode” selection (noise cancellation, transparency, etc.)

I happen to prefer a segmented picker but can appreciate both view points.


IMG_8331

IMG_8330

PR #1035, opened 2026-03-29
Fix Smoothing Stop Gap Logic

dnzxy:
* Fix smoothing regression where recent CGM gaps caused all historical smoothed values to be overwritten with raw data as reported on Discord 
* Fallback smoothing is now limited to the most recent contiguous window instead of the full dataset.


MikePlante1 (approved):
Approved based on testing. (This fixed the issue I reported in a private chat)

PR #1036, opened 2026-03-29
feat: Add PebbleService for offline Pebble integration (Phase 1)

MinimusClawdius:
- Added PebbleService.swift (Trio-adapted service)
- Added TrioDataBridge.swift (data conversion layer)
- Added Localizable.strings for settings UI
- Implements local HTTP server on 127.0.0.1:8080 (configurable)
- Uses Trio's Broadcaster for real-time updates
- Safety limits: 10U max bolus, 200g max carbs
- iOS confirmation flow for commands
- Zero cloud dependencies (offline-only)


MikePlante1:
Looks like this was opened accidentally… but it sounds exciting! 

From the description, I’d suggest using the MaxCarbs and MaxBolus setting already in Trio rather than creating new settings, if possible.

Issue #882, opened 2025-12-13
iOS 26.2 UI Glitch — Safe Area at Bottom of Screen


MikePlante1:
## Describe the bug
With iOS 26.2 on my iPhone 14 Pro Max, sometimes the banner on the Main view gets placed behind the tab bar. When this happens, modal screens also lose real estate at the bottom of the screen the size of the tab bar. I haven't found a trigger for it, just that it sometimes occurs when I open Trio. Force closing Trio and reopening has always resolved the UI glitch. If I just minimize Trio and open it again later, sometimes the glitch stays and sometimes it resolves. I do not yet know if other sized phones are affected, but will update my spare iPhone SE (2nd gen) later to see if it happens on it as well.

## To Reproduce
*Steps to reproduce the behavior:*
1. Update iPhone to iOS 26.2
2. Use Trio as usual until the UI glitch randomly appears.

## Screenshots + Expected behavior

| Glitchy Behavior | Expected Behavior |
|---|---|
| ![](https://github.com/user-attachments/assets/0631c93b-b58d-47a4-9408-3bad7b56d2db) | ![](https://github.com/user-attachments/assets/794ac1f0-fd71-42aa-8114-163690a2e835) |
| ![](https://github.com/user-attachments/assets/c4dd0c99-93b5-4359-82ca-a9aed70eaab4) | ![](https://github.com/user-attachments/assets/1978bdcf-d2bc-4c0a-8767-fe034672210f) |
| ![](https://github.com/user-attachments/assets/528d17ed-e291-40c0-83bb-4380263f2d3a) | ![](https://github.com/user-attachments/assets/0e7461c5-b0e7-4036-a0fe-f5e1a2e257b6) |
| ![](https://github.com/user-attachments/assets/4181337e-47f3-4c23-ab9a-027f381738aa) | ![](https://github.com/user-attachments/assets/1c46492a-0a55-45b4-bf8d-ac7753ef5376) |

### Smartphone:
* Hardware: iPhone 14 Pro Max
* OS Version: iOS 26.2 (public release, not just RC1 and RC2)

### Pump:
* Manufacturer: MiniMed
* Model: 723

### CGM:
* Device: Dexcom G6 (Anubis)
* Manager app: Dexcom G6 App

### Trio Version:
* Version Number: 0.6.0.26
* Repo: mikeplante1/trio-dev
[...]

PR #1041, opened 2026-03-31
Fix active bolus view being obscured

kingst:
Remove .ignoresSafeArea(.keyboard) from HomeRootView which could
intermittently cause the bottom content (active bolus/adjustment
views) to be laid out outside the visible area due to SwiftUI safe
area inset miscalculation.

~~Also, fixed a race condition in APSManager so that isSuspended is
always accessed from the main dispatch queue (it's accessed by the
UI).~~

### DIFF
--- a/Trio/Sources/APS/APSManager.swift
+++ b/Trio/Sources/APS/APSManager.swift
@@ -197,7 +197,7 @@ final class BaseAPSManager: APSManager, Injectable {
             .store(in: &lifetime)
 
         deviceDataManager.suspended
-            .receive(on: processQueue)
+            .receive(on: DispatchQueue.main)
             .sink { suspended in
                 self.isSuspended = suspended
             }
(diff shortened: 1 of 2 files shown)


MikePlante1:
This does look to solve the Adjustments/Bolusing banner from being shifted below the tab bar, which is the bigger UI issue so I think it's good to merge as-is, but whenever that UI error occurs it also adds this block to the bottom of every modal sheets, which still happens after applying this PR. Just like before this PR, opening the keyboard or restarting the app gets rid of this UI error (until it occurs again, of course).

t1dude:
Built to my phone and testing now...

t1dude (approved):
Solves the UI issue in bug #882 

I see @MikePlante1 's comment about this PR not solving the UI issue in modal sheets, but I still think it should be merged because it solves the bigger UI issue.

Trio/Sources/APS/APSManager.swift
MikePlante1, line 193: Should this one be switched to `DispatchQueue.main` as well?

kingst: I backed out the race condition fix -- we should take that on in a separate PR.

marv-out (approved):

PR #1047, opened 2026-04-02
Fix chart y-axis label mmol/L display (add parsing)

### LINKED ISSUE #1045: Insulin sensitivity graph legend units hardcoded
In the therapy settings for insulin sensitivities, the graph legend on RHS appears to be hardcoded to mg/dL instead of switching to mmol/L when that unit is selected in preferences. 

Image

dnzxy:
Fixes #1045 

| Before PR | With PR Fix |
|--------|--------|
| image | image |


MikePlante1 (approved):
I would have used

```swift
                    let displayValue = state.rateValues[item.rateIndex].asUnit(state.units)
```

instead of

```swift
                    let displayValue = state.units == .mgdL ? state.rateValues[item.rateIndex] : state.rateValues[item.rateIndex]
                        .asMmolL
```

but since both worth, I'll approve it as is.

PR #1049, opened 2026-04-02
Add filter for manual glucose to fix backfill handling

### LINKED ISSUE #855: Backfill skipped when near manual glucose entries
Testing setup:

* iPhone SE (2nd gen)
* iOS 26.1
* Anubis + G6 Sensor Simulator Hardware
* Dexcom G6 app v1.14.0 (SW11677)
* OrangeLink + Medtronic 723
* Trio:dev v0.6.15

Test:

* Close the Dexcom G6 app to simulate the phone being out of range from the transmitter
* Wait for 2 missed readings
* Enter a manual glucose reading and force a loop cycle
* Open Dexcom G6 app and wait for the next reading
* **Observe:** First missed reading was backfilled, but not the second one that would be close to the same time as the manual glucose entry

| Trio Main | Trio History | Dexcom G6 |
|---|---|---|
|  |  |  |
|  |  |  |

So say you're at a concert or an apple store and your Dexcom isn't reading for half an hour... so you fingerstick and enter that reading. When your Dexcom finally connects and backfills, it'll skip the backfill for the timeslot of the fingerstick reading that was manually entered. Not a big deal, and I think oref might pool cgm readings anyway? But if Trio is your only source of glucose readings in Nightscout, it'd be nice for it to include all of the readings.

Forwarded reply from @kingst :
> We could handle this but it would complicate the logic. With this feature I wanted to keep the logic very simple and make sure that we nail the common backfill case, like leaving your phone in another room for 10 minutes and making sure those two missing readings get logged. If we wanted to handle this case we'd track sources, filter out manual reading, etc but since we've already had one false start with it due to differences in how xdrip and dexcom readings get backfilled, so I wanted to keep the logic super simple.

dnzxy:
Fixes #855

### DIFF
--- a/Trio/Sources/APS/Storage/GlucoseStorage.swift
+++ b/Trio/Sources/APS/Storage/GlucoseStorage.swift
@@ -81,7 +81,8 @@ final class BaseGlucoseStorage: GlucoseStorage, Injectable {
             let withoutDeletedGlucose = self.filterGlucoseValues(
                 glucose,
                 fetchRequest: DeletedGlucoseStored.fetchRequest(),
-                timeBuffer: 1
+                timeBuffer: 1,
+                additionalPredicate: NSPredicate(format: "isManual == %@", false as NSNumber)
             )
 
             // check for a 3.5 minute difference between existing values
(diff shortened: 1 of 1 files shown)


dnzxy:
After having talked this over with @kingst holding off on this at this time and closing this PR.

PR #1051, opened 2026-04-03
Add scroll-to and highlight for settings search results

gordolio:
## Summary

Adds scroll-to and highlight behavior when tapping a settings search result. Previously, search navigated to the correct screen but left the user to find the setting themselves. Now it scrolls directly to the setting and briefly highlights it.

This is the first PR of what will I'm sure be several. This contains the core logic and enables it for Algorithm > Dynamic Settings. Remaining screens will follow in subsequent PRs (many will add a single `.settingsHighlightScroll()` line per view).

## How it works

Two view modifiers drive the feature:

- **`.settingsHighlightScroll()`** Added once per destination view. Wraps the view in a `ScrollViewReader` and watches for a highlighted setting, then scrolls to it.
- **`.settingsSearchTarget(label:)`** Added to each setting row. Sets `.id(label)` so `ScrollViewReader` can find it, and adds a highlight animation that flashes when the label matches.

`SettingInputSection` (the shared component used by most settings) applies `.settingsSearchTarget(label:)` automatically using its existing `label` parameter. Most screens will only need a single `.settingsHighlightScroll()` line to opt in. Custom sections (like the Dynamic ISF picker) will need a `.settingsSearchTarget(label:)` added manually.

The search result list now uses `NavigationLink(value: SearchResultTarget(...))` which carries both the destination screen and the scroll target label. On navigation, the label is passed to `SettingsSearchHighlight` (an `@Observable` object in the environment), which triggers the scroll and highlight.

`scrollTargetLabels` on `SettingItem` handles cases where the search term differs from the UI label (e.g. searching "Max IOB" scrolls to "Maximum Insulin on Board (IOB)"). This mapping is data-only and has no effect until a screen opts-in with `.settingsHighlightScroll()`.

## Demo

https://github.com/user-attachments/assets/e8743199-cb06-431f-afb3-48f92045a9de

## Scroll demo on smaller screen.

https://github.com/user-attachments/assets/51a5d7b0-7dc7-4a84-b5fd-45dbc6183215

### DIFF
--- a/Trio/Sources/Views/SettingsSearchHighlight.swift
+++ b/Trio/Sources/Views/SettingsSearchHighlight.swift
@@ -29,13 +29,8 @@ private struct SettingsHighlightScrollModifier: ViewModifier {
 
     private func scrollToHighlight(proxy: ScrollViewProxy) {
         guard let target = searchHighlight.highlightedSetting else { return }
-        Task { @MainActor in
-            // Give the view time to finish initial layout
-            try? await Task.sleep(nanoseconds: 400_000_000)
+        DispatchQueue.main.asyncAfter(deadline: .now().advanced(by: .milliseconds(500))) {
             withAnimation { proxy.scrollTo(target, anchor: .center) }
-            // Wait for scroll animation + newly-visible section's onAppear to fire
-            try? await Task.sleep(nanoseconds: 500_000_000)
-            searchHighlight.highlightedSetting = nil
         }
     }
 }
@@ -62,9 +57,9 @@ private struct SettingsSearchHighlightAnimationModifier: ViewModifier {
     }
 
     private func startHighlightAnimation() {
+        searchHighlight.highlightedSetting = nil
         highlightOpacity = 0.6
-        Task { @MainActor in
-            try? await Task.sleep(nanoseconds: 1_000_000_000)
+        DispatchQueue.main.asyncAfter(deadline: .now().advanced(by: .milliseconds(1000))) {
             highlightOpacity = 0.0
         }
     }
--- a/Trio/Sources/Views/SettingsSearchHighlight.swift
+++ b/Trio/Sources/Views/SettingsSearchHighlight.swift
@@ -1,0 +1,85 @@
+import SwiftUI
+
+@Observable final class SettingsSearchHighlight {
+    var highlightedSetting: String?
+}
+
+/// Wraps a Screen value with the scroll-target label for search-result navigation.
+struct SearchResultTarget: Hashable {
+    let screen: Screen
+    let scrollLabel: String
+}
+
+private struct SettingsHighlightScrollModifier: ViewModifier {
+    @Environment(SettingsSearchHighlight.self) private var searchHighlight
+
+    func body(content: Content) -> some View {
+        ScrollViewReader { proxy in
+            content
+                .onAppear {
+                    scrollToHighlight(proxy: proxy)
+                }
+                .onChange(of: searchHighlight.highlightedSetting) { _, newValue in
+                    if newValue != nil {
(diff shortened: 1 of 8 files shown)


dnzxy (approved):
I have tested this in a simulator and it's good. 
Code looks good to me, especially for a feature branch :)

Trio/Sources/Views/SettingsSearchHighlight.swift
```diff
@@ -0,0 +1,80 @@
+import SwiftUI
+
+@Observable final class SettingsSearchHighlight {
```
marv-out: Since this only lives on the main thread I would mark it with `@MainActor` as you also did in your Unit tests

Trio/Sources/Views/SettingsSearchHighlight.swift
```diff
@@ -0,0 +1,80 @@
...
+    let screen: Screen
+    let scrollLabel: String
+}
+
+private struct SettingsHighlightScrollModifier: ViewModifier {
+    @Environment(SettingsSearchHighlight.self) private var searchHighlight
+
+    func body(content: Content) -> some View {
+        ScrollViewReader { proxy in
+            content
```
marv-out, line 18: I would probably make use of `.task` here. Like so: 

```
 func body(content: Content) -> some View {
        ScrollViewReader { proxy in
            content
                .task(id: searchHighlight.highlightedSetting) {
                    guard let target = searchHighlight.highlightedSetting else { return }
                    try? await Task.sleep(for: .milliseconds(500))
                    guard !Task.isCancelled else { return }
                    withAnimation { proxy.scrollTo(target, anchor: .center) }
                }
        }
    }

```

this has a few advantages: 

- task gets cancelled automatically when the view disappears
- simplifies the code because it eliminates the need for the separate `onChange`and `onAppear` handling

Trio/Sources/Views/SettingsSearchHighlight.swift
```diff
@@ -0,0 +1,80 @@
...
+                guard searchHighlight.highlightedSetting == label else { return }
+                startHighlightAnimation()
+            }
+            .onChange(of: searchHighlight.highlightedSetting) { _, newValue in
+                guard newValue == label else { return }
+                startHighlightAnimation()
+            }
+    }
+
+    private func startHighlightAnimation() {
```
marv-out, line 50: I would use Task again and probably set highlightedSetting at a later point to nil, not immediately. Like so: 
```
private func startHighlightAnimation() {
        highlightOpacity = 0.6
        // Clear after scroll has had time to complete, not immediately
        Task { @MainActor in
            try? await Task.sleep(for: .milliseconds(800))
            searchHighlight.highlightedSetting = nil
            highlightOpacity = 0.0
        }
    }
```

marv-out (changes requested):
Few minor comments

Trio/Sources/Views/SettingsSearchHighlight.swift
gordolio: Marked as `@MainActor`. Thanks!

Trio/Sources/Views/SettingsSearchHighlight.swift
gordolio: Good idea. I've applied the change. Thanks!

Trio/Sources/Views/SettingsSearchHighlight.swift
gordolio: Thanks for this suggestion! I applied this and added a small delay at the beginning. It felt like the highlight was appearing as the scroll was occurring.

dnzxy (approved):

PR #1092, opened 2026-04-07
Fix ForecastView rendering past forecast points

k9track:
## Summary
- ForecastView's `AreaMark` and `LineMark` were rendering for timestamps in the past, causing visual artifacts on the glucose chart
- Added a lower bound check (`xValue >= Date()`) to all three render locations so forecast marks only render for future timestamps
- The existing upper bound check (`xValue <= Date(timeIntervalSinceNow: TimeInterval(hours: 2.5))`) was already in place — this adds the missing lower bound

## Changes
**File:** `Trio/Sources/Modules/Home/View/Chart/ChartElements/ForecastView.swift`

Three identical changes at lines 44, 60, and 83:
```swift
// Before
if xValue <= Date(timeIntervalSinceNow: TimeInterval(hours: 2.5)) {

// After
if xValue >= Date(), xValue <= Date(timeIntervalSinceNow: TimeInterval(hours: 2.5)) {
```

## Test plan
- [ ] Verify forecast cone only renders for future timestamps on the glucose chart
- [ ] Confirm no visual artifacts from past forecast data points
- [ ] Check that the 2.5-hour upper bound still works correctly

### DIFF
--- a/Trio/Sources/Modules/Home/View/Chart/ChartElements/ForecastView.swift
+++ b/Trio/Sources/Modules/Home/View/Chart/ChartElements/ForecastView.swift
@@ -41,7 +41,7 @@ struct ForecastView: ChartContent {
                         Decimal(minForecast[index] + 1)
                         .asMmolL
 
-                    if xValue <= Date(timeIntervalSinceNow: TimeInterval(hours: 2.5)) {
+                    if xValue >= Date(), xValue <= Date(timeIntervalSinceNow: TimeInterval(hours: 2.5)) {
                         AreaMark(
                             x: .value("Time", xValue),
                             // maxValue is already parsed to user units, no need to parse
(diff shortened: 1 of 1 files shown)


dnzxy:
Do you have an example or an associated issue report for such artifacts and faulty renderings? 

Edit: reason I am asking this is that a forecasted cone that begins in the past is actually intended behavior if the last loop cycle in fact in the past, or the map even have skipped a loop or two. So just trying to identify what issue you are seeing because from a brief overview of the proposed changes it looks like this is breaking intentional behavior.

dnzxy:
Hi @k9track 
can you please provide an example of the issue this fixes? Otherwise I'll go ahead and close this, as I _think_ you are changing intended behavior here.
Thanks!

dnzxy:
Closing this as there is no response from @k9track .

Issue #1095, opened 2026-04-08
After selecting a settings search result, scroll-to and highlight the setting


gordolio:
## Is your feature request related to a problem? Please describe.

Settings search navigates to the correct screen but doesn't always scroll to or highlight the target setting. Testing 149 searchable items revealed failures in several categories: label mismatches between the search index and UI, settings hidden behind disabled toggles, navigation stopping short of nested sub-screens, and missing scroll target annotations on custom UI sections.

## Describe the solution you'd like

Complete scroll-to-highlight coverage for all searchable settings:

1. **Add `.settingsHighlightScroll()` modifier** to all settings sub-screens
2. **Add `.settingsSearchTarget(label:)` annotations** to custom UI sections (e.g. picker-style layouts in User Interface settings)
3. **Fix search index label mismatches** where `searchContents` entries don't match actual UI labels
4. **Add `scrollTargetLabels` mappings** where the search term intentionally differs from the UI label
5. **Handle conditionally-hidden settings** gated behind a toggle that can't be scrolled to when the toggle is off
6. **Fix nested navigation** for screens behind intermediate navigation (Apple Watch, Nightscout Upload/Fetch)

## Describe alternatives you've considered

- **Ignore scroll-to for hidden settings:** Only show them in search results when visible. Reduces scope but changes search behavior.
- **Auto-enable parent toggles:** Rejected because toggling user settings as a side effect of search is unexpected and potentially unsafe for a medical device app.

## Additional context

| Category | Count | Example |
|---|---|---|
| Hidden behind disabled toggle | ~15 | "Reduced Bolus Percentage" (toggle OFF) |
| Nested navigation stops short | ~8 | Apple Watch, Nightscout Upload/Fetch |
| Label mismatch | ~4 | "Glucose Units", "Max Carbs", "DIA" |
| Missing scroll target | ~2 | "Glucose Notifications", "Anonymized Data Sharing" |

### Devices

- [x] Devices (navigation)
- [x] Insulin Pump (navigation)

### Devices > Continuous Glucose Monitor

- [x] CGM (navigation)
- [x] Smooth Glucose Value

### Devices > Smart Watch

- [x] Smart Watch (navigation)
- [ ] Apple Watch (navigation): nav stops at Watch list
- [ ] Display on Watch: nav stops at Watch list
- [ ] Show Protein and Fat: nav stops at Watch list
- [ ] Confirm Bolus Faster: nav stops at Watch list
[...]

PR #1096, opened 2026-04-09
Add scroll targets and fix search index labels for settings search

gordolio:
Part 3 of #1095 

- Replace .listRowBackground(Color.chart) with .settingsSearchTarget(label:) on 7 User Interface sections so they become scroll targets for search
- Add .settingsHighlightScroll() to User Interface screen
- Add scrollTargetLabels mappings for Units/Limits, SMB, Target Behavior, Additionals, and User interface
- Fix incorrect searchContents in Bolus Calculator and User Interface

## About `scrollTargetLabels`
`scrollTargetLabels` maps the search term to the actual UI label:
```swift
scrollTargetLabels: [
    "Max IOB": "Maximum Insulin on Board (IOB)",
    "Max Bolus": "Maximum Bolus",
]
```

There are two cases where these are needed:
1. Abbreviated search terms - the search result/index uses a shorter/common name but the label within the UI shows the full name (e.g. "Max Basal" → "Maximum Basal Rate")
2. Multiple search terms targeting one section - several terms scroll to the same UI element
  - "Dark Mode", "Light Mode", and "Appearance" all scroll to the Appearance picker
  - "Cone" and "Lines" both scroll to the "Forecast Display Type"

### DIFF
--- a/Trio/Sources/Modules/Settings/SettingItems.swift
+++ b/Trio/Sources/Modules/Settings/SettingItems.swift
@@ -8,16 +8,22 @@ struct SettingItem: Identifiable {
     let view: Screen
     let searchContents: [String]?
     let path: [String]?
+    /// Maps a `searchContents` string to the exact label used in `SettingInputSection`
+    /// when the two differ (e.g. `"Max IOB"` → `"Maximum Insulin on Board (IOB)"`).
+    /// Entries whose searchContents string already matches the label don't need an entry here.
+    let scrollTargetLabels: [String: String]?
 
     init(
         title: String,
         view: Screen,
         searchContents: [String]? = nil,
+        scrollTargetLabels: [String: String]? = nil,
         path: [String]? = nil
     ) {
         self.title = title
         self.view = view
         self.searchContents = searchContents
+        self.scrollTargetLabels = scrollTargetLabels
         self.path = path
     }
 }
@@ -26,6 +32,11 @@ struct FilteredSettingItem: Identifiable {
     let id = UUID()
     let settingItem: SettingItem
     let matchedContent: String
+    /// The label string used as the scroll/highlight target in the destination view.
+    /// Falls back to `matchedContent` when no explicit mapping exists.
+    var scrollLabel: String {
+        settingItem.scrollTargetLabels?[matchedContent] ?? matchedContent
+    }
 }
 
 enum SettingItems {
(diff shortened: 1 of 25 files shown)


Trio/Sources/Modules/Settings/SettingItems.swift
dnzxy, line 1: All the labeling changes in this file are breaking localization for our settings search terms, i.e. before this change you were able to search a localized setting name or an English one, and it would list the localized result. Now it lists the hardcoded label as of the collections that were added here. 

Example:
-> Working
Image

-> Broken (see English items here)
Image

Trio/Sources/Modules/UserInterfaceSettings/View/UserInterfaceSettingsRootView.swift
dnzxy, line 1: Sorry, but shouldn't the color be retained, and the settings search target modifier be added here?

dnzxy (changes requested):

Trio/Sources/Modules/Settings/SettingItems.swift
gordolio: Thanks for flagging the localization concern! I investigated and found that the untranslated search results ("Enable SMB With Temporary Target", "Allow SMB With High Temporary Target", "Max UAM SMB Basal Minutes", "Max Allowed Glucose Rise for SMB") are actually a pre-existing issue on dev. These strings were never added to the xcstrings file when they were introduced.

Here's the same "SMB" search on dev:

trio_dev_3

I had initially included some additional searchContents entries but have since removed them to keep this PR scoped strictly to scrollTargetLabels mappings and scroll target modifiers.

Trio/Sources/Modules/UserInterfaceSettings/View/UserInterfaceSettingsRootView.swift
gordolio: The `.settingsSearchTarget(label:)` modifier includes `.listRowBackground(Color.chart)` as its base, with the highlight overlay on top. So the color is retained.

dnzxy:
Planning to review this over the weekend – apologies for the delay @gordolio 🙇🏻‍♂️

Trio/Sources/Modules/Settings/SettingItems.swift
dnzxy: @gordolio Just for completeness sake, should the additional strings (the calues in your key-value pairs here) then be wrapped in `String(localized: '...', comment: '...')` so they can get properly translations by our Crowdin contributors?

dnzxy (approved):
Follow-up question re: localization, but changes are good. 
Localization can be addressed in a follow-up PR, if need be.

PR #1100, opened 2026-04-13
Fix crash in TidepoolManager

marv-out:
- Fix `EXC_BREAKPOINT (SIGTRAP)` crash in BaseTidepoolManagers `uploadDose(_:)` caused by force-unwrapping `event.amount` on a bolus event where amount is nil
- Replace `event.amount!` with `guard let amount = event.amount else { break }` to safely skip invalid bolus events

 The IPS crash report provided the crash address `0x103356038` with Trio's base address at `0x102e6c000`. Using atos with the dSYM debug symbols from the build artifacts, the crash address was symbolicated to:

  `closure #1 in BaseTidepoolManager.uploadDose(_:) (TidepoolManager.swift:313)`

Line 313 is the start of an `events.reduce` block inside a `backgroundContext.perform` closure. Inspecting the code revealed a force-unwrap `event.amount!` on line 327 for bolus events. Tracing the data flow back to `getPumpHistoryNotYetUploadedToTidepool()` confirmed that amount is set via optional chaining (`event.bolus?.amount as Decimal?`) and can be nil when the CoreData bolus relationship is missing.

Looking at the logs a CoreData batch delete of `BolusStored` entities ran at `18:32:34`, two minutes before the crash at `18:34:33` — This could have been the issue here.

 **Device Logs (crash session 2026-04-10)**

  18:32:34 [CoreData] Successfully deleted BolusStored data related to PumpEventStored older than 90 days ✅
  18:34:21 [DeviceManager] New glucose found → loop cycle starts
  18:34:30 [OpenAPS] determine_basal.js completes → pump events processed
  18:34:31 [Service] HealthKit uploadInsulin running (temp basal entries)
  18:34:33 CRASH → uploadDose() → event.amount! is nil

**Crash Report (symbolicated)**

  Exception Type:  EXC_BREAKPOINT (SIGTRAP)
  Crashed Thread:  4 — Queue: NSManagedObjectContext fetchContext

  Thread 4:
    Frame 0: closure #1 in BaseTidepoolManager.uploadDose(_:) (TidepoolManager.swift:313)
    Frame 1: partial apply for closure #1 in BaseTidepoolManager.uploadDose(_:)
    Frame 2-4: CoreData (NSManagedObjectContext.perform)
    Frame 5-11: libdispatch / libsystem_pthread

### DIFF
--- a/Trio/Sources/Services/Network/TidepoolManager.swift
+++ b/Trio/Sources/Services/Network/TidepoolManager.swift
@@ -320,11 +320,12 @@ extension BaseTidepoolManager {
                                     .processTempBasalEvent(event, existingTempBasalEntries: existingTempBasalEntries)
                             )
                     case .bolus:
+                        guard let amount = event.amount else { break }
                         let bolusDoseEntry = DoseEntry(
                             type: .bolus,
                             startDate: event.timestamp,
                             endDate: event.timestamp,
-                            value: Double(event.amount!),
+                            value: Double(amount),
                             unit: .units,
                             deliveredUnits: nil,
                             syncIdentifier: event.id,
--- a/Trio/Sources/Services/Network/TidepoolManager.swift
+++ b/Trio/Sources/Services/Network/TidepoolManager.swift
@@ -320,7 +320,7 @@ extension BaseTidepoolManager {
                                     .processTempBasalEvent(event, existingTempBasalEntries: existingTempBasalEntries)
                             )
                     case .bolus:
-                        guard let amount = event.amount else { break }
+                        guard let amount = event.amount else { return result }
                         let bolusDoseEntry = DoseEntry(
                             type: .bolus,
                             startDate: event.timestamp,


kingst (approved):
I agree with the analysis and the proposed fix, approved!

Do you know the root cause for event.amount to be nil for a bolus event?

marv-out:
I saw a batch delete in the logs two minutes before and batch operations are always tricky in Core Data as they operate directly on the database and other contexts may still keep on seeing those objects as faults. But then again this only applies for objects older than 90 days and the upload function only fetches current objects that aren't uploaded yet. So unfortunately I'm not sure why that happens...

Trio/Sources/Services/Network/TidepoolManager.swift
```diff
@@ -320,11 +320,12 @@ extension BaseTidepoolManager {
                                     .processTempBasalEvent(event, existingTempBasalEntries: existingTempBasalEntries)
                             )
                     case .bolus:
+                        guard let amount = event.amount else { break }
```
dnzxy: I'd suggest to not break here, but to return w/o changing the accumulator (i.e. `result`), as it makes it more obvious that we're continuing the reduce (and its looping flow), rather than exiting it all, but just skipping the event.

So:
```swift
guard let amount = event.amount else { return }
```

marionbarker:
> I agree with the analysis and the proposed fix, approved!
> 
> Do you know the root cause for event.amount to be nil for a bolus event?

I don't know the details of the code in question.  But the question about the nil value caught my attention.

If you are using the deliveredUnits field from a pump manager for an event, that value is always nil when the isMutable field is true.  Once the pump event is finalized, isMutable is set to false and the deliveredUnits field is updated to the final value.

* For a bolus of 3 U that is in progress, isMutable: true, units: 3, deliveredUnits: nil
* For a bolus of 3 U completed, isMutable: false, units: 3, deliveredUnits: 3
* For an interrupted bolus of 3 U that is finalized: isMutable: false, units: 3, deliveredUnits of 2.2 (amount indicates when bolus was interrupted)

For a temporary basal rate of 1.5 U/hr
* while it is active, you have isMutable: true, units: 1.5 and deliveredUnits: nil
* once it completes or is canceled, isMutable: false, units: 1.5, deliveredUnits: insulin delivered by that TBR while it was active

dnzxy:
Completely unrelated to what’s happening here and the respective logic and potential nil value.

kingst:
@marionbarker this almost certainly has to do with an idiosyncrasy of CoreData. My working theory is that it comes out of the cases where we update bolus records after a cancel, but I haven't been able to convince myself that that is it for sure.

dnzxy (approved):

PR #1104, opened 2026-04-18
Feature/high percentage override

### LINKED ISSUE #1105: Override percentage cap of 150% is insufficient for some users
                                                                                                                                            
  The Basal Rate Adjustment picker in the Override forms is capped at 150%. For some users, this limit is too restrictive. A concrete example: a 14-year-old's insulin needs increase to roughly 160% of her baseline settings (CR, ISF, and Basal) during her luteal phase. Her caregiver has been managing this by maintaining two separate sets of settings, but the goal is to allow her to activate a simple override independently — which the current 150% cap makes impossible.                                                                             
                                                                                                                                            
  Raise the maximum override percentage to 200% to better support users with higher insulin resistance needs. A PR implementing this change is ready for review.                    

  Maintaining two separate profiles (one for normal days, one for the luteal phase) and switching between them manually. This works but requires caregiver involvement and is not practical for day-to-day independence.

  ## Additional context                       
                                          
  Discussion thread where this came up:
  https://www.facebook.com/groups/diytrio/posts/2109262199807624/comment_id=2109288003138377&reply_comment_id=2109620686438442             
   
  ## Technical Details                                                                                                                      
                                              
  The change requires updating the `stride` range in two files:
  - `AddOverrideForm.swift`: `stride(from: 40, through: 150)` → `through: 200`
  - `EditOverrideForm.swift`: `stride(from: 40.0, through: 150.0)` → `through: 200.0`                                                       
                                              
  No logic changes — only the picker range is extended.                                                                                     
                  
  ## User Impact                                                                                                                            
   
  Users managing conditions that require insulin adjustments above 150% (such as hormonal cycles, illness, or high insulin resistance) currently have no in-app way to set a sufficient override without manually changing their base settings.

mahref:
closes https://github.com/nightscout/Trio/issues/1105

Increase max override percentage from 150% to 200%                                                                                        
                                                                                                                                            
  Summary                                                                                                                                   
                                                                                                                                            
  - Raises the upper limit of the Basal Rate Adjustment picker in both Add Override and Edit Override forms from 150% to 200%               
  - No logic changes — only the picker range is extended
                                                                                                                                            
  Motivation                                                                                                                                
                                                                                                                                            
  Some users require basal rate overrides above 150% to manage high insulin resistance periods. The current 150% cap is too restrictive for these cases.                            
                                                                                                                                            
  Changes                                                                                                                                   
  
  - AddOverrideForm.swift: stride(from: 40, through: 150) → through: 200                                                                    
  - EditOverrideForm.swift: stride(from: 40.0, through: 150.0) → through: 200.0
                                              
  Test plan                                                                                                                                 
  
  - Open Add Override — verify picker scrolls up to 200%                                                                                    
  - Open Edit Override on an existing override — verify picker scrolls up to 200%
  - Verify values below 150% still work as before

### DIFF
--- a/Trio/Sources/Modules/Adjustments/View/Overrides/AddOverrideForm.swift
+++ b/Trio/Sources/Modules/Adjustments/View/Overrides/AddOverrideForm.swift
@@ -110,7 +110,7 @@ struct AddOverrideForm: View {
                                 set: { state.overridePercentage = Double($0) }
                             ), label: Text("")
                         ) {
-                            ForEach(Array(stride(from: 40, through: 150, by: percentageStep)), id: \.self) { percent in
+                            ForEach(Array(stride(from: 40, through: 200, by: percentageStep)), id: \.self) { percent in
                                 Text("\(percent) %").tag(percent)
                             }
                         }
--- a/Trio/Sources/Modules/Adjustments/View/Overrides/EditOverrideForm.swift
+++ b/Trio/Sources/Modules/Adjustments/View/Overrides/EditOverrideForm.swift
@@ -172,7 +172,7 @@ struct EditOverrideForm: View {
                             label: Text("")
                         ) {
                             ForEach(
-                                Array(stride(from: 40.0, through: 150.0, by: Double(percentageStep))),
+                                Array(stride(from: 40.0, through: 200.0, by: Double(percentageStep))),
                                 id: \.self
                             ) { percent in
                                 Text("\(Int(percent)) %").tag(percent)


dnzxy:
Hi @mahref 

first of all, thank you for your absolutely exemplary approach to contributing to Trio (issue, PR, concise description, etc.) – appreciated!

However, at this point, we likely won’t be able to accept this change as proposed, mainly because it has a fairly large safety implication that may not be obvious to users. By allowing an override to increase basal by 200%, the maximum SMB “base” can effectively scale up to double as well. For example, someone running a flat 1 U/hr basal with 60 SMB/UAM SMB minutes could normally receive up to about 1 U per loop cycle as an SMB. With a 150% override that becomes 1.5 U, and with 200% it becomes 2 U before even considering additional effects from COB and possible CR changes.

That means the change does not just make the system “more aggressive” in a small or isolated way, like a Temporary Target would. It can materially increase the amount of insulin available for SMB delivery, and users may not fully realize that this is happening. Since SMB dosing already relies on multiple guardrails, changing the basal/CR side of the calculation could create a slippery slope with real overdose risk, especially in edge cases.

The underlying use case absolutely makes sense, and there may be a safer way to support it in the future, for example by making override behavior more explicit or configurable. But that would probably need a broader design discussion and may fit better into the future post oref-swift work rather than changing the current JavaScript implementation as present in Trio now.

So for now, the concern is not that the need is invalid, but that this particular implementation changes dosing behavior in a way that is potentially too safety-sensitive and not transparent enough for users.

PR #1106, opened 2026-04-18
Add loop-failure alarm sound

bjorkert:
## Summary

Adds an audible loop-failure alarm that plays through the mute switch when the loop hasn't run for a configurable delay.

## What it does

- Plays a looping sound via `AVAudioSession.Category.playback` so it's audible even when the mute switch is on, with optional system-volume override.
- Fires when `Date - lastLoopDate ≥ delayMinutes` and the loop is genuinely broken (skips open-loop, manual temp basal, pump suspended, and stale-app cases over 60 min).
- Two dismissal surfaces:
  - **In-app red banner** (`AlarmBannerView`) on the home screen with inline Snooze (20m/1h/3h/6h) + Acknowledge.
  - **Lock-screen / Apple Watch notification** with the same action buttons — only posted when the app isn't active, so it doesn't duplicate the banner.
- Snooze silences the alarm for the chosen duration and re-fires afterward if the loop is still broken. Acknowledge silences it until the loop resumes (at which point the acknowledge flag resets automatically).

## Settings

Under **Notifications → Loop Failure Alarm**:
- Enable toggle (off by default)
- Delay (minutes)
- Volume override + target volume
- Test button

## Notes

- New service: `AlarmSound` (`Trio/Sources/Services/AlarmSound/AlarmSound.swift`) — shared singleton, @MainActor, @Published `isAlarmActive`.
- Reuses the existing `NotificationResponseAction` enum; loop-failure notification snoozes are routed to `AlarmSound.snooze(for:)`, glucose snoozes keep their existing `applySnooze` path.
- Per-minute check driven by a new `timerHeartbeat` publisher on `FetchGlucoseManager` (fires from the existing glucose fetch timer, regardless of new glucose).


dnzxy:
We had a longer conversation about this with our friends over in Loop-land, too. 
Let's get together for a quick call or something, and discuss this in detail. I think you'll like the idea(s) and feedback @bjorkert ☺️

PR #1107, opened 2026-04-19
Core Data Performance Improvements, Fixes and Adjustments

marv-out:
**Core Data Context Rework**
Eliminates long-lived shared `NSManagedObjectContext` instances across all storages and service classes which was the reason for inconsistencies in the synchronization of our background contexts and probably also the root cause for the crash that was solved differently with one of my previous PRs #1100. Each data operation now creates a fresh task context via `CoreDataStack.shared.newTaskContext()`, preventing cross-context data races and stale object access. Every background context now is only active for a special task and it has the latest data from the `PersistentStore`. This is done similarly in Apples Earthquakes Sample: https://github.com/TomaszLizer/Apple-Earthquakes/

Those changes need to be tested thoroughly. I am testing it on myself for about 2 months now without any problems. But in theory I could have accidentally introduced a wrong `NSManagedObjectContext` access in a class that I simply don't use. This should be deteced via our `ConcurrencyDebug` flag though which then makes the app crash. So be aware of that.

 **N+1 Query Elimination**
 Adds `relationshipKeyPathsForPrefetching` for bolus/tempBasal on all `PumpEventStored` fetches (`OpenAPS`, `APSManager`, `WatchManager`, `GarminManager`)and `forecastValues` on `OrefDetermination` fetches. Eliminates hundreds of individual `WHERE Z_PK = ?` round-trips per Loop that I could identify in the SQLDebug logs. When we were fetching a pump event with its bolus relationship for example the relationships don't get loaded automatically, but instead created as a fault. So when we were doing 1 query only to iterate over the items to get its bolus property after that this was unefficient as this would fire several faults. `relationshipKeyPathsForPrefetching` gets rid of this as the bolus property is now already in the context cache after the first query.

**TDD Aggregate Optimization**
Replaces full ZTDDStored row loads (up to 1,041 rows) in `calculateWeightedAverage` with a single SQL aggregate query `(SELECT total(), COUNT())` — reducing fetch time from ~6 ms to ~0.8 ms per call.

**Carb/FPU & Determination Prefetch**
 Adds `SELF IN` batch-prefetch hop for CarbEntryStored and OrefDetermination in the Home state model to prevent N+1 faults during `getNSManagedObject` materialization on the view context.

**NSFetchedResultsController migration**
New NSFetchedResultsControllerDelegate plus migration of HomeStateModel, TreatmentsStateModel, NightscoutManager, TidepoolManager, and HealthKitManager to FRCs bound to the viewContext, replacing the previous objectID/publisher logic.

**History-sourced change feed (CoreDataObserver.swift)**
The hand-rolled `changedObjectsOnManagedObjectContextDidSavePublisher` (based on `NSManagedObjectContextDidSave`) is replaced by `CoreDataStack.entityChangePublisher`, sourced from persistent history. This now also captures batch operations and cross-process changes.

**New dedup logic in `PumpHistoryStorage`**
The new dedup key combines `timestamp` AND `type` so that a bolus and temp basal event can coexist at the same time. Also add a unique constraint that ensures this behaviour.


marv-out:
Updated the pr description to really include every change and pulled in dev (which was also quite annoying with lots of merge conflicts xD)

kingst:
This PR is big and introduces a number of different changes, so I'll break down my reviews into these topics:

## 1. `CoreDataStack.swift`
- [x] New `entityChangePublisher`
- [x] Removal of `context.name`/`transactionAuthor` on fetch contexts

## 2. Schema migration
- [x] `.xcdatamodel`: new **uniqueness constraint on `PumpEventStored` (`timestamp` + `type`)**.
- [x] `PumpHistoryStorage.swift`: new dedup key = `timestamp` AND `type`.

## 3. History-sourced change feed
- [x] `CoreDataObserver.swift`: old `NSManagedObjectContextDidSave` publisher deleted, only `filteredByEntityName` kept.
- [x] Replaced by persistent-history feed — now also catches batch insert/delete + cross-process changes.

## 4. NSFetchedResultsController migration
New `NSFetchedResultsControllerDelegate.swift`; migrated:
- [x] `HomeStateModel.swift`
- [x] `TreatmentsStateModel.swift`
- [x] `NightscoutManager.swift`
- [x] `TidepoolManager.swift`
- [x] `HealthKitManager.swift`
- [x] Verify FRCs are bound to the viewContext and update the UI on the main thread.

## 5. "Fresh context per operation" pattern
- [x] Read **one** representative storage file end-to-end (e.g. `GlucoseStorage.swift`).
- [x] Then skim the rest, confirming they follow the same pattern (no leftover shared/long-lived context):
  `CarbsStorage`, `DeterminationStorage`, `OverrideStorage`, `TempTargetsStorage`, `TDDStorage`, `ContactImageStorage`, plus the `*Manager` and `*IntentRequest` files.
- [x] Audit all context and make sure that none of them are long lived (including `BaseBolusSafetyValidator`)

## 6. Performance optimizations
- [x] N+1 elimination: `relationshipKeyPathsForPrefetching` (bolus/tempBasal on `PumpEventStored`; `forecastValues` on `OrefDetermination`) in `OpenAPS`, `APSManager`, `AppleWatchManager`, `GarminManager`.
- [x] `TDDStorage.calculateWeightedAverage`: full row load → single `total()`/`COUNT()` SQL aggregate.
- [x] `SELF IN` batch-prefetch for carbs/determination in the Home state model.

Model/CoreDataStack.swift
kingst, line 179: This is async and will race with the send below, I suggest moving the send inside the `perform` block 

https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext/perform(_:)

kingst: forced unwrap, not related to this PR but might be worth fixing since we're here.

kingst: forced unwrap, not related to this pr but since we're changing this file consider removing it

kingst (commented):
Overall this looks like a good change. I really don't understand CoreData well and we're in the deep waters here so my review is low confidence. With that being said I did have a few comments on item (1) from my review decomposition.

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
```diff
@@ -47,187 +47,145 @@ final class BasePumpHistoryStorage: PumpHistoryStorage, Injectable {
...
                 switch event.type {
                 case .bolus:
-
                     guard let dose = event.dose else { continue }
                     let amount = self.roundDose(
                         dose.unitsInDeliverableIncrements,
                         toIncrement: Double(self.settings.preferences.bolusIncrement)
                     )
+                    // restrict entry to now or past
+                    let timestamp = event.date > Date() ? Date() : event.date
```
kingst: If we're using timestamps as IDs we shouldn't modify them they should always be equal to `event.date` even if they're in the future, IMO

kingst (changes requested):
This is a review for (2) from my list. The updated dedup logic taking into account type is the right change to make. My concerns are:
* [Minor] We should make bolus events different if they're from the pump vs external in the dedup logic
* [Major] Adding a constraint to coredata is unsafe without migration code. My suggestion is to drop the coredata constraint, but if you want to keep it we need to have a proper migration just in case there are databases out there with a duplicate.

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
dnzxy: This has been in place since basically forever and ensures, that (external) bolus event cannot be set into the future. You _cannot_ add future insulin.

kingst: Is there a way to enforce that external bolus events aren't in the future that doesn't interfere with our deduplication logic? If we're going to use timestamps for event matching we shouldn't have code that can change them.

dnzxy: Events emitted from pump managers will never be in the future, this is merely a second layer of defense, and those of external boluses theoretically can be (via the UI), but we hard-cap in the applicable path for that. I'm not entirely certain why this line exists for the PM-given pathway.

This is all we need:
image

kingst:
Items (1) and (2) from my review list are now ✅ there is still the minor comment about making external bolus events different from bolus events for dedup logic, which is worth a confirmation that you are not going to do, but otherwise this code LGTM. Moving on to Item (3)...

kingst (commented):
Item (3) from my list LGTM

Trio/Sources/Modules/Home/HomeStateModel.swift
```diff
@@ -135,21 +135,220 @@ extension Home {
...
+        // bound to the viewContext. The controllers keep their `fetchedObjects` continuously in sync
+        // with the viewContext (which in turn is fed by the persistent history merge in CoreDataStack)
+        // and notify us via their delegate's `onContentChange` closure. This replaces the previous
+        // hand-rolled `changedObjectsOnManagedObjectContextDidSavePublisher` + re-fetch approach.
+
+        @ObservationIgnored let glucoseControllerDelegate = FetchedResultsControllerDelegate()
+        @ObservationIgnored private(set) lazy var glucoseController: NSFetchedResultsController<GlucoseStored> = {
+            let request = NSFetchRequest<GlucoseStored>(entityName: "GlucoseStored")
+            request.sortDescriptors = [NSSortDescriptor(keyPath: \GlucoseStored.date, ascending: true)]
+            request.predicate = NSPredicate.glucose
```
kingst, line 157: because of the way this NSFetchResultsController is created (lazy var, never updated beyond first access) this predicate never changes so the predicate data remains the same even as Trio continues to run. In the previous version this predicate was create fresh each time the results were fetched. There are a number of these predicates that stay frozen in time beyond this one.

In other words, this predicate captures a call to Date() when the app starts on first access and never updates it so the time window for the fetch continues to increase as Trio runs.

We should fix all of these, but I'm not sure of the cleanest way to do that.

Trio/Sources/Services/Network/TidepoolManager.swift
```diff
@@ -128,38 +188,6 @@ final class BaseTidepoolManager: TidepoolManager, Injectable {
...
-        coreDataPublisher?.filteredByEntityName("GlucoseStored").sink { [weak self] _ in
-            guard let self = self else { return }
-            Task { [weak self] in
-                guard let self = self else { return }
-                await self.uploadGlucose()
-            }
-        }.store(in: &subscriptions)
-
-        // Register for settings that aren't saved from a single editor screen
-        broadcaster.register(SettingsObserver.self, observer: self)
```
kingst, line 159: Do we still need the SettingObserver and PreferencesObserver? It looks like these were dropped and not replaced.

kingst (commented):
Review for Item (4) from my list.

kingst (commented):
Item (5) from my list generally looks good, it's big improvement. I did find one stray long-lived context in `BaseBolusSafetyValidator` so we should fix this and double check just to make sure there aren't any others lingering.

kingst (commented):
Item (6) LGTM

Trio/Sources/Modules/Home/HomeStateModel.swift
marv-out: Yes, youre absolutely right. Good catch. I implemented a relatively simple solution by updating the predicate every time Trio comes into the foreground. I will push that later and we can discuss on that

marv-out:
> Items (1) and (2) from my review list are now ✅ there is still the minor comment about making external bolus events different from bolus events for dedup logic, which is worth a confirmation that you are not going to do, but otherwise this code LGTM. Moving on to Item (3)...

I didn't intend to implement that because external bolus events are being added by "storeExternalInsulinEvent" in a different path. Or did I missunderstand your comment?

Trio/Sources/Services/Network/TidepoolManager.swift
marv-out: You are right, will add them back

marv-out:
> Item (5) from my list generally looks good, it's big improvement. I did find one stray long-lived context in `BaseBolusSafetyValidator` so we should fix this and double check just to make sure there aren't any others lingering.

Fixed with my latest commit! And found another moc in GlucoseAlertCoordinator. Now there shouldn't be any other long lived contexts left

marv-out:
I've also added a few minor changes which rather belong to this PR due to the new architecture this PR introduces. They include: 

- reanchoring the FRCs in HomeStateModel when Trio comes to the foreground
- debouncing Core Data remote-change merges and forecast recomputes — collapse NSPersistentStoreRemoteChange bursts into one history fetch + merge (200 ms), and coalesce rapid determination changes into a single forecast recompute (cancel-previous, 150 ms).
- preventing stale forecast runs from overwriting fresher results — check Task.isCancelled after every suspension point in updateForecastData(); register the startup run in forecastUpdateTask so newer changes cancel it
- computing temp basal bars in one main-actor pass — replace the quadratic first(where:) scan with a single linear sweep; fixes an off-main read of viewContext-bound objects and removes the stale-assignment race of the unstructured task.
- tearing down FRC pipelines on dismissal in TreatmentsStateModel— nil the onContentChange closures and cancel in-flight tasks in cleanup, so a dismissed sheet no longer re-runs the bolus calculator on every viewContext merge
- removing dead observer boxes in the broadcaster — compact WeakObjectSet on every registration; dead boxes were never removed and grew the observer sets per screen open/close

marv-out:
I won't make any new additions to this PR from now on because it is a rather big PR anyway unless the review of the last changes show major problems. It would be cool if we could move this PR forward soon. But so far, thanks for your thorough review @kingst !

kingst (approved):
LGTM. A PR this large is extremely difficult to review, but I do think that this change is an improvement and should go in. One thing to double check on is that I think some of the submodules are out of sync with dev. In particular I had to `git submodule update` for these:
* DanaKit
* G7SensorKit
* MedtrumKit

marv-out:
Should be good to go now @kingst :)

dnzxy (approved):

PR #1111, opened 2026-04-24
Add optional adjustments confirmation

### LINKED ISSUE #1110: Override Confirmation
Hello, I wanted to share a idea that came to my mind due to an experience I had where I accidentally activated an Override.

I would suggest that when you click on saved overrides, there should appear a popup message in which you need to confirm again, that you want to activate the covering/temp goal. I cant be the only one that has mistakenly clicked on an override which then mistakenly activated. 🙃

dnzxy:
* Introduce new setting `requireAdjustmentsConfirmation`
* Refactor adjustments presets activation handling
   * If `requireAdjustmentsConfirmation` is `true` / enabled, show confirmation dialog before firing preset activation workflow
   * If `requireAdjustmentsConfirmation` is `false` / disabled, immediately activate preset (like before)
* Addresses #1110

| New Setting | *New* With Confirmation | Without Confirmation |
|--------|--------|--------|
| image | image | image |

### DIFF
--- a/Trio/Sources/Models/TrioSettings.swift
+++ b/Trio/Sources/Models/TrioSettings.swift
@@ -75,6 +75,7 @@ struct TrioSettings: JSON, Equatable, Encodable {
     var smartStackView: LockScreenView = .simple
     var bolusShortcut: BolusShortcutLimit = .notAllowed
     var timeInRangeType: TimeInRangeType = .timeInTightRange
+    var requireAdjustmentsConfirmation: Bool = false
 
     /// Selected Garmin watchface (Trio or SwissAlpine)
     var garminWatchface: GarminWatchface = .trio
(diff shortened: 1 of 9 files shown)


dnzxy:
~~Note to self:~~
- ~~add new setting to settings search~~
- ~~add new setting to settings export~~

✅

MikePlante1:
Should the messages here more closely match the “stop” text?

> Stop the Override “name of override”?

> Stop the Temp Target “name of TT”?

to be:

> Activate the Override “name of override”?

> Activate the Temp Target “name of TT”?

MikePlante1:
Would “Start” be simpler/better than “Activate”? 
Either is fine for me, just asking.

dnzxy:
> Should the messages here more closely match the “stop” text?
> 
> > Stop the Override “name of override”?
> 
> > Stop the Temp Target “name of TT”?
> 
> to be:
> 
> > Activate the Override “name of override”?
> 
> > Activate the Temp Target “name of TT”?

Where is "here" referring to? Button text? Dialog title? Sorry, bit confused 😅 
Yes, we can absolute go with `Start` instead of `Activate`. We also use that label in the forms.

MikePlante1:
> > Should the messages here more closely match the “stop” text?
> > > Stop the Override “name of override”?
> > 
> > 
> > > Stop the Temp Target “name of TT”?
> > 
> > 
> > to be:
> > > Activate the Override “name of override”?
> > 
> > 
> > > Activate the Temp Target “name of TT”?
> 
> Where is "here" referring to? Button text? Dialog title? Sorry, bit confused 😅 Yes, we can absolute go with `Start` instead of `Activate`. We also use that label in the forms.

From the screenshot in the PR description. I did not build it yet.

| new dialog here | existing dialog |
|---|---|
|  |  |

kristinkand:
We tested both with/without confirmation. Both work. In our opinion the extra confirmation is obsolete. Great, that it can be turned off under settings! 👍

bjorkert (approved):

PR #1118, opened 2026-04-28
Smoothing: fix using newest 350 values of last 24hrs

mountrcg:
Currently the fetch BG for smoothing uses ascending: true (oldest first) combined with fetchLimit: 350. Core Data applies sort + limit at the SQL level, so it returns the oldest 350 rows that match the predicate — not the newest. This changes it to newest 350 values of the day.

**Is this a real issue:**
I encountered the situation when experimenting with the outstanding smoothing change recently. I was backfilling the sim with different NS servers and ended up with more than the regular 288 BG values. At that moment I did not get current values being smoothed. Checking Glucose History, it showed smoothing progressing every 5 min but only 5 hrs in the past.

It might become relevant whenever something will be done with 1min CGM values. Nevertheless as @dnzxy said: it still should be handled gracefully.

### DIFF
--- a/Trio.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Trio.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -1,5 +1,168 @@
 {
+  "originHash" : "b271d5de8862534ec5ccf9ccfd7ce226afa7ed4c799b7066ccbe2281782402a1",
   "pins" : [
+    {
+      "identity" : "abseil-cpp-binary",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/google/abseil-cpp-binary.git",
+      "state" : {
+        "revision" : "bbe8b69694d7873315fd3a4ad41efe043e1c07c5",
+        "version" : "1.2024072200.0"
+      }
+    },
+    {
+      "identity" : "app-check",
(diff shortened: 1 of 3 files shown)


kingst (commented):
Thanks for catching this @mountrcg, this is a real issue and your fix LGTM. Would you mind adding a unit test for this? Once you have that then I'll approve it.

mountrcg:
added two more tests:
**testFetchGlucoseKeepsMostRecentWhenOverLimit**
Title: "fetchGlucose retains the most recent 350 readings (not the oldest) when 24h holds more than 350"
Seeds 360 readings, calls fetchGlucose directly, asserts 350 returned, chronological order, newest reading present, oldest excluded.
**testExponentialSmoothingCoversCurrentBGAboveLimit**
Title: "Exponential smoothing writes a smoothed value for the current BG when 24h holds more than 350 readings"
Seeds 360 contiguous readings, runs exponentialSmoothingGlucose, asserts the most recent reading has smoothedGlucose != nil.

Tested with old and fixed code:

State | New tests
-- | --
Buggy code (ascending: true + limit 350) | ❌ both fail with the exact symptom described (newest readings dropped, current BG never smoothed)
Fixed code (ascending: false + reversed) | ✅ all 11 tests pass

AndreasStokholm (approved):
Great spot and fix! Love the addition of the tests to prevent regressions 👏

PR #1120, opened 2026-04-28
Add context menu to all swipeable list items

### LINKED ISSUE #1119: History Entry - Long Press to Edit/Delete
## Problem

Some users are unfamiliar with the row swipe gesture to reveal the Edit/Delete actions (e.g. remove a meal entry).

## Solution

In addition to potentially adding a visual cue with a row bounce and actions peek, adding a long press action menu popup would be very helpful as this is a more commonly used UI interaction. I think an edit button/icon should also be added in place of the global "Add Glucose +" on the History screen. 

(See attached mockup for potential UI implementation)

## Note

The design style used in the attached image is based on iOS prior to Liquid Glass (the current style used in Trio). iOS 26/Liquid Glass does not blur the background (instead relying on shadows and glass effect) and uses a slightly different menu style.

The mockup i made is for reference only. I made it on my phone in 15 min so not sure it can be relied upon for faithfulness to the native OS design.

Image

dnzxy:
## Summary

Adds context menus to list items that previously only exposed actions through swipe gestures, improving action discoverability and accessibility across the app. This makes delete/edit actions available via long press or secondary click in addition to existing swipe actions.

Addresses #1119.

## Changes

- Added context menus for Adjustments
  -  Override presets
  - Temp Target presets
- Added context menus for History items:
  - Glucose entries
  - Insulin/bolus entries
  - Carb and FPU meal entries
- Added context menu delete support to therapy setting rows in onboarding.
- Updated delete buttons to use `.destructive` role for consistent system styling and semantics.

## Screenshots

| Carb Entries | Treatments | Glucose |
|--------|--------|--------|
| image | image | image | 

| Override presets | Temp Target presets | Therapy Item |
|--------|--------|--------|
| image | image | image | 

### Notes
The placement / alignment of the context menu for therapy items seems odd. However, this is the default render mode for the menu; there's no custom alignment implemented. I'm not sure it's a good addition for therapy items.

### DIFF
--- a/Trio/Sources/Modules/Adjustments/View/Overrides/AdjustmentsRootView+Overrides.swift
+++ b/Trio/Sources/Modules/Adjustments/View/Overrides/AdjustmentsRootView+Overrides.swift
@@ -93,7 +93,7 @@ extension Adjustments.RootView {
 
     func actionButtonsForOverrides(for preset: OverrideStored) -> some View {
         Group {
-            Button(role: .none) {
+            Button(role: .destructive) {
                 selectedOverride = preset
                 isConfirmDeletePresented = true
             } label: {
--- a/Trio/Sources/Modules/Adjustments/View/Overrides/AdjustmentsRootView+Overrides.swift
+++ b/Trio/Sources/Modules/Adjustments/View/Overrides/AdjustmentsRootView+Overrides.swift
@@ -19,8 +19,11 @@ extension Adjustments.RootView {
                 overridesView(for: preset, showCheckMark: showOverrideCheckmark) {
                     enactOverridePreset(preset)
                 }
+                .contextMenu {
+                    actionButtonsForOverrides(for: preset)
+                }
                 .swipeActions(edge: .trailing, allowsFullSwipe: true) {
-                    swipeActionsForOverrides(for: preset)
+                    actionButtonsForOverrides(for: preset)
                 }
             }
             .onMove(perform: state.reorderOverride)
@@ -88,7 +91,7 @@ extension Adjustments.RootView {
         }
     }
 
-    func swipeActionsForOverrides(for preset: OverrideStored) -> some View {
+    func actionButtonsForOverrides(for preset: OverrideStored) -> some View {
         Group {
             Button(role: .none) {
                 selectedOverride = preset
--- a/Trio/Sources/Modules/Adjustments/View/TempTargets/AdjustmentsRootView+TempTargets.swift
+++ b/Trio/Sources/Modules/Adjustments/View/TempTargets/AdjustmentsRootView+TempTargets.swift
@@ -21,7 +21,7 @@ extension Adjustments.RootView {
             ForEach(state.scheduledTempTargets) { tempTarget in
                 tempTargetView(for: tempTarget)
                     .swipeActions(edge: .trailing, allowsFullSwipe: true) {
-                        swipeActionsForTempTargets(for: tempTarget)
+                        actionButtonsForTempTargets(for: tempTarget)
                     }
             }
             .listRowBackground(Color.chart)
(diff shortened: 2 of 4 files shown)


marionbarker:
## Test

I tested the long-press as well as the swipe to get to edit/delete.
I found this alternative pretty natural.

I think this behavior would have helped the problem I reported to @bastiaanv where I did not understand how to clear an alert on a different pump manager.

marionbarker (approved):
Approve from test. Someone else should do the code review.

munchyk:
Nice work @dnzxy! Simple and does the job. Did some quick tests using Xcode simulator and its working fine for me.

One note: manual glucose entries are not added/removed from the list until the user changes app screens using the tab/navigation bar and returns back to the list. This was an issue before the current change, though.

Also, would it make sense to add "edit" functionality to the manual glucose and external insulin entries? Editing a glucose entry is crucial as of now because there is no time selection for the entry. Instead, the time of entry is automatically applied. If a user needed to change a glucose value, they'd be forced to delete and add a new entry meaning the time would no longer be accurate.

Not sure if this all falls under the current scope of the PR.

munchyk:
Something else that fits alongside this would be the addition of an edit button where the current "Add Glucose" button is on the History screen. This would also work for the Therapy screens. Its inline with current iOS UX though sometimes they use a "select" button when there are multiple actions

I think a "plus" icon (for adding entries of external insulin and manual/external glucose) and "edit" icon/button (for editing/deleting entries) that are specific to the currently displayed list would be helpful if not more useful. The global “Add Glucose” button is a bit confusing being that its visible on all list views (I've mistakenly pressed it a bunch of times) and i think having all external entries (glucose and insulin) in the same place is a little more intuitive - though i completely understand that there may be reason to keep external insulin right where it is being that people may be dosing alongside meals.

dnzxy:
Thanks for testing you two :) 

@munchyk all of your raised points were discussed in length during various iterations in the (closed) beta and what you see in Trio now is the "approved" final result, so intentional. External insulin used to actually be right there and behave like you describing and it was moved to our centralized treatment view :) 

Can you open an issue for the glitch you see for manual glucose entries? I‘m not sure I understand and going by what I think you‘re saying I cannot say I see the same in my own setup or in simulators.

munchyk:
@dnzxy I can appreciate that i may not have been privy to the discussion that led to the final UI results! 

I do think that it couldn't hurt to *also* have an "Add Insulin/Bolus/Dose" or "+" icon in the History tab while the insulin list is in view but that's just me. 

Independent of that, i think that the "Add Glucose" button at the top right of the History tab should be limited to the glucose list view only. 

Also, an "edit"/"select" button or sf symbol would be helpful and is part of the three tiered approach that Apple implemented throughout iOS.

Regardless, great work - as usual!

dnzxy:
@munchyk , I appreciate you putting in thought on this, so I will reply to your comment, but want to preface this with a note that I just want to give context, not open this up for discussion. Just neutral info, so to speak.

> I do think that it couldn't hurt to _also_ have an "Add Insulin/Bolus/Dose" or "+" icon in the History tab while the insulin list is in view but that's just me.

Treatments are entered via a centralized treatment view, i.e. via the view opened by the big `(+)`. There are also quite a few guardrails (programmatically and through UI confirmations) present in treatment view, plus the bolus calculator, and replicating that was decided against.

> Independent of that, i think that the "Add Glucose" button at the top right of the History tab should be limited to the glucose list view only.

It used to behave like that in the past, but since it's the only entity that can be added via History, user feedback was to eliminate the second tap (1. history, 2. glucose tab) and just offer it "generally". We used to only have a big `+`, but added a descriptive label, again, to support user in navigating around the app.

> Also, an "edit"/"select" button or sf symbol would be helpful and is part of the three tiered approach that Apple implemented throughout iOS.

You're absolutely correct, Apple does offer this typically in list views, especially for "mass" deletions. Typically you have an `Edit` in the top leading and a `+` in the top trailing corner of the screen. For Trio and our context here, there are certain caveats. Not all treatments can be deleted, those than can require a confirmation, some treatments require additional authentication upon deletion, others don't, some can be edited, other cannot, and all of those caveats are intentional. I.e., not being able to edit (manual) glucose entries and insulin dosages is intentional, for example.

This has gone through I believe 5 iterations and we had a ton of feedback on it and while this – and most of what we have in Trio – is attempting to follow Apple HIGs as closely as possible, we have made context-aware adjustments and "design decisions" that are now present in the app as you see it today. Never fully set in stone, like this PR goes to show, but usually very intentional and well thought out :)

munchyk:
@dnzxy Thank you for the reply.

Issue #1122, opened 2026-04-29
glucose date encoded as string breaks JS meal.currentDeviation


mountrcg:
## Root cause

AlgorithmGlucose.encode(to:) (Trio/Sources/Models/AlgorithmGlucose.swift:71-72) encodes the glucose date field as a JSON string of digits:

let dateAsUnixTimestamp = String(format: "%.0f", … * 1000)
try container.encode(dateAsUnixTimestamp, forKey: .date)
Producing "date": "1777454386860" in the JSON handed to the JS algorithm bundle.

In oref0/lib/determine-basal/cob.js, the most recent glucose bucket is created by reference (bucketed_data[0] = glucose_data[0]) and then read with bgTime = new Date(bucketed_data[i].date). In JavaScript, new Date("1777454386860") returns Invalid Date — the constructor accepts numeric epoch ms but does not parse a numeric string as one.

## Effect

For every meal/COB calculation, the i == 0 iteration in cob.js (the only iteration that sets currentDeviation) runs with an Invalid bgTime. That cascades:

1. `iob_inputs.clock` becomes Invalid Date.
2. `iob_total` does `treatment.date <= now` against `NaN`, which is false for every treatment, so activity sums to 0.
3. `bgi = -activity * sens * 5 = 0.`
4. `currentDeviation = avgDelta - bgi = avgDelta`, i.e. the BGI term is silently dropped.

The error magnitude equals the lost BGI: bias = -bgi_correct = activity * sens * 5. It is small or invisible at low IOB (Swift's bgi rounds to ~0 anyway, so the discrepancy hides), and grows to ~1–2 mg/dL during active insulin absorption. 

autoISF amplifies it by raising sens, which is why the divergence is most visible when autoISF is on, but the bug is deterministic and present every cycle regardless of autoISF or standard oref0. .date's JSON type doesn't affect any Swift consumer — only the JS bundle, so I was able to identify the error with @kingst  compare algorithms that runs both the active and the inactive algorithm path on identical inputs each loop tick, writes paired timings and value diffs to log.

A correct downstream mealCOB / currentDeviation underestimates real BG impact attributable to insulin, biasing carb-absorption detection and any consumers of currentDeviation.

## Fix

Encode date as a JSON number (Int64 epoch ms):

```
let dateAsUnixTimestamp = Int64(((date?.timeIntervalSince1970 ?? Date().timeIntervalSince1970) * 1000).rounded())
[...]

Issue #1071, opened 2026-04-07
Serializing glucose `date` as string instead of int


kingst:
_Copied from https://github.com/nightscout/Trio-dev/issues/539_

In our Swift code, when we serialize `GlucoseStorage` entries to JSON we convert the unix time stamp, `date`, to a String. At least some of the JS code expects this to be an int, like in [cob.js](https://github.com/nightscout/trio-oref/blob/37896e5306f8dcf4dfb1d6f4140828df6571f39e/lib/determine-basal/cob.js#L31). This code uses the first glucose entry directly for the `bucket_data` element, and downstream code tries to convert `date` to a Date assuming it's a timestamp, which causes an error. I don't think this is a big error, but it does cause an inconsistency nonetheless. 

Here is a [link to the code we would want to change](https://github.com/nightscout/Trio-dev/blob/a1577e581576772d970ffcc9df032720e7893f90/Model/Helper/GlucoseStored%2Bhelper.swift#L154). I don't know the full impact of this change but we should look into changing it.

PR #1124, opened 2026-04-30
Use Int instead of String for date in AlgorithmGlucose

kingst:
This PR fixes an encoding mistake when we convert AlgorithmGlucose to JSON where we use an Int instead of a String to represent the `date` property. It also includes deleting an unused helper function from GlucoseStorage that has the same mistake.

Fixes: https://github.com/nightscout/Trio/issues/1122 and https://github.com/nightscout/Trio/issues/1071

### DIFF
--- a/Model/Helper/GlucoseStored+helper.swift
+++ b/Model/Helper/GlucoseStored+helper.swift
@@ -124,42 +124,6 @@ extension NSPredicate {
     }
 }
 
-extension GlucoseStored: Encodable {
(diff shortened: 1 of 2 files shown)


dnzxy (approved):
Approve based on code review. 
I‘ll wait for @mountrcg to test it in-vitro before merging!

mountrcg:
Merged it into Tai SwiftOref -> no `currentDeviation` drift between JS and SwiftOref. All good. Thank you.

PR #1130, opened 2026-05-01
Show in-progress Bolus on Treatments screen (#1003)

### LINKED ISSUE #1003: Visualization Upgrade to "Bolus in Progress" Button in Treatments
I have an efficiency issue when trying enact a treatment while an existing bolus is in progress. Bolus/SMBs/UAMs make the user have to wait till completion for an additional treatment to be added. Depending on the current bolus size this can be frustrating because the user can't see the duration remaining on the bolus in progress. I often have to decide if it's worth it to exit the treatment screen to see the bolus in progress only to cancel the bolus and reenter the meal to unify the treatments. Perhaps I need more trust in a "log carbs without treatment" but I want to be done mentally with the whole treatment going into a meal. 

I think it would be helpful to have the "Bolus in Progress"  button contain a progress visualizer like a download progress bar. This would give the user a better understanding of wait times when deciding how to move forward with a concurrent treatment.  

Obviously, it would be nice to have the ability to add a treatment to a current bolus in production but I am aware this a pump hardware limit. Another solution could be to have a Bolus Que that would stack a treatments and deliver when the previous is finished. 

Image

dnzxy:
Replaces static 'Bolus In Progress…' disabled button in Treatment view with a Home view style card (cross.vial icon, 'Bolusing X of Y U' label, progress bar, inline cancel button), so users can see remaining bolus time / percentage without leaving Treatment view.

- Extract shared BolusProgressBar from HomeRootView into Trio/Sources/Views
- Add subscription to apsManager.bolusProgress to TreatmentStateModel
- Add bolus progress control to Treatment View
- Add ability to cancel bolus from Treatment View
- Addresses #1003

| Home View | Treatment View |
|--------|--------|
| image | image | 

https://github.com/user-attachments/assets/45a7e679-6d02-4e77-a893-b4e7b9b1e576

### DIFF
--- a/Trio.xcodeproj/project.pbxproj
+++ b/Trio.xcodeproj/project.pbxproj
@@ -190,6 +190,7 @@
 		38E98A3725F5509500C0CED0 /* String+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38E98A3625F5509500C0CED0 /* String+Extensions.swift */; };
 		38EA05DA261F6E7C0064E39B /* SimpleLogReporter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38EA05D9261F6E7C0064E39B /* SimpleLogReporter.swift */; };
 		38EA0600262091870064E39B /* BolusProgressViewStyle.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38EA05FF262091870064E39B /* BolusProgressViewStyle.swift */; };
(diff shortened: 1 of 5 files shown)


AndreasStokholm (approved):

kristinkand:
Really nice! Worked well we tried it at bf. 👍

bjorkert (approved):

PR #1140, opened 2026-05-06
fix: update basal rounding for 0.01U/hr pumps

bastiaanv:
fixes https://github.com/bastiaanv/DanaKit/issues/40

This fixes the double rounding issue for all pumps with a 0.01U/hr step (currently only Dana pumps)

### DIFF
--- a/Trio/Sources/Modules/Onboarding/View/TherapySettingEditorView.swift
+++ b/Trio/Sources/Modules/Onboarding/View/TherapySettingEditorView.swift
@@ -7,6 +7,13 @@ struct TherapySettingEditorView: View {
     var valueOptions: [Decimal]
     var validateOnDelete: (() -> Void)?
     var onItemAdded: (() -> Void)?
+    
+    private let basalFormatter: NumberFormatter = {
+        let numberFormatter = NumberFormatter()
+        numberFormatter.maximumFractionDigits = 2
+        numberFormatter.minimumFractionDigits = 2
+        return numberFormatter
+    }()
 
     @State private var selectedItemID: UUID?
     @Namespace var bottomID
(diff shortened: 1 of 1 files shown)


Trio/Sources/Modules/Onboarding/View/TherapySettingEditorView.swift
```diff
@@ -8,6 +8,13 @@ struct TherapySettingEditorView: View {
     var validateOnDelete: (() -> Void)?
     var onItemAdded: (() -> Void)?
 
+    private let basalFormatter: NumberFormatter = {
+        let numberFormatter = NumberFormatter()
+        numberFormatter.maximumFractionDigits = 2
```
MikePlante1, line 13: Does this take into consideration Minimed x23 and x54 pumps, which can iterate basal in 0.025 U/hr increments (3 digits).

bastiaanv: I do not think so (wauw, that is really small). Could you test this? I assume it will round to 0.03U/hr

MikePlante1: I would much prefer History to reflect the unrounded rate, though. I did choose to keep SMBs rounded to 2 decimal places on the graph, but left the values unrounded in History for bolus and basal.

I remember the rounding being an issue for Minimed before, but it got fixed a while ago. I'll try to look into where that got fixed. (showing things like 0.4999999999 U/hr in History)

MikePlante1: Oh, I think I'm confusing two different issues, but the basal rate picker should also show the full unrounded rate for Minimed.

I wanna say it happened because of a Decimal vs Double conflict before, though? (Still looking)

MikePlante1:
EDIT: Nevermind, suggestion deleted as it would cause issues for people who are already using the wrongly-rounded values.

dnzxy:
So, are we good to proceed with this from your perspective, @MikePlante1 ?

MikePlante1:
I tested it with `numberFormatter.maximumFractionDigits = 3` and that'll work fine. I tested it in a simulator with a simulator pump (switching between Minimed and Dana)

| Dana | MiniMed x23/x54 |
|---|---|
|  |  |

dnzxy:
@bastiaanv was this a mistakenly closed PR due to branch deletion?

PR #1145, opened 2026-05-11
APSManager thread safety fixes

marv-out:
## Summary

  Hardens `APSManager` against several latent concurrency bugs around loop
  serialization, bolus-progress state, and the background-task lifecycle.
  None of these reproduced reliably in normal use, but each is a real race
  that could manifest under timing pressure (concurrent heartbeats, fast
  bolus end-then-start cycles, app backgrounding mid-loop).

  No public API changes. Single file touched (`Trio/Sources/APS/APSManager.swift`).

  Three concrete patterns in pre-PR `APSManager` were not safe under
  concurrent access:

  1. **Loop serialization was check-then-act on a `CurrentValueSubject`.**
     `canStartNewLoop()` read `isLooping.value`; if false, the caller then
     called `isLooping.send(true)`. Two concurrent `recommendsLoop` events
     could both observe `false` before either had sent `true`, and the
     loop would run twice.

  2. **`bolusReporter` was a plain mutable property mutated from a Combine
     sink on `processQueue` and from `cancelBolus()`'s async context.**
     No isolation; a data race on every reporter swap.

  3. **`backgroundTaskID` was instance state shared between the loop
     coroutine and the expiration handler.** The handler nulled it while
     the loop was still using it; cleanup paths were duplicated and
     inconsistent.

  ## What changed

  ### 1. Loop serialization via an actor

  **Before:**

  ```
private func canStartNewLoop() async -> Bool {
      if lastLoopDate > lastLoopStartDate {
          guard lastLoopStartDate.addingTimeInterval(Config.loopInterval) < Date() else { return false }
      }
      guard !isLooping.value else { return false }
      return true
  }

  // caller:
  guard await canStartNewLoop() else { return }
  let (loopStatRecord, backgroundTask) = await setupLoop()  // sends isLooping(true) here
```

  **After**:

```
  private actor LoopGuard {
      private var isRunning = false

      func tryStart(minInterval: TimeInterval, lastLoopDate: Date, lastLoopStartDate: Date) -> Bool {
          if lastLoopDate > lastLoopStartDate {
              guard lastLoopStartDate.addingTimeInterval(minInterval) < Date() else { return false }
          }
          guard !isRunning else { return false }
          isRunning = true
          return true
      }

      func finish() { isRunning = false }
  }

  // caller:
  guard await loopGuard.tryStart(...) else { return }
```

  The check-and-set is now a single atomic operation on the actor. `isLooping` becomes a downstream notification mechanism only — never the source of truth for "can I start?".

  ### 2. Bolus progress state via an actor + generation token

  **Before**:
```

  private var bolusReporter: DoseProgressReporter?  // mutated from multiple contexts

  private func createBolusReporter() {
      bolusReporter = pumpManager?.createBolusProgressReporter(reportingOn: processQueue)
      bolusReporter?.addObserver(self)
  }

  private func clearBolusReporter() {
      bolusReporter?.removeObserver(self)
      bolusReporter = nil
      processQueue.asyncAfter(deadline: .now() + 0.5) { self.bolusProgress.send(nil) }
  }
```

  Bugs:
  - Data race on bolusReporter between the bolusTrigger sink and cancelBolus.
  - The 500 ms delayed send(nil) could clobber a fresh send(0) if a new bolus started during the grace period.

  **After**:

 ```
 private actor BolusProgressState {
      private var reporter: DoseProgressReporter?
      private var generation: UInt64 = 0

      func setReporter(_ newReporter: DoseProgressReporter?, observer: any DoseProgressObserver) {
          generation &+= 1
          reporter?.removeObserver(observer)
          reporter = newReporter
          reporter?.addObserver(observer)
      }

      func clear(observer: any DoseProgressObserver) -> UInt64 {
          generation &+= 1
          reporter?.removeObserver(observer)
          reporter = nil
          return generation
      }

      func isCurrent(_ token: UInt64) -> Bool { token == generation }
  }

```
  create / clear are funneled through a single Task per bolusTrigger event so they cannot reorder against each other at the actor:

  ```
deviceDataManager.bolusTrigger
      .receive(on: processQueue)
      .sink { [weak self] bolusing in
          Task { [weak self] in
              guard let self else { return }
              if bolusing { await self.createBolusReporter() }
              else        { await self.clearBolusReporter() }
          }
      }
```

  The delayed `send(nil)` is gated by the generation token: if a new `setReporter` arrives during the 500 ms sleep, the token no longer matches and the stale nil is dropped.

  The observer is passed in on every call rather than cached. A cached weak ref could nil out between `setReporter` and `clear` (leaking the observer through whatever internal list `DoseProgressReporter` holds); caching it strong would form a retain cycle `BaseAPSManager → state → observer = BaseAPSManager`.

  ### 3. Background task lifecycle

  **Before**: `backgroundTaskID` was a shared property; the expiration handler mutated it from a separate Task, and the cleanup paths in loop() and loopCompleted() both touched it.

  **After**: the background-task identifier is a local in the `loop()` async scope. It is started right after the actor admits the loop and explicitly awaited at the end of the do/catch:

 ```
 let bgTask = await UIApplication.shared.beginBackgroundTask(withName: "Loop starting")
  isLooping.send(true)

  do {
      try await executeLoop(...)
      await finalizeLoop(...)
  } catch { ... }

  if bgTask != .invalid {
      await UIApplication.shared.endBackgroundTask(bgTask)
  }
```

  ### 4. Removed viewContext use

  The dead reference to CoreDataStack.shared.persistentContainer.viewContext in APSManager was removed. Nothing in this manager ever touches the view context — only privateContext.

  ### 5. `[weak self] `in subscribe sinks

  Most subscriptions in `subscribe()` already used `[weak self]`. The remaining ones now also do.

  Deliberate behavior changes

  These are intentional and worth flagging during review:

  - expirationHandler dropped. The pre-PR handler ended the background task from a follow-up Task, which is itself buggy for the same reason described in §3. Reintroducing a correct handler is out of scope; for now, if iOS expires the task before the loop finishes, the loop coroutine will complete on best-effort. This is not a regression in the negative sense — the previous handler did not provide a real guarantee either.
  - `lastLoopStartDate` is still written outside the actor. Today this is safe because only one caller wins `tryStart`, so only one writer is ever active at once. Any future "force loop" code path that bypasses `tryStart` would break this invariant — flagged here so it doesn't become a footgun.
  - `canStartNewLoop`, `setupLoop`, `loopCompleted` removed. Folded into `loop()` and `finalizeLoop()` to keep the lifecycle contained.


Trio/Sources/APS/APSManager.swift
```diff
@@ -1207,15 +1205,15 @@ final class BaseAPSManager: APSManager, Injectable {
     }
 
     private func createBolusReporter() {
-        bolusReporter = pumpManager?.createBolusProgressReporter(reportingOn: processQueue)
-        bolusReporter?.addObserver(self)
+        let reporter = pumpManager?.createBolusProgressReporter(reportingOn: processQueue)
+        Task { await bolusProgressState.setReporter(reporter, observer: self) }
```
dnzxy: Two unstructured Tasks for create / clear can reorder against each other at the actor.

`bolusTrigger` in practice also fires false→true (bolus end immediately before a new bolus). The Tasks leave the sink in spawn order but enter the actor in scheduler order — not guaranteed FIFO. Result: `setReporter` lands before its matching `clear`, the new reporter gets wiped and stays `nil`.

On top of that, the 500 ms delayed `send(nil)` in `clearBolusReporter` (line 1216): if a new bolus starts within those 500 ms, the late `send(nil)` clobbers its `bolusProgress.send(0)`. Old `processQueue.asyncAfter(0.5)` had the same bug, but since we're touching this anyway it should go.

Fix: make both methods `async` and `await` them inside a single `Task` per sink event — order is then guaranteed. For the delayed send, add a generation token inside the actor so a stale send can't clobber an active bolus.

Trio/Sources/APS/APSManager.swift
```diff
@@ -233,116 +278,80 @@ final class BaseAPSManager: APSManager, Injectable {
...
+
+            var loopStatRecord = LoopStats(
+                start: loopStartDate,
+                loopStatus: "Starting",
+                interval: interval
+            )
 
+            defer {
+                if bgTask != .invalid {
+                    Task { await UIApplication.shared.endBackgroundTask(bgTask) }
```
dnzxy: Two things here:

1. Calling `endBackgroundTask` from a follow-up unstructured Task is not guaranteed to complete before iOS suspends the app. Apple's contract is `endBackgroundTask` must land before suspension / before the expiration handler returns. The enclosing scope is already async — defer doesn't need a separate Task if we instead use `do { … } catch { … }` and explicitly `await UIApplication.shared.endBackgroundTask(bgTask)` at the end of both paths.

2. No `expirationHandler` anymore. The old one was buggy (also ended the task from a Task), but dropping it is a deliberate behavior change and belongs in the PR description.

Trio/Sources/APS/APSManager.swift
```diff
@@ -354,34 +363,25 @@ final class BaseAPSManager: APSManager, Injectable {
         }
     }
 
-    // Loop exit point
-    private func loopCompleted(error: Error? = nil, loopStatRecord: LoopStats) async {
+    /// Single exit point for loop — replaces the old `loopCompleted()`.
+    private func finalizeLoop(error: Error? = nil, loopStatRecord: LoopStats) async {
         isLooping.send(false)
+        await loopGuard.finish()
```
dnzxy, line 369: Swap order: `await loopGuard.finish()` first, then `isLooping.send(false)`.

As written, subscribers (Home UI, `deviceDataManager.loopInProgress`) see the loop as finished before the actor is actually free. A heartbeat in that exact window triggers `recommendsLoop` → `loop()` → waits on the actor → `tryStart` returns `false` and logs "Loop skipped (already running or too soon)". No functional harm, but unnecessary log noise at the tail of every loop.

Trio/Sources/APS/APSManager.swift
```diff
@@ -70,6 +70,52 @@ enum APSError: LocalizedError {
...
+
+// MARK: - Thread-safe bolus progress state
+
+/// Isolates the mutable `DoseProgressReporter` reference to prevent
+/// data races between Combine sinks (processQueue) and async contexts (cancelBolus).
+private actor BolusProgressState {
+    private var reporter: DoseProgressReporter?
+    private weak var observer: (any DoseProgressObserver)?
+
+    func setReporter(_ newReporter: DoseProgressReporter?, observer: any DoseProgressObserver) {
```
dnzxy, line 103: `setReporter` takes `observer` as a parameter, `clear` reaches for the stored weak ref. Functionally fine while the caller is always `self` — but: if `BaseAPSManager` goes away between `setReporter` and `clear`, the weak ref is `nil` and `removeObserver` is skipped. `DoseProgressReporter` may keep a strong reference to the dead observer internally → leak.

Either hold the observer non-weak (its lifetime is already bounded by `BaseAPSManager`) and drop the parameter, or make `clear` tear the reporter down even without a stored observer.

dnzxy (changes requested):
Direction is right: actor gate on the loop, `[weak self]` in the sinks, dropping the dead `viewContext` and the shared `backgroundTaskID` are all good.

Four things should go before merge, see my inline comments:

Also: please fill the PR description 😄

Trio/Sources/APS/APSManager.swift
```diff
@@ -70,6 +70,62 @@ enum APSError: LocalizedError {
...
+        guard !isRunning else { return false }
+        isRunning = true
+        return true
+    }
+
+    func finish() {
+        isRunning = false
+    }
+
+    var running: Bool { isRunning }
```
dnzxy: This is not used anywhere and could potentially be dropped.

Trio/Sources/APS/APSManager.swift
```diff
@@ -632,8 +653,7 @@ final class BaseAPSManager: APSManager, Injectable {
                 )
             )
         }
-        bolusReporter?.removeObserver(self)
-        bolusReporter = nil
+        _ = await bolusProgressState.clear(observer: self)
```
dnzxy, line 634: This assignment could be dropped.
We are awaiting a command to the pump, then sending something through the observer. 
This could _potentially_ open another "gap" where commands could come "in between".
Chances are probably super low for this, so I'd leave it.

dnzxy (commented):
I think the recent changes have addressed the earlier change requests. 
Let's maybe scale back on the Claude-driven big chunks of comments?
Two small nit comments – those do not alter or block any functionality though.

Trio/Sources/APS/APSManager.swift
```diff
@@ -233,116 +282,77 @@ final class BaseAPSManager: APSManager, Injectable {
...
                 )
+                await finalizeLoop(loopStatRecord: loopStatRecord)
             } catch {
-                var updatedStats = loopStatRecord
-                updatedStats.end = Date()
-                updatedStats.duration = roundDouble((updatedStats.end! - updatedStats.start).timeInterval / 60, 2)
-                updatedStats.loopStatus = error.localizedDescription
-                await loopCompleted(error: error, loopStatRecord: updatedStats)
+                loopStatRecord.end = Date()
+                loopStatRecord.duration = roundDouble((loopStatRecord.end! - loopStatRecord.start).timeInterval / 60, 2)
```
kingst, line 318: I know that this forced unwrap was already here but we should remove it.

Trio/Sources/APS/APSManager.swift
```diff
@@ -233,116 +282,77 @@ final class BaseAPSManager: APSManager, Injectable {
...
-            loopStatRecord.duration = roundDouble((loopStatRecord.end! - loopStatRecord.start).timeInterval / 60, 2)
-            loopStatRecord.loopStatus = "Success"
-            await loopCompleted(loopStatRecord: loopStatRecord)
-            return
-        }
-
-        // Handle closed loop
-        try await enactDetermination()
         loopStatRecord.end = Date()
         loopStatRecord.duration = roundDouble((loopStatRecord.end! - loopStatRecord.start).timeInterval / 60, 2)
```
kingst, line 334: I know that you didn't change this loc but since you're here let's get rid of this forced unwrap.

kingst (commented):
There are some small changes here, which look great:
- remove unused viewContext
- move tir() out of a nested perform
- add missing [weak self]

And some race condition fixes that also look great around receiving values on the main queue, which fixes real race conditions.

There are two forced unwraps that I flagged that should be fixed. From looking at the code it's clear that these will always have values, but as a general principle I don't think we should have forced unwraps in our code base. Note: these existed before your changes.

My suggestion is to trim down this PR to handle all of the above, which I'm ready to 👍 now. Then in two subsequent PRs:
- LoopGuard + background task handling
- bolus progress actor

The core LoopGuard mechanism looks good, but I have feedback on the background task handling, and it'll make things easier to focus on these in a separate PR and there is a clear path to make some small changes and get this in quickly.

The bolus progress actor is probably a bit of a longer discussion where I have some architectural feedback around the approach overall, so should also be in its own PR.

marv-out:
Done! I will close this PR in favor of PR #1225 #1226 and #1227

PR #1151, opened 2026-05-13
Fix spurious 'Bolus failed' notification after manual bolus

### LINKED ISSUE #1150: Bolus error messages during CGM warm-up
## Describe the bug
When doing a manual bolus during CGM warm-up Trio throws error message notifications even though the bolus goes through successfully.

| pic1 | pic2 |
| -- | -- |
| Image | Image |

## Attach a Log
[bolus_failed_notification_log.txt](https://github.com/user-attachments/files/27692743/bolus_failed_notification_log.txt)

## Expected behavior
no error notifications for successful manual bolus

## pre-Analysis
Log evidence (around 19:43:35Z):

- APSManager.swift:648 — Enact bolus 0.6, manual true
- Pump confirms delivery: New pump status Bolus: inProgress(... value: 0.6 ...) then deliveredUnits: Optional(0.6)
- Then APSManager.swift:517 — determineBasal() ... Not enough glucose data (G6 warm-up, no CGM readings)
- Then APSManager.swift:659 — Bolus failed with error: apsError(... Not enough glucose data ...)
- Then Sending noLoopFirstNotification notification for Bolus failed and the pumpError "Invalid Algorithm Response" toast.

So the bolus did go through — but the post-bolus determineBasalSync() (which Trio runs after a manual bolus to refresh recommendations) threw because the G6 was warming up with <3 glucose readings in 6h. That throw landed in the same do/catch as the bolus call, so the user sees "Bolus failed" + the pumpError, even though delivery succeeded.

2024-05-31 — a80425fa "refactoring in apsManager to use async await…"  The async/await rewrite moved the post-bolus call inside the bolus do/catch, but kept it non-throwing:

```
if !isSMB {
//                determineBasal()
    determineBasalSync()   // not thrown
}
```
At this point, errors from determineBasalSync() were not propagated, so a missing-glucose case wouldn't surface as a "Bolus failed" toast.

2024-08 (PR [#50](https://github.com/mountrcg/Trio-dev/issues/50), f18f79d7 "Async rework 2.0"): changed to await determineBasalSync() — still non-throwing.

2025-02-18 — 902dbe01 "Fixes". Single one-line change that turned it into try await determineBasalSync(). That made any throw from the algorithm path (including "Not enough glucose data") fall through to the bolus catch, which fires bolusDidFail() and the pumpError notification — even though the pump already delivered the bolus successfully.

So the regression entered in commit 902dbe01 on 2025-02-18 (in a broad "Fixes" commit, no targeted intent visible). Before that, post-bolus algorithm failures during e.g. G6 warm-up were silently swallowed; after, they masquerade as bolus failures.

mountrcg:
fixes #1150 
The post-bolus determineBasalSync() call was inside the bolus do/catch, so any algorithm error (e.g. 'Not enough glucose data' during CGM warm-up) was misreported as a bolus failure even though the pump had already delivered the bolus successfully. Wrap the algorithm refresh in its own do/catch so it cannot trigger bolusDidFail or a pumpError notification.


mountrcg:
the stale Glucose value error still pops up, but no more failed bolus notifications

dnzxy (commented):
While this change in principle fixes the issue that @mountrcg was seeing, it does actually  just uncover (or mask, depends on how we look at it), that we need to be able to run specific oref modules indepedently of each other.

In the specific bolus case here, we actually don't want / need to run the entirety of oref, but just IOB – especially because in this case we are mid-loop-cycle. 

So I'm not sure if this is a good approach to go ahead with the proposed fix, or rather only run IOB when a user triggers a bolus command through treatment view (i.e., manual pump bolus or external insulin dose).

@kingst @marv-out thoughts?

kingst:
We should fix this but I agree that adding a `updateIoB` function to OpenAPS and APSManager for it.

In OpenAPS you'd copy the logic from `determineBasal` that does this:

```swift
// IOB calculation
let iob = try await self.iob(
    pumphistory: pumpHistoryJSON,
    profile: profile,
    clock: clock,
    autosens: autosens.isEmpty ? .null : autosens,
    useSwiftOref: useSwiftOref
)

// TODO: refactor this to core data
storage.save(iob, as: Monitor.iob)
```

and in APSManaager you call this new OpenAPS function and notify any listeners:

```swift
iobFileDidUpdate.send(())
```

You don't need to wait for the oref swift port, this should work at the layer above.

dnzxy:
@kingst on second thought, if we split IOB update and determine basal after a manual bolus, we do not get an updated forecast post-bolus. I think we want to keep that, don't we? 

I'd just go ahead with the PR as-is. We can potentially address this in oref-swift with the new `ForecastGenerator`, where we can get forecasts and update IOB without running the entirety of oref.

dnzxy (approved):
Discussed this in private with @kingst and @marv-out .
We'll move forward with this as-is. 
Thanks for finding this and offering a fix @mountrcg 🙏

PR #1159, opened 2026-05-17
Fix: Floor CGM glucose at 39 mg/dL in GlucoseStorage

dnzxy:
## Summary

Addresses nightscout/Trio-dev#631 — since LibreTransmitter ignores Abbott given limits and propagates raw glucose, oref can receive values below the official 39 mg/dL CGM floor (LibreTransmitter explicitly bypasses the Libre algorithm's floor, forwarding values down to 1 mg/dL). On dev those sub-39 readings are silently dropped by the JS oref glucose-get-last filter (>38), leaving determination with no fresh BG and the pump stuck in whatever temp it was last on. The reporter saw ~1h 7min of no Swift determinations with a BG nadir of 23 mg/dL.

This PR adds a defensive chokepoint in BaseGlucoseStorage:

  - New Config.minimumGlucose = 39 and a private clampToMinimum(_:) helper that clamps both glucose and sgv on each
  BloodGlucose, emitting debug(.deviceManager, ...) with the raw value and timestamp so the original reading is preserved
  in support logs.
  - storeGlucose and backfillGlucose both run their input through clampToMinimum before any filtering or persistence.
  storeCGMState receives the clamped array too.
  - addManualGlucose is intentionally left untouched — user-entered fingerstick values are not silently rewritten.

  The floor aligns with BloodGlucose.isStateValid (already sgv >= 39) and FetchGlucoseManager.minimumSmoothedGlucose: 39.

### DIFF
--- a/Trio/Sources/APS/Storage/GlucoseStorage.swift
+++ b/Trio/Sources/APS/Storage/GlucoseStorage.swift
@@ -41,6 +41,7 @@ final class BaseGlucoseStorage: GlucoseStorage, Injectable {
 
     private enum Config {
         static let filterTime: TimeInterval = 3.5 * 60
+        static let minimumGlucose: Int = 39
     }
 
     private let context: NSManagedObjectContext
@@ -75,10 +76,11 @@ final class BaseGlucoseStorage: GlucoseStorage, Injectable {
     ///  it isn't within 3.5 minutes of an existing glucose reading, which is simple but not perfect.
     ///  But since this is a corner case that really shouldn't happen often, it's good enough.
     func backfillGlucose(_ glucose: [BloodGlucose]) async throws {
+        let clamped = clampToMinimum(glucose)
         try await context.perform {
             // remove already deleted glucose values
             let withoutDeletedGlucose = self.filterGlucoseValues(
-                glucose,
+                clamped,
                 fetchRequest: DeletedGlucoseStored.fetchRequest(),
                 timeBuffer: 1
             )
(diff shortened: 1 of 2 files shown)


kingst (commented):
The code generally looks good and I think this is a good change to make. Two clarifying questions:
- If we get a manually entered glucose value of 38 will that cause issues?
- Did you consider putting the clamp in the algorithm side, leaving the storage unaltered? I can see it both ways, but it would be worth understanding your rationale for picking the storage layer.

dnzxy:
Thanks for the review and very valid questions @kingst .

For manual glucose, we currently have this input validation guard in place
```swift
let limitLow: Decimal = state.units == .mgdL ? Decimal(14) : 14.asMmolL
let limitHigh: Decimal = state.units == .mgdL ? Decimal(720) : 720.asMmolL
```
The lower limit is in line with glucometers, at least the ones that I have used and know.
The upper limit is quite too high, I don't think any glucometer goes beyond 500. 
This technically leaves another potential spot where we can run into this issue, but one I felt was not worth touching, plus we want to store actual precise blood glucose measurements, not clamp them beyond what manufacturers clamp.
In terms of Libre, the CGM driver is actually ignoring an OEM limit and forwarding raw readings; something the official app wouldn't / shouldn't do. xDrip4iOS limits Libre readings to 39 for example, so Libre through xDrip does not have this issue. 

My rationale for adding it to storage is that we don't want "faulty" CGM readings as part of our database (that is also feeding external services), and it saves us from having to clamp in many places, and keeps it pretty centralized.

kingst (approved):
Agreed with your rationale, so the PR LGTM. If we did want to avoid 38 values we could add something like `sensorErrorGlucose = 38` to the `Config` struct and if someone tries to add a manual reading of 38 we change it to 39 before storing it. A change of 1 mg/dl is still well within the meter's error bounds and helps us avoid this algorithmic magic number. but as I'm typing it out it feels very hacky, so I'm fine with either one but if we leave it let's add an issue to fix this later.

PR #1164, opened 2026-05-23
Fix Apple Health flag by checking for sample type permission

dnzxy:
A tester reported appleHealthEnabled: false despite a working Apple Health integration.

Cause: `hasGrantedFullWritePermissions` required all five Trio-managed sample types (glucose, insulin, carbs, fat, protein) to be authorized.

Changes in `TelemetryClient.buildPayload`: 
- `appleHealthEnabled` → true when any write permission is granted (matches the state-based nightscoutPaired /
  tidepoolPaired pattern already in place).
- New `appleHealthWrites: { glucose, insulin, carbs, fat, protein }`check per-type authorization snapshot. Reuses the existing `checkWriteToHealthPermissions` of HealthKitManager.

### DIFF
--- a/Trio/Sources/Services/Telemetry/TelemetryClient.swift
+++ b/Trio/Sources/Services/Telemetry/TelemetryClient.swift
@@ -1,4 +1,5 @@
 import Foundation
+import HealthKit
 import LoopKit
 import Swinject
 import UIKit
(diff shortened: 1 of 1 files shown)


AndreasStokholm:
Seems to work correctly with the new properties. On my test phone they all started out as `false`. I then switched them one by one and they all reflected correctly in the "What's sent" screen.

AndreasStokholm (approved):

Issue #920, opened 2026-01-03
unexpected COB rise


redheadlawyer:
## Describe the bug
COB rise after carb entry to an amout larger than the original entered carbs; this leads to overboluses; happened on multiple occasions 

## To Reproduce
enter carbs and most times the bs drops afterwards

## Expected behavior
carbs decrease overtime and won't exceed the entered amount

## Screenshots

![Image](https://github.com/user-attachments/assets/b6ebc479-e61d-43e2-9785-03f5d1cc6c39)

### Smartphone:
* Trio v0.6.0.34
* iphone 17 pro, iOS 26.0.1

### Pump:
* Omnipod Dash

### CGM:
* Dexcom G6
* Manager app: Dexcom

### Trio Version:
* Version Number: *[e.g. 1.9.2]*
* Repo: *nightscout/trio*
* Git Reference: *[e.g. commit hash]*

## Technical Details
*If applicable, provide any technical details that might help in diagnosing the problem. This could include logs, error messages, or relevant configuration details.*

## Additional context
*Add any other context about the problem here.*

dnzxy:
Logs?

redheadlawyer:
[log_prev.txt](https://github.com/user-attachments/files/24421242/log_prev.txt)
[log.txt](https://github.com/user-attachments/files/24421241/log.txt)

> Logs?

Blanderson86:
[log_prev.txt](https://github.com/user-attachments/files/24919701/log_prev.txt)
[log.txt](https://github.com/user-attachments/files/24919700/log.txt)

Per Theresa’s request on facebook

kingst:
Thanks, @Blanderson86  do you ever enter protein or fat macronutrients along with carbs?

kingst:
Also @Blanderson86 do you have glucose "smoothing" enabled?

dnzxy:
@Blanderson86 shared this screenshot on Facebook with the below post: 

<img width="738" height="1600" alt="Image" src="https://github.com/user-attachments/assets/3ce2b49b-0ef5-40a0-b670-e25abe7f2dc0" />

[...]

PR #1171, opened 2026-05-27
Change fetchLimit for glucose oref input to be dynamic based on maxMealAbsorptionTime #920

dnzxy:
## Summary
The glucose history fed into the meal/COB algorithm was capped at a fixed fetchLimit: 72 (72 readings × 5 min = 6h).
That constant predates the user-selectable maxMealAbsorptionTime (MMAT) setting, which ranges 4–10h. When MMAT exceeds 6h, the COB calculation considers carb absorption out to MMAT hours but only has 6h of glucose to work with.

The result is the bug reported in #920: mealCOB rises after a carb entry (to an amount larger than the carbs entered)
as larger absorption buckets slide out the back of the 6h window and are replaced by smaller front buckets, then crashes to zero when the meal is evicted at mealTime + MMAT. This drives over-bolusing and hypo risk.

This affects both the JS oref0 path and the native oref-swift port, since the limit is applied upstream in OpenAPS.determineBasal before either path runs.

The fix makes the fetch limit dynamic: ceil(maxMealAbsorptionTime × 12) readings, so the glucose window always covers the full absorption window. Preferences are loaded up front (the existing async load is just hoisted and reused) to derive the limit before the parallel fetches. Tying it to the setting rather than a static cap means it can't silently drift again if the MMAT range changes.

Credit to @robertjwellwood for the diagnosis and simulation in #920.

## To Test

  - Build and run with default MMAT (6h), then verify COB behaves as before
  - Set MMAT to 8–10h, enter a carb dose, and confirm COB decays smoothly toward 0 without rising above the entered amount or cliff-dropping
  - Verify both useSwiftOref on and off produce consistent COB trajectories (so JS-based oref and native Swift port)
  - Confirm glucoseFetchLimit scales as expected (4h→48, 6h→72, 10h→120)

### DIFF
--- a/Trio/Sources/APS/OpenAPS/OpenAPS.swift
+++ b/Trio/Sources/APS/OpenAPS/OpenAPS.swift
@@ -393,13 +393,20 @@ final class OpenAPS {
         // Perform asynchronous calls in parallel
         async let pumpHistoryObjectIDs = fetchPumpHistoryObjectIDs() ?? []
         async let carbs = fetchAndProcessCarbs(additionalCarbs: simulatedCarbsAmount ?? 0, carbsDate: simulatedCarbsDate)
-        async let glucose = fetchAndProcessGlucose(context: context, shouldSmoothGlucose: shouldSmoothGlucose, fetchLimit: 72)
+
+        var preferences = await storage.retrieveAsync(OpenAPS.Settings.preferences, as: Preferences.self) ?? Preferences()
+        let glucoseFetchLimit = Int((preferences.maxMealAbsorptionTime * 12).rounded(scale: 0, roundingMode: .up))
+        async let glucose = fetchAndProcessGlucose(
+            context: context,
+            shouldSmoothGlucose: shouldSmoothGlucose,
+            fetchLimit: glucoseFetchLimit
+        )
+
         async let prepareTrioCustomOrefVariables = prepareTrioCustomOrefVariables()
         async let profileAsync = loadFileFromStorageAsync(name: Settings.profile)
         async let basalAsync = loadFileFromStorageAsync(name: Settings.basalProfile)
         async let autosenseAsync = loadFileFromStorageAsync(name: Settings.autosense)
         async let reservoirAsync = loadFileFromStorageAsync(name: Monitor.reservoir)
-        async let preferencesAsync = storage.retrieveAsync(OpenAPS.Settings.preferences, as: Preferences.self) ?? Preferences()
         async let hasSufficientTddForDynamic = tddStorage.hasSufficientTDD()
 
         // Await the results of asynchronous tasks
@@ -449,8 +456,6 @@ final class OpenAPS {
         if !simulation {
             storage.save(iob, as: Monitor.iob)
         }
-
-        var preferences = await preferencesAsync
 
         if !hasSufficientTdd, preferences.useNewFormula || (preferences.useNewFormula && preferences.sigmoid) {
             debug(.openAPS, "Insufficient TDD for dynamic formula; disabling for determine basal run.")
--- a/Trio/Sources/APS/OpenAPS/OpenAPS.swift
+++ b/Trio/Sources/APS/OpenAPS/OpenAPS.swift
@@ -99,13 +99,20 @@ final class OpenAPS {
     func fetchAndProcessGlucose(
         context: NSManagedObjectContext,
         shouldSmoothGlucose: Bool,
-        fetchLimit: Int?
+        fetchLimit: Int?,
+        fetchHours: Decimal = 24
     ) async throws -> String {
-        // make it async and await it
+        // Time window from `fetchHours` hours ago up to now. determineBasal feeds
+        // `maxMealAbsorptionTime + 0.5h` (just enough glucose to cover the longest
+        // tracked meal absorption plus a small lead-in); Autosens uses the default
+        // 24h because its sensitivity algorithm needs that full window.
+        let cutoff = Date().addingTimeInterval(-(Double(truncating: fetchHours as NSNumber) * 3600))
+        let timePredicate = NSPredicate(format: "date >= %@", cutoff as NSDate)
+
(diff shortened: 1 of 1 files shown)


dnzxy:
Side note: I PR'd this to `feat/dev-oref-swift` as that branch will make it to `dev` once this gets tested and merged, so it kills 2 birds with one stone (fix for JS and fix for Swift port)

Trio/Sources/APS/OpenAPS/OpenAPS.swift
```diff
@@ -393,13 +393,20 @@ final class OpenAPS {
         // Perform asynchronous calls in parallel
         async let pumpHistoryObjectIDs = fetchPumpHistoryObjectIDs() ?? []
         async let carbs = fetchAndProcessCarbs(additionalCarbs: simulatedCarbsAmount ?? 0, carbsDate: simulatedCarbsDate)
-        async let glucose = fetchAndProcessGlucose(context: context, shouldSmoothGlucose: shouldSmoothGlucose, fetchLimit: 72)
+
+        var preferences = await storage.retrieveAsync(OpenAPS.Settings.preferences, as: Preferences.self) ?? Preferences()
+        let glucoseFetchLimit = Int((preferences.maxMealAbsorptionTime * 12).rounded(scale: 0, roundingMode: .up))
```
kingst: This code is implicitly assuming that we get readings every 5 minutes but since the algorithm will drop readings that are too old already and the fetchAndProcessGlucose method already limits to 24 hours, it might be cleaner to just drop the limits altogether and fetch the full 24 hour trace. It also brings back the parallel fetch mechanism.

kingst (commented):
LGTM, this is fixing a real bug and this change is the right way to go directionally. I have one minor comment that's worth discussing though.

Trio/Sources/APS/OpenAPS/OpenAPS.swift
dnzxy: I partially agree and disagree, or at least am unsure. It is fetching the last 120 readings, so actually would surpass the 10 hr window if readings are not available in (the assumed) 5min intervals. However, speaking for the native Swift port, in `MealCob.detectCarbAbsorption` we proceed to bucket over the entire glucose array, and we do not need more glucose to iterate over than the max meal absorption window, and so by passing in > 2x the "required" glucose, we bucket / iterate many more times than actually required every time the algo runs or the meal module runs. 
I haven't benchmarked this, so I don't know if it _really_ makes a _noticeable_ difference, but just want to mention it. I'd be okay with removing it altogether and just providing the already provided 24 hr limit.

kingst: We could pass a date into the fetch function if we wanted to keep things tight for the bucketing algorithm

robertjwellwood: Just a final thought from me that may be relevant to your decision whether to use count-based or date-based fetch, and whether to deliberately over-reach on the fetch to provide a buffer. The original `fetchLimit: 72`, even when matched with MMAT 6h, ends up 1 cycle short of fully spanning the MMAT. The effect is really small and really transient: it only ever loses the single oldest bucket of absorption, and it only lasts for the single loop immediately prior to MMAT. If you want to avoid this same short-span effect you would need to add a margin to `glucoseFetchLimit = Int((preferences.maxMealAbsorptionTime * 12).rounded(scale: 0, roundingMode: .up))`. Same goes for an MMAT date-based fetch. The problem doesn't exist with `fetchLimit: nil`.

dnzxy: I like @kingst 's suggestion of just adding a time-based condition to the predicate, then @robertjwellwood 's comment points to even that not being "enough", cause we need* MMAT window + 1 reading basically. 

How about this: we add a time-based condition and give it 0.5 hr (up to 6 readings more, if system runs nominally) buffer?

In a perfect world, that gives us for default MMAT 78 readings, for max. MMAT 126 readings, for min. MMAT 60 readings, and so on.

kingst: Yes I agree this sounds like the cleanest solution

robertjwellwood: That'll work.

dnzxy: Changed code accordingly.

Trio/Sources/APS/OpenAPS/OpenAPS.swift
```diff
@@ -393,13 +400,21 @@ final class OpenAPS {
         // Perform asynchronous calls in parallel
         async let pumpHistoryObjectIDs = fetchPumpHistoryObjectIDs() ?? []
         async let carbs = fetchAndProcessCarbs(additionalCarbs: simulatedCarbsAmount ?? 0, carbsDate: simulatedCarbsDate)
-        async let glucose = fetchAndProcessGlucose(context: context, shouldSmoothGlucose: shouldSmoothGlucose, fetchLimit: 72)
+
+        var preferences = await storage.retrieveAsync(OpenAPS.Settings.preferences, as: Preferences.self) ?? Preferences()
+        let glucoseFetchHours = preferences.maxMealAbsorptionTime * 12 + 0.5 // MMAT + half hour buffer
```
kingst: I believe you can drop the * 12 from this

kingst (approved):
LGTM. I'm approving this but can you double check that `* 12`? I believe it should just be:

```swift
let glucoseFetchHours = preferences.maxMealAbsorptionTime + 0.5 // MMAT + half hour buffer
```

Without the `* 12` applied to MMAT.

Trio/Sources/APS/OpenAPS/OpenAPS.swift
dnzxy: I shouldn't be doing this so late in the day 😂 
Will correct it, apologies!

PR #1172, opened 2026-05-28
Update telemetry payload: fix buildDate format, add locale/TZ/appName

jwoglom:
Fixes a few issues identified with the Trio telemetry payloads:

* On non-English locales, the buildDate was improperly formatted. e.g., a builder on a French language-default OS would emit `lun. mai 2`. Fix: force capture-build-details to POSIX default locale (english).
* Parse the timestamp and emit as YYYY-MM-DD instead of “Tue May 26” format
* Add iOS system locale + timezone to telemetry, with the above bug as motivation for why we should track this. Plus, this will be beneficial for users when managing translations.
* Add appName to telemetry -- hardcoded as "Trio", but can be overridden for forks

### DIFF
--- a/Trio/Sources/Services/Telemetry/TelemetryClient.swift
+++ b/Trio/Sources/Services/Telemetry/TelemetryClient.swift
@@ -158,6 +158,12 @@ final class TelemetryClient: Injectable {
         var payload: [String: Any] = [:]
 
         if let v = info["CFBundleShortVersionString"] as? String { payload["appVersion"] = v }
+        // Display name from Info.plist ($(APP_DISPLAY_NAME) in Config.xcconfig,
+        // default "Trio"). Forks typically rebrand by changing that one line,
+        // so this lets dashboards distinguish upstream Trio from forks.
+        let displayName = (info["CFBundleDisplayName"] as? String)
+            ?? (info["CFBundleName"] as? String)
+        payload["appDisplayName"] = displayName ?? "unknown"
         // appDevVersion is Trio's 4-component dev counter (e.g. "0.7.0.14") —
         // the most precise build identifier we have. Always emit, even when
         // the Info.plist key is missing, so dashboards can rely on the field.
--- a/Trio/Sources/Services/Telemetry/TelemetryClient.swift
+++ b/Trio/Sources/Services/Telemetry/TelemetryClient.swift
@@ -39,6 +39,14 @@ final class TelemetryClient: Injectable {
     private static let dailyInterval: TimeInterval = 24 * 60 * 60
     private static let maxPayloadBytes = 4096
 
+    private static let buildDateFormatter: DateFormatter = {
+        let f = DateFormatter()
+        f.dateFormat = "yyyy-MM-dd"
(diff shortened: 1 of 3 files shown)


bjornoleh:
Sounds good, except that adding appDisplayName can in some cases make the data collection not so anonymous any more, as some will have their own or their kid's name in the appDisplayName.

kristinkand:
> Sounds good, except that adding appDisplayName can in some cases make the data collection not so anonymous any more, as some will have their own or their kid's name in the appDisplayName. 

Very valid point! The same goes for the branch name which I noticed when I checked the sharing details.

Why would you need to collect "app DisplayName" and branch name in the first place?

Trio/Sources/Services/Telemetry/TelemetryClient.swift
```diff
@@ -150,17 +158,23 @@ final class TelemetryClient: Injectable {
         var payload: [String: Any] = [:]
 
         if let v = info["CFBundleShortVersionString"] as? String { payload["appVersion"] = v }
+        // Display name from Info.plist ($(APP_DISPLAY_NAME) in Config.xcconfig,
+        // default "Trio"). Forks typically rebrand by changing that one line,
+        // so this lets dashboards distinguish upstream Trio from forks.
+        let displayName = (info["CFBundleDisplayName"] as? String)
```
dnzxy: Please change this to a hard coded `Trio` and potentially update the variable to something like `appSource` or `source`. 

Please update the privacy policy in-app and what we capture and add this item as a captured metric.

Trio/Sources/Services/Telemetry/TelemetryClient.swift
```diff
@@ -173,6 +187,8 @@ final class TelemetryClient: Injectable {
         payload["device"] = Self.hardwareIdentifier()
         payload["platform"] = Self.detectPlatform()
         payload["osVersion"] = UIDevice.current.systemVersion
+        payload["locale"] = Locale.current.identifier
```
dnzxy, line 189: Please add these two metrics to privacy policy.

dnzxy (changes requested):

dnzxy:
@kristinkand 
> Why would you need to collect "app DisplayName" and branch name in the first place?

We have other Trio-derivate systems that will also be part of our telemetry infrastructure but need to identified as such. `appDisplayName` is just the "wrong" thing to track. It's more-so a telemetry-source identifier.

Tracking branch names is a common metric and does not realistically fall under PID. It helps us have insights into the user share between `main` and `dev` for the majority of* telemetry data, for example.

Trio/Sources/Services/Telemetry/TelemetryClient.swift
jwoglom: Done. Replaced with a 'Trio' string const and field 'appName'

Trio/Sources/Services/Telemetry/TelemetryClient.swift
jwoglom: Done

dnzxy (approved):

PR #1180, opened 2026-05-29
Add temp target display to live activity lock screen widget

t1dude:
When a temp target is active, an orange badge showing the temp target name appears in the top right corner of the widget. Both badges stack vertically if an override and a temp target are active at the same time. I have tried to mirror the existing Override code pattern.

PS: screen shots contain an issue that has been fixed already, where the TT badge does not make the graph shift to the left.

| Screenshots | Screenshots |
|--------|--------|
|  |  |
|  |  |


dnzxy:
I have been meaning to add this for so long, gawd, thank you! 

Two thoughts: 
- Badge color should be green, same green we use for all TT related UI elements throughout Trio.
- Have you tried to horizontally stack the badges? Vertically stacking them hides most recent glucose, which seems counter intuitive to a quick-glance element.  

Finally, someone adding this 🫶😂

dnzxy:
Oh, another thought before I forget it — have you checked how this looks or works for the Apple watch smart stack?

t1dude:
Thanks @dnzxy!

I will change the color to green. The graph, including last reading, moves to the left when there’s an active override or TT. This is how it behaves today with overrides. The screen shot was taken before that was added for temp targets. I’ll make a new screen shot after the color is changed.

t1dude:
IMG_0103
IMG_0101

t1dude:
IMG_0102

t1dude:
Regarding the watch LA, I haven’t touched that. It doesn’t show overrides currently, so I haven’t added TT. I could have a look at the watch LA in a separate PR.

t1dude:
Would you still like to see a version with horizontal stacking or do we leave it as is?

t1dude:
This is an alternative view.

dnzxy:
Okay, a couple thoughts:
- How about keeping them stacked vertically
- Potentially move them to the left side (6 hrs ago) of the chart
   - Potentially introduce the forecast to the Live Activity, but maybe a follow-up PR
- Replace the label (`Override`, `TempTarget`) with the SF Symbols that we use for Overrides and Temp Targets in the app and watch app, and color the badge accordingly (like you already do)
  - Override: `clock.arrow.2.circlepath`
  - Temp Target: `target`

If you want to keep it as-is I _think_ the horizontal stack is the preferable one, although you have text zoom enabled which I guess is why the TempTarget label and some of the LA widgets are showing text clipping.

t1dude:
If we replace the labels with the SF symbols, we won’t be able to see which override or temp target is active. It would be cleaner, but we would miss out on that information. However, that would make it possible to add to the watch LA as well.

I’d like to hear your thoughts on that. 

Regarding size and overlap, I am using enlarged text in iOS. I’ll make sure to add screenshots without that. 

Forecast lines: I think most people use delta for the glance view, but I could ask around for some input?

dnzxy:
> If we replace the labels with the SF symbols, we won’t be able to see which override or temp target is active. It would be cleaner, but we would miss out on that information. However, that would make it possible to add to the watch LA as well.

Oh, true. See, I don't ever use overrides, so I missed that. Good point! I think this will be different for caregivers/givees and PwD, where the latter know what is running anyways, cause they activated it. 

I guess then potentially keeping it horizontally stacked and potentially moving it to the left may work better. If we move it to the left, we don't need to add "spacing" when the OR / TT is active anymore, though.

> Forecast lines: I think most people use delta for the glance view, but I could ask around for some input?

Sounds good. It would just be an addition to the chart. Follow-up PR kinda stuff, anyways, not relevant here.

kristinkand:
It looks great! I like it on the right side too, but see your point @dnzxy 
Yes, the symbols are good however if you have several ORs/TTs you won't know which one runs.
Great work!

t1dude:
Currently there is no limitation on the length of the override/temp target name. 

Skjermbilde 2026–05–30 kl  20 20 21

Skjermbilde 2026–05–30 kl  20 23 13

Should I move them to the left and remove the feature that shortens the graph?

For the record, this is what it looks like with what I guess are more normal names for adjustments:

Skjermbilde 2026–05–30 kl  20 26 01

t1dude:
Moved the badges to the top left. Removed the code that cropped the graph. (Note: it looks a bit off in the screenshot, that's because I've had a site leak and CGM failure and was without readings for a while. Yes, screenshots are from my actual phone :) )

Top Right Full Graph

Top Right Full Graph Long Adjustment Names

dnzxy:
I think this looks great! 
Let me do a code review later today or tomorrow morning and then this is good to go! 

Thanks for adding this. It‘s really been bugging me for ages. Fist bump! 👊

munchyk:
In general, I like the idea of having an emoji input as a required field within the override/TT creation flow. 

It would also work well for live activities as the emoji could be placed on a circular pink/purple background for overrides or green for TT, eliminating the need for a line of text.

dnzxy:
@munchyk 
> In general, I like the idea of having an emoji input as a required field within the override/TT creation flow.

We contemplated doing that, but unfortunately there isn't a clean way for an emoji-only keyboard and we did not feel like adding a hacky workaround with a custom keyboard being fed emojis is a good way to get this working, plus it's a breaking change for every single Trio user and every single one of their presets.

munchyk:
I hear that. 

Couldn’t you create a field validation that only accepts emoji inputs? If the user tries entering text it could present an ‘error’ message saying “Please use the emoji keyboard to select an emoji.”

All current overrides could be automatically updated to use a default emoji to circumvent the “breaking change” aspect.

Would be helpful for watch too.

just a thought ☺️

munchyk:
Though pretty ridiculous that Apple doesn’t provide a way to present emoji keyboard by default!

dnzxy:
Let me steam on that a bit, maybe doable.
I remember having implemented something like that 75% through 2 years ago and then we binned it, but… it's been 2 years ago 😂

munchyk:
Lol. I don’t mind being the first upvote if that’s worth anything 😂

kristinkand:
> In general, I like the idea of having an emoji input as a required field within the override/TT creation flow. 
> 
> 
> 
It would be confusing because one might not remember which emoji was picked for which OR/TT. Additionally some emojis may not describe fully in which situation you want to use this specific OR/TT.

A limit to the amount of letters for the name of an OR/TT would make more sense.

munchyk:
IMO a character limit makes less sense because words differ in length across all languages.

In which cases would an emoji not accurately describe a preset? There are emojis for all sports, workouts, illness, sleep, hormones, etc.. 

Also, not sure what else a 🏃‍♀️emoji could be confused for - especially if you assigned it!

kristinkand:
> In which cases would an emoji not accurately describe a preset? There are emojis for all sports, workouts, illness, sleep, hormones, etc.. 
> 
..high night, low treatment vs low night, weekends, before breakfast, false high, giving a slight bump
> 
> Also, not sure what else a 🏃‍♀️emoji could be confused for - especially if you assigned it!

There is also LF which is used by caregivers (remote commands) and Trio is not in all cases used by just one singular person.

munchyk:
Using emojis and having text are not mutually exclusive.

Emojis allows for broader implementation such as in the case of live activities or watch.

A caregiver would still see the preset name in Loop Follow.

I think you may be overusing presets but here are some examples I’d use based on what I think you are describing (though ultimately people would need to pick from the hundreds of options based on what is meaningful to them):

Boost (“slight bump”) - 🚀

Before breakfast - ☕️

High Night - 🌕 or 🌝

Low Night - 🌘 or🌜

Low Treatment - 🧃 or 🍭 or 🍬

Weekends - 🍾 or 🛀 or 🔚 or ⏭️ (not sure what your use case/intention is)

kristinkand:
> I think you may be overusing presets but here are some examples I’d use based on what I think you are describing (though ultimately people would need to pick from the hundreds of options based on what is meaningful to them):
> 
I would like to remind you respectfully that our settings may not work for you and vice versa and please keep in mind YDMV.


@dnzxy and @t1dude I rebuilt this morning and we just tested the new feature. It only shows an OR, but not OR and TT at the same time.

t1dude:
Let’s leave the emoji discussion out of this PR, please. This PR is only addressing Temp Targets, that are currently not shown on the lock screen widget. I suggest that we move the emoji discussion to Discord,  a different PR or a GitHub Issue.

t1dude:
> @dnzxy and @t1dude I rebuilt this morning and we just tested the new feature. It only shows an OR, but not OR and TT at the same time.

Weird! Let’s look into that. Did it show the OR on the left or right side?

kristinkand:
> Weird! Let’s look into that. Did it show the OR on the left or right side?

It works! Aligned on the left side and both show. I had just "updated" anticipating it had been merged already. After building PR1180 it's all there. Pumpkin likes it and thanks!😊

LiveActivity/Views/LiveActivityChartView.swift
```diff
@@ -33,13 +33,13 @@ struct LiveActivityChartView: View {
...
         let isOverrideActive = additionalState.isOverrideActive == true
+        let isTempTargetActive = additionalState.isTempTargetActive == true
 
         let calendar = Calendar.current
         let now = Date()
 
         let startDate = calendar.date(byAdding: .hour, value: isWatchOS ? -3 : -6, to: now) ?? now
-        let endDate = isOverrideActive ? (calendar.date(byAdding: .hour, value: 2, to: now) ?? now) :
-            (calendar.date(byAdding: .minute, value: isWatchOS ? 5 : 0, to: now) ?? now)
+        let endDate = calendar.date(byAdding: .minute, value: isWatchOS ? 5 : 0, to: now) ?? now
```
marv-out, line 42: I am personally not a big fan of removing those 2 hours for active overrides/ temp targets as I like seeing my changed target in those cases instantly. But if the majority thinks that we should handle it like that thats fine

Trio/Sources/Services/LiveActivity/Data/DataManager.swift
```diff
@@ -67,6 +67,34 @@ extension LiveActivityManager {
         }
     }
 
+    func fetchAndMapTempTarget() async throws -> TempTargetData? {
```
marv-out, line 71: Thats more or less the exact same code as in `fetchAndMapOverride()`. No blocker, but we could make it generic which is more elegant. Something like:

```
  private func fetchAndMapLatest<Entity: NSManagedObject, Output>(
      ofType type: Entity.Type,
      map: @escaping ([String: Any]) -> Output
  ) async throws -> Output? { ... }
```

marv-out (commented):
Overall looking good. Made 2 small comments, see above. 

Another thing (we could maybe adress here or in a separate PR) I've noticed is when we add an indefinite `Override` with a different target its target bar never gets shown in the LA due to the way we're drawing the end date. For the indefinite case `duration` is 0 and then end = start + 0 = start

Trio/Sources/Services/LiveActivity/Data/DataManager.swift
t1dude: I have refactored this to make it cleaner. Could you have a look @marv-out ?

LiveActivity/Views/LiveActivityChartView.swift
t1dude: I'm not sure I understand what you are missing out on. The green target line is still moving up/down with the target from the OR/TT. How does an opening in the last part of the graph make this more visible?

dnzxy: Not sure either. I think what Marvin wants to point to is that say you have a target at 100 mg/dL, then the override / TT changes it to 80 mg/dL, then this "bump" or "push" should make (?) it more visible, but currently we don't plot a bar, but we just a single green line.

dnzxy (approved):

dnzxy:
@marv-out @t1dude I have added a new issue for the finding Marvin had during code review, so we can address it in a follow-up PR. See https://github.com/nightscout/Trio/issues/1191.

t1dude:
> @marv-out @t1dude I have added a new issue for the finding Marvin had during code review, so we can address it in a follow-up PR. See #1191.

Thanks @dnzxy ! I'll have a look at that one as well.

LiveActivity/Views/LiveActivityChartView.swift
marv-out: No when you remove the shift of the graph you won't see the rule mark we are drawing when you enact the temp target or override 

image

PR #1184, opened 2026-05-30
Fix COB string, localize live activity widget item labels

dnzxy:
- Correct label for Carbs on Board (parenthesis used to say "IOB")
- Make all Live Activity widget item labels localizable

Thanks for reporting @AndreasStokholm 🙏


AndreasStokholm (approved):
👏

t1dude (approved):
Yay!

Issue #1183, opened 2026-05-30
Tidepool authentication status misleading


hugomrosa:
When a user changes their account's password on Tidepool, Trio's Tidepool screen still shows a successful connection, although in the background the module is failing with an HTTP 400 status code error.
I would be expecting, in such scenarios, for Trio to show an error message on the screen "authentication error", or Red Cross to alert the user something's wrong.

## To Reproduce
1. Change Tidepool account password
2. Go to Trio > Services > Tidepool
3. Screen wil show a "Connected to tidepool" message with a globe and a green checkbox, as per the screenshot.

## Expected behavior![image](https://github.com/user-attachments/assets/09f8726f-a9c7-485a-a65a-94daa2b50f0f)
I would expect this screen to show a message "authentication error" or "not connected to Tidepool" and a globe with a red cross.

## Screenshots
*If applicable, add screenshots to help explain your problem.*

### Smartphone:
* Hardware: iphone 17
* OS Version: 26.5

### Pump:
* Manufacturer: Insulet
* Model: Omnipod Dash

### CGM:
* Device: Dexcom G6
* Manager app: Dexcom App

### Trio Version:
* Version Number: v0.8.0
* Repo: *nightscout/trio*
* Git Reference: *[e.g. commit hash]*

## Technical Details
Error message seen in the logs
2026-05-28T00:01:31+1000 [Nightscout] TidepoolManager.swift - uploadGlucose(_:) - 646 - DEV: Error synchronizing glucose data: responseMalformedJSON(<NSHTTPURLResponse: 0x130654080> { URL: https://auth.tidepool.org/realms/tidepool/protocol/openid-connect/token } { Status Code: 400, Headers {...}}, DecodingError.keyNotFound: Key 'access_token' not found in keyed decoding container. Debug description: No value associated with key CodingKeys(stringValue: "access_token", intValue: nil) ("access_token").)

PR #1186, opened 2026-05-30
Fix Tidepool auth state in service view

dnzxy:
- New TidepoolHealth enum with healthPublisher on TidepoolManager
- All 7 upload-completions in TidepoolManager now record success/failure
- TidepoolStartView subscribes via .onReceive and switches the connection indicator

Addresses #1183.

### DIFF
--- a/Trio/Sources/Services/Network/TidepoolManager.swift
+++ b/Trio/Sources/Services/Network/TidepoolManager.swift
@@ -9,6 +9,26 @@ import Swinject
 import TidepoolServiceKit
 import UIKit
 
+/// Observed state of the Tidepool integration, derived from upload outcomes to Tidepool.
+///
+/// `getTidepoolServiceUI() != nil` only tells the UI that a service object
+/// is configured locally — it doesn't say whether the credentials are still
+/// accepted by Tidepool. This enum is the truth source for the "connected"
+/// indicator in `TidepoolStartView`. It's updated on every upload success /
+/// failure callback (carbs, insulin, glucose, settings, deletes).
+enum TidepoolHealth: Equatable {
+    /// No upload attempt yet this process — show optimistic UI.
+    case unknown
+    /// Most recent upload returned 2xx at the given time.
+    case healthy(at: Date)
+    /// Refresh-token grant rejected or API token explicitly refused (401/403).
+    /// User needs to re-authenticate via the existing Tidepool settings sheet.
+    case authFailed(at: Date)
+    /// Any other failure (5xx, network blip, decode error, other 4xx).
+    /// Self-clears on the next successful upload.
+    case transient(at: Date)
+}
+
 protocol TidepoolManager {
     func addTidepoolService(service: Service)
     func getTidepoolServiceUI() -> ServiceUI?
@@ -20,6 +40,9 @@ protocol TidepoolManager {
     func uploadGlucose() async
     func uploadSettings() async
     func forceTidepoolDataUpload()
+    /// Live updates whenever an upload returns; backed by a `CurrentValueSubject`
+    /// so subscribers receive the current value on subscribe.
+    var healthPublisher: AnyPublisher<TidepoolHealth, Never> { get }
 }
 
 final class BaseTidepoolManager: TidepoolManager, Injectable {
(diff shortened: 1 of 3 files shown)


dnzxy:
Is there a good way for you to test this @loudestnoise? Can you provoke authentication failures in a good, easy way? :)

loudestnoise:
## Test
Testing to see if when Tidepool consent is revoked by Tidepool if error state is presented in Tidepool in Settings - Services

Built Trio using `fix/tidepool-auth-state` commit [64c9598](https://github.com/nightscout/Trio/pull/1186/commits/64c95987280b5b586c0d6bf4b4b30d5803656bb7)
### Test Device
- iPhone SE 2nd gen, iOS 18.7.8
- CGM & Pump Simulator in Trio
### Test Execution
- Logged into test Tidepool account on production
- Revoked consent in Tidepool admin interface after confirming data upload in Tidepool Data Platform
- Confirmed in Settings - Services - Tidepool `Tidepool Auth Error - Tap to re-login` error state is presented
image

- Tapping error state allows user to delete service
- New `Tidepool Sync Error` state, which when tapped allows user to authenticate into Tidepool again

Error_state_to_new_auth

### Test Results
- ✅ Successful error state presented

Trio/Sources/Services/Network/TidepoolManager.swift
```diff
@@ -164,6 +188,60 @@ final class BaseTidepoolManager: TidepoolManager, Injectable {
...
+    /// transient network/server hiccup. Called from every `.failure` branch
+    /// in the upload completion handlers.
+    fileprivate func noteUploadFailure(_ error: Error) {
+        notifyHealth(classify(error))
+    }
+
+    /// Fans a health change out to all registered `TidepoolHealthObserver`s
+    /// on the main thread. Upload completion handlers can fire from arbitrary
+    /// background queues, and observers update SwiftUI `@Published` state.
+    private func notifyHealth(_ health: TidepoolHealth) {
```
marv-out, line 211: One thing to discuss: is it necessary to inform the broadcaster on every single upload of glucose, carbs, etc although the UI state hasn't changed at all? we could simply put a guard in here that guards against the same case and only informs the broadcaster if the `TidepoolHealth` cases are different. Sth like: 
```

  private var lastBroadcastHealth: TidepoolHealth = .unknown

  private func notifyHealth(_ health: TidepoolHealth) {
// only notify if its not the same TidepoolHealth case
      guard !isSameCase(lastBroadcastHealth, health) else { return }
      lastBroadcastHealth = health
      DispatchQueue.main.async {
          self.broadcaster.notify(TidepoolHealthObserver.self, on: .main) {
              $0.tidepoolHealthDidChange(health)
          }
      }
  }

// if the same TidepoolHealth case, return true
  private func isSameCase(_ a: TidepoolHealth, _ b: TidepoolHealth) -> Bool {
      switch (a, b) {
      case (.unknown, .unknown), (.healthy, .healthy), (.authFailed, .authFailed), (.transient, .transient):
          return true
      default:
          return false
      }
  }
```

marv-out (commented):
LGTM from what I've seen during code review

Trio/Sources/Services/Network/TidepoolManager.swift
dnzxy: It's not strictly necessary, but during nominal use, glucose &/or insulin is pushed, on average, once per loop (I am ignoring user input) and so we check around every 5 min if the connection is still valid. 

We can of course drop this. Should we, i.e. is this a blocker for you, @marv-out ?

marv-out: Hmm good question. I wanted to discuss that and I tend to say yes to your question 😄. I think I would rather implement sth like I mentioned above. It still checks, but only informs the broadcaster if there is a different TidepoolHealth case

dnzxy:
Putting this into draft mode until #1203 land. TidepoolService actually emits an authorization alert, and so this feature may likely become obsolete once we land proper LoopKit Alert framework based alerting. 

For later reference: https://github.com/LoopKit/TidepoolService/blob/dev/TidepoolServiceKit/TidepoolService.swift#L147-L153

Issue #1181, opened 2026-05-29
Statistics ! Meals chart shows zero for Carbs, Fat and Protein even though some has been entered


kfahling:
## Describe the bug
When going into the Stats ! Meals selection 0.0 g is always shown for Carbs, Fat and Protein even though values have been entered.  However, by slightly scrolling to the previous day the correct values are filled in.  Then if I scroll all the way to the current day again they are zeroed out invalidly.  Note that I also scrolled to previous day so that only  that day is shown and carbs did total correctly.  So, it's something about the fact that it's the current day or that the current day doesn't have all 24 hours included as some haven't happened yet.

## Attach a Log
LMK if this is needed but I suspect not as it is very easily reproducible.  I did reproduce on latest 0.8.1 level of code. 

## To Reproduce
1. Ensure Carbs are entered for the day.
2. Go to Stats
3. Click on Meals
4. See that Daily Carbs value is zero. Fast and Protein also show zero even if they were entered.

## Expected behavior
Above steps would show correct current day total of Carbs, Fat & Protein

## Screenshots
First screenshot shows incorrect value of zero for carb total.  Second screenshot is when I scrolled slightly earlier thus including previous day which didn't have carbs but did cause total to be correct.  I didn't include fat and protein as I don't typically record them but did confirm previously that those are also summed incorrectly when the current day is shown.

### Smartphone:
* Hardware: iPhone 14 Pro
* OS Version: iOS 26.5

### Pump:
* Manufacturer: Insulet
* Model: Omnipod Dash

### CGM:
* Device: Dexcom G7
* Manager app: Dexcom App

### Trio Version:
* Version Number: 0.8.1
* Repo: nightscout/trio
* Git Reference: 35499ad

## Technical Details
Nothing I can think of other than above.

## Additional context
Gotta add some carbs, protein or fat for the day

<img width="590" height="1278" alt="Image" src="https://github.com/user-attachments/assets/c2631853-98b8-47ab-b032-a25041298930" />

<img width="590" height="1278" alt="Image" src="https://github.com/user-attachments/assets/8f813fc3-0482-4938-9707-94e17701c22e" />

PR #1188, opened 2026-05-31
Fix Stats Meals showing zero macros for current day

dnzxy:
## Summary 
* `calculateAveragesForDateRange` filtered cached daily totals with a strict `date >= startDate` against day-keys that are startOfDay. The .day-interval
* initial scroll position is startOfDay(today) + 1s, which leaves today's `today 00:00:00` bucket just outside the range, zeroing today's carbs/fat/protein in the Meals chart (reported as #1181).
* Scrolling slightly back into yesterday made today appear again.

## Fix
* Compare against the day window (overlap semantics) instead of the day's start instant: a cached day belongs in the range if any moment of it overlaps.
* Backward-compatible for `.week`/`.month`/`.total` because those intervals already align `startDate` to midnight.

## Additional Change
* Simplify calculateAndCacheDailyAverages (no behavior change) 
* `dailyMealStats` is already one entry per day from `fetchMealStats`, so the previous `Dictionary(grouping:)` and per-group reduce + divide-by-count was a passthrough (count==1, total/1==total).
* Replace with a direct `Dictionary(_, uniquingKeysWith:)` re-key by `startOfDay`, keeping the defensive sum-merge in case a future caller constructs MealStats with mid-day timestamps.

---

Addresses #1181.

### DIFF
--- a/Trio/Sources/Modules/Stat/StatStateModel+Setup/MealStatsSetup.swift
+++ b/Trio/Sources/Modules/Stat/StatStateModel+Setup/MealStatsSetup.swift
@@ -107,37 +107,30 @@ extension Stat.StateModel {
         }
     }
 
-    /// Calculates and caches the daily averages of macronutrients
+    /// Caches per-day macro totals keyed by `startOfDay`, for fast range
+    /// lookups in `calculateAveragesForDateRange`.
     ///
-    /// This function:
-    /// 1. Groups meal statistics by day
-    /// 2. Calculates average carbs, fat and protein for each day
-    /// 3. Caches the results for later use
+    /// `dailyMealStats` already has one entry per day (built that way by
+    /// `fetchMealStats`'s `dailyGrouped`), so re-grouping and dividing by
+    /// `count == 1` was a no-op. The `uniquingKeysWith:` merge is
+    /// defensive against any future call site that constructs `MealStats`
+    /// with mid-day timestamps for the same day.
     ///
-    /// This only needs to be called once during subscribe.
+    /// Only needs to be called once during subscribe.
     private func calculateAndCacheDailyAverages() async {
         let calendar = Calendar.current
 
-        // Calculate averages in context
         let dailyAverages = await mealTaskContext.perform { [dailyMealStats] in
-            // Group by days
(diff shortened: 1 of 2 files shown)


Trio/Sources/Modules/Stat/StatStateModel+Setup/MealStatsSetup.swift
```diff
@@ -107,37 +107,30 @@ extension Stat.StateModel {
...
+    /// defensive against any future call site that constructs `MealStats`
+    /// with mid-day timestamps for the same day.
     ///
-    /// This only needs to be called once during subscribe.
+    /// Only needs to be called once during subscribe.
     private func calculateAndCacheDailyAverages() async {
         let calendar = Calendar.current
 
-        // Calculate averages in context
         let dailyAverages = await mealTaskContext.perform { [dailyMealStats] in
```
marv-out: actually we don't need the `mealTaskContext.perform` closure here anymore as we are not accessing Core Data properties

Trio/Sources/Modules/Stat/StatStateModel+Setup/MealStatsSetup.swift
```diff
@@ -107,37 +107,30 @@ extension Stat.StateModel {
...
-    /// 3. Caches the results for later use
+    /// `dailyMealStats` already has one entry per day (built that way by
+    /// `fetchMealStats`'s `dailyGrouped`), so re-grouping and dividing by
+    /// `count == 1` was a no-op. The `uniquingKeysWith:` merge is
+    /// defensive against any future call site that constructs `MealStats`
+    /// with mid-day timestamps for the same day.
     ///
-    /// This only needs to be called once during subscribe.
+    /// Only needs to be called once during subscribe.
     private func calculateAndCacheDailyAverages() async {
```
marv-out: After the refactoring we're not calculating the **averages** in this function anymore but the **totals** instead. You are describing that correctly in your comments. I would suggest changing the variable names then for clarity

marv-out (changes requested):
Tested in a simulator, works! 2 small comments though

marv-out (approved):

PR #1189, opened 2026-05-31
Change profile target dynISF condition with HighTTRaisesSensitivity

dnzxy:
~~- Before this change, condition checked whether highTTRaisesSens was active AND user's profile target >= 118~~
~~- We received an issue report by a parent whose child runs with profile target set at >118, thus dynISF NEVER activated~~
~~- Change now moves this check condition to a saner target of halfBasalExerciseTarget, to ONLY disable dynISF if the user ACTUALLY sets a TT at a target value >= HBT (which by default is 160)~~

Removes a legacy (pre Trio) dynISF condition that would disable dynISF.

### DIFF
--- a/Trio/Sources/APS/OpenAPSSwift/DetermineBasal/DynamicISF.swift
+++ b/Trio/Sources/APS/OpenAPSSwift/DetermineBasal/DynamicISF.swift
@@ -159,7 +159,7 @@ extension Preferences {
 
         // checks for 'exercise mode' like conditions
         if profile.highTemptargetRaisesSensitivity,
-           let profileTarget = profile.profileTarget(trioCustomOrefVariables: trioCustomOrefVariables), profileTarget >= 118
+           let profileTarget = profile.profileTarget(trioCustomOrefVariables: trioCustomOrefVariables), profileTarget >= profile.halfBasalExerciseTarget
         {
             return .off
         }
@@ -167,3 +167,4 @@ extension Preferences {
         return sigmoid ? .sigmoid : .logrithmic
     }
 }
+g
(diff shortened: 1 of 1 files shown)


tmhastings:
This child had a profile target of 125 mg/dL because of honeymooning and being a toddler. Sigmoid was enabled in settings and showing in Trio Settings export csv, yet autosens ratio was running at 0.5 even though glucose was well above target (200+). OpenAPS pill in Nightscout, looping popup in Trio, and logs did not show Sigmoid as enabled. Ruled out all other known and expected causes:
✅ Dynamic ISF Sigmoid enabled
✅ Looping success rate 97.8% over 7 days
✅ No Temp Target >100 enabled

When `High Temp Target Raises Sensitivity` was DISABLED by the parent, Sigmoid appeared correctly in the looping popup. Setting a profile target to 115 mg/dL, with `High Temp Target Raises Sensitivity` ENABLED, also resulted in a successful Sigmoid looping cycle.

To test this another way, I set a target only override (not a Temp Target) of 120 mg/dL for myself. Sigmoid was disabled in the following loop cycle. It does not seem to activate the sensitivity override, but does disable dynamic ISF.

dnzxy:
@kingst have a look please :)

Trio/Sources/APS/OpenAPSSwift/DetermineBasal/DynamicISF.swift
```diff
@@ -157,13 +157,7 @@ extension Preferences {
...
-        if profile.highTemptargetRaisesSensitivity,
-           let profileTarget = profile.profileTarget(trioCustomOrefVariables: trioCustomOrefVariables), profileTarget >= 118
-        {
-            return .off
-        }
-
         return sigmoid ? .sigmoid : .logrithmic
     }
 }
+g
```
marv-out: I guess that made it in by mistake :)

marv-out (changes requested):
One comment, otherwise LGTM. @kingst may have a quick look as well

marv-out:
LGTM. @kingst ?

kingst (approved):
LGTM!

Issue #1162, opened 2026-05-20
Trio Watch complication becomes non responsive


dirkgastaldo:
## Describe the bug
*A clear and concise description of what the bug is. Describe what you see versus what you expect to see.*
Install new version of Trio including Watch app.
Complication works, that is it opens Trio app on Watch.
Within hours tap Watch complication again and it is nonresponsive that is nothing happens when you touch complication. App does not open on Watch. EDIT: Turns out it's not just the Complication, it's the Watch app. I tried to open the Trio app on my phone yesterday from the Watch desktop/home, and it did not open when I touched the icon. I installed the new version that hit TestFlight today May 24 it I can open the app from the desktop today. I'll try again periodically and update this status.

## Attach a Log
*Tap the Trio settings icon at the bottom of the screen, then tap 'Share logs' on the bottom of the list and attach it to this ticket.*

## To Reproduce
*Steps to reproduce the behavior:*
1. *Go to '...'*
2. *Click on '....'*
3. *Scroll down to '....'*
4. *See error*

## Expected behavior
*A clear and concise description of what you expected to happen.*

## Screenshots
*If applicable, add screenshots to help explain your problem.*

### Smartphone:
* Hardware: *[e.g. iPhone 15 Pro]*
* OS Version: *[e.g. iOS 17.5]*
17 Pro Max iOS 26.5

### Pump:
* Manufacturer: *[e.g. Insulet]*
* Model: *[e.g. Omnipod Dash or Eros]*
Omnipod Dash

### CGM:
* Device: *[e.g. Dexcom G7]*
* Manager app: *[e.g. Dexcom App or xDrip4iOS]*
Dexcom G7

### Trio Version:
* Version Number: *[e.g. 1.9.2]*
* Repo: *nightscout/trio*
* Git Reference: *[e.g. commit hash]*
Main 0.7

## Technical Details
*If applicable, provide any technical details that might help in diagnosing the problem. This could include logs, error messages, or relevant configuration details.*

## Additional context
*Add any other context about the problem here.*

PR #1190, opened 2026-06-01
Fix watch app back-sync after long staleness

dnzxy:
* Unify didReceiveMessage + didReceiveUserInfo through one handleIncomingWatchStatePayload
* Add guard on wall-clock staleness, then perform monotonicity dedupe
* Add state.date with Date() at send time, not build time
* Drop lastSent < state.date guard
* Watch-requested re-pushes no longer get silenced when no CGM tick has bumped the build-time date Potentially addresses #1162

### DIFF
--- a/Trio Watch App Extension/WatchState.swift
+++ b/Trio Watch App Extension/WatchState.swift
@@ -62,6 +62,11 @@ import WatchConnectivity
 
     var recommendedBolus: Decimal = 0
 
+    /// Snapshots older than this are dropped at the top of the WC delegate
+    /// methods. Single source of truth for both `didReceiveMessage` and
+    /// `didReceiveUserInfo`.
+    private static let maxAcceptableMessageAge: TimeInterval = 15 * 60
+
     // MARK: - Debouncing and batch processing helpers
 
     /// Temporary storage for new data arriving via WatchConnectivity.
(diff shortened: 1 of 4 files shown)


t1dude:
...testing

dnzxy:
Just to report from my self-testing – which shouldn't be the passing-the-barrier element for this PR, as I am the author:

I ran this last night on my own Apple Watch Series 8 @ watchOS 10, it seemed to update faster. 
The following recording was this morning with the same watch that wasn’t touched for ~11.5 hrs. Looks to be updating pretty much instantaneously. 

https://github.com/user-attachments/assets/4b411914-b141-444c-98df-56532694cb33

Curios to see what other testers report 👀

mountrcg:
I looked at the logs and Dan's observation: "buttons don't disable like in Trio does", even though Tai has no differences in that respect.

The busy/pending overlay (`AcknowledgementPendingView`) that should stay on screen with greyed-out buttons after a treatment is sent from the watch instead flashes red (*"Error! Invalid or incomplete data received from watch."*) for a split-second and bounces back to the main view — even when the action actually succeeded on the phone. To a user it reads as *"buttons don't disable like Trio does"*, because the disabled state is invisibly brief.

---

### Bug 1 — `watchLogs` payloads fall through to the failure ack

The log handler at [L558–L560](https://github.com/nightscout/Trio/blob/ebebe817e3f74aee3ed407884a3dda0ff9f0430c/Trio/Sources/Services/WatchManager/AppleWatchManager.swift#L558-L560) runs `SimpleLogReporter.appendToWatchLog(logs)` but **never returns**. The `Task { @MainActor }` at L561 then re-evaluates the *same* `message`; a logs-only dict matches none of `requestWatchUpdate` / `snooze` / `bolus` / `carbs` / `combo`, so it lands in the `else` at [L606–L614](https://github.com/nightscout/Trio/blob/ebebe817e3f74aee3ed407884a3dda0ff9f0430c/Trio/Sources/Services/WatchManager/AppleWatchManager.swift#L606-L614) and fires a failure ack back at the watch:

```swift
} else {
    debug(.watchManager, "📱 Invalid or incomplete data received from watch. Received:  \(message)")
    // Acknowledge failure
    self.sendAcknowledgment(
        toWatch: false,
        message: "Error! Invalid or incomplete data received from watch.",
        ackCode: .genericFailure
    )
}
```

**iOS `2026-06-01.log` — 18 `Invalid…` lines, 18/18 are `watchLogs` payloads (0 are real treatments):**

```
2026-06-01T07:27:07Z [WatchManager] AppleWatchManager.swift - session(_:didReceiveMessage:) - 607 - DEV: 📱 Invalid or incomplete data received from watch. Received:  ["watchLogs": [2026-06-01T09:27:07+0200] [WatchState.swift:251] session(_:didReceiveUserInfo:) → ⌚️ Invalid snapshot received]
```

**Watch `watch_2026-06-01.log` — 188 spurious failure acks bounced back, all carrying the exact else-block string (0 failure acks from any other path):**

```
[2026-06-01T09:27:07+0200] [WatchState.swift:351] processWatchMessage(_:) → ⌚️ Received acknowledgment: Error! Invalid or incomplete data received from watch., success: false
[2026-06-01T09:27:07+0200] [WatchState.swift:120] handleAcknowledgment(success:message:isFinal:) → ⌚️ Acknowledgment failed: Error! Invalid or incomplete data received from watch.
```

> The iOS log retained only 18 of these while the watch received 188 — the gap is iOS staging-log lossiness (rotated/capped at ~13 MB/day), not a separate cause. The watch recorded all 188; since no override/TT/bolus-rec was sent these days, all 188 are log-shipping messages hitting this `else`. (193 across both days: 188 on 06-01 + 5 on 06-02.)

The same logs-handler-without-return exists in `didReceiveUserInfo` at [L691–L693](https://github.com/nightscout/Trio/blob/ebebe817e3f74aee3ed407884a3dda0ff9f0430c/Trio/Sources/Services/WatchManager/AppleWatchManager.swift#L691-L693), but that path has no `else`, so it produces no spurious ack — only the foreground `sendMessage` path is affected.

**Fix:** `return` after `appendToWatchLog`. Safe — the watch always ships logs as a standalone message (18/18 observed dicts were single-key `watchLogs`), never bundled with a treatment.

---

### Bug 2 — override / temp-target / bolus-rec handlers run *after* the failure ack already fired

The handlers at [L616–L687](https://github.com/nightscout/Trio/blob/ebebe817e3f74aee3ed407884a3dda0ff9f0430c/Trio/Sources/Services/WatchManager/AppleWatchManager.swift#L616-L687) are standalone `if` blocks (**not** `else if`), placed **below** the `else` that just fired the failure ack:

```swift
if message[WatchMessageKeys.cancelOverride] as? Bool == true { ... }                  // L616
if let presetName = message[WatchMessageKeys.activateOverride]   as? String { ... }    // L621
if let presetName = message[WatchMessageKeys.activateTempTarget] as? String { ... }    // L626
if message[WatchMessageKeys.cancelTempTarget] as? Bool == true { ... }                // L631
if message[WatchMessageKeys.requestBolusRecommendation] as? Bool == true { ... }      // L636
```

Flow when the watch sends e.g. `activateOverride`:

1. The `if/else if` chain at L582–L605 doesn't recognize the key → `else` (L606) runs → **failure** ack sent.
2. The standalone `if` at L621 then matches → `handleActivateOverride` ([L938](https://github.com/nightscout/Trio/blob/ebebe817e3f74aee3ed407884a3dda0ff9f0430c/Trio/Sources/Services/WatchManager/AppleWatchManager.swift#L938)) processes the override correctly and sends a **success** ack at [L1019–L1027](https://github.com/nightscout/Trio/blob/ebebe817e3f74aee3ed407884a3dda0ff9f0430c/Trio/Sources/Services/WatchManager/AppleWatchManager.swift#L1019-L1027) (`toWatch: true, ackCode: .overrideStarted`).

**Net result:** two acks per override/TT/bolus-rec from the watch — **failure first** (flips `acknowledgementStatus = .failure`, pops the overlay), then success (too late, UI already gone).

**Not observed in these two days.** The logs neither confirm nor refute this bug: there is no `📱 Received … request from watch` marker (the signature of any watch-originated treatment) and no watch-side send marker anywhere in 06-01 or 06-02. We can't tell whether no on-watch treatment was attempted, or one was attempted that the frozen watch never managed to send. Either way the path was never exercised on tape.

The override activity that *does* appear came from **App Intents / Shortcuts** (`OverridePresetsIntentRequest.swift`), not the watch:

```
2026-06-02T07:33:21Z [Default] OverridePresetsIntentRequest.swift - enactOverride(_:) - 113 - DEV: Enacting override: Food  100/60 TH
2026-06-02T07:33:21Z [Default] OverridePresetsIntentRequest.swift - disableAllActiveOverrides(shouldStartBackgroundTask:) - 168 - DEV: Disabling all active overrides
2026-06-02T07:33:22Z [Nightscout] NightscoutManager.swift - uploadOverrides(_:) - 985 - DEV: Overrides uploaded
```

This bug is established from source structure alone. It would first surface in logs as an iOS-side pair like:

```
📱 Invalid or incomplete data received from watch. Received:  ["activateOverride": "<preset name>"]
📱 Received activate override request from watch for preset: <preset name>
```

followed on the watch side by a failure ack immediately succeeded by an `.overrideStarted` success ack for the same override.

**Fix:** chain the trailing handlers into the same `if/else if` ladder so the `else` (failure ack) only runs when *nothing* matched.

---

### Relationship to this PR

Separately, every background `transferUserInfo` push was being rejected watch-side as `⌚️ Invalid snapshot received` (316 hits on 06-01, 248 on 06-02), which is the stale-data/freeze that this PR fixes. That freeze is *why* the watch was reduced to shipping logs (feeding Bug 1) and never sent a treatment (leaving Bug 2 unobserved). Bug 1 and Bug 2 are independent of this here and they need the two fixes above.

dnzxy:
My observation was that the buttons don‘t _look like they disabled_ like Trio does. And everything that follows is hallucination which I‘m going to ignore if you don’t mind. 

The bug that was addressed here was the transferUserInfo bit and another one with nested Dict structure.

t1dude (approved):
**Test rig:**
Dev + PR 1190
iOS 26.5 / iPhone (17) Air
WatchOS 26.5 / Apple Watch Ultra 3

**Findings:**
During the day, setting the Trio watch app to stay open for 1 hour. Every time I raise my wrist, the app updates with new numbers after a few seconds. No crashes, no stale data issues, no need for force-closing the app.

Bolus from watch works well every time. No error messages, no need for force-closing the app.

After a night of inactivity, opening the watch app leads to automatic reload of data. No need for force-closing the app.

As far as I can see, this PR solves all issues I have had in the past with the watch app not updating and having to force-close and re-open to get updated data.

**Note:**
In my experience, any re-build / re-install of the watch app leads to a temporary performance boost. I do not yet know if performance will become worse over time. However, the immediate improvement from this PR is so obvious that I don't think it is necessary to "wait and see" before approving.

dnzxy:
I will move this to `dev` going by personal testing and the approval by @t1dude and positive test feedback on Discord.
Let's keep watching this one :)

PR #1192, opened 2026-06-05
Fix indefinite override not drawn in Live Activity chart

### LINKED ISSUE #1191: Indefinite Override Not Drawn in LiveActivity Chart
## Describe the bug
@marv-out observed that
> when we add an indefinite Override with a different target its target bar never gets shown in the LA due to the way we're drawing the end date. For the indefinite case duration is 0 and then end = start + 0 = start

## To Reproduce
1. Activate indefinite override
2. Lock phone / open dynamic island extended view
3. Notice how override bar is not drawn

## Expected behavior
Indefinite override should be drawn in LA chart.

t1dude:
When override duration is 0 (indefinite), end date equaled start date, giving the RuleMark zero width and preventing it from rendering. Treat duration == 0 as indefinite and extend the bar to now + 2 hours, matching the chart's x-axis domain.

Fixes #1191

### DIFF
--- a/LiveActivity/Views/LiveActivityChartView.swift
+++ b/LiveActivity/Views/LiveActivityChartView.swift
@@ -113,7 +113,9 @@ struct LiveActivityChartView: View {
         let duration = context.state.detailedViewState.overrideDuration
         let durationAsTimeInterval = TimeInterval((duration as NSDecimalNumber).doubleValue * 60) // return seconds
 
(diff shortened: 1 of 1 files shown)


dnzxy (approved):
LGTM.

Nit comment: add an inline comment what `duration == 0` actually means.

marv-out (approved):

PR #1194, opened 2026-06-06
Live Activity: Adding glucose forecast

t1dude:
New feature adding glucose forecast to the Live Activity lock screen widget. 

Quick walkthrough:

- The feature is controlled by a setting under Settings - Notifications - Live Activity
- Default setting is off
- Forecast display type (cone or lines) is controlled by the same setting as the Main chart

| Setting, default is off | Setting explanation | Example with prediction lines | Example with cone |
| LA_Pred_Setting | LA_Pred_Setting_Help | LA_Prediction_Lines_2 | LA_Cone_2 | 

Runs fine on my test device and on my live device, but needs more testers and a code review.

### DIFF
--- a/LiveActivity/LiveActivity.swift
+++ b/LiveActivity/LiveActivity.swift
@@ -128,7 +128,9 @@ private extension LiveActivityAttributes.ContentState {
             tempTargetTarget: 120,
             widgetItems: LiveActivityAttributes.LiveActivityItem.defaultItems,
             minForecast: [],
-            maxForecast: []
+            maxForecast: [],
+            forecastLines: [],
+            forecastDisplayType: "cone"
         )
 
     // 0 is the widest digit. Use this to get an upper bound on text width.
--- a/LiveActivity/LiveActivity.swift
+++ b/LiveActivity/LiveActivity.swift
@@ -126,7 +126,11 @@ private extension LiveActivityAttributes.ContentState {
             tempTargetDate: Date().addingTimeInterval(-1800),
             tempTargetDuration: 60,
             tempTargetTarget: 120,
-            widgetItems: LiveActivityAttributes.LiveActivityItem.defaultItems
+            widgetItems: LiveActivityAttributes.LiveActivityItem.defaultItems,
+            minForecast: [],
+            maxForecast: [],
+            forecastLines: [],
+            forecastDisplayType: "cone"
         )
 
     // 0 is the widest digit. Use this to get an upper bound on text width.
--- a/LiveActivity/LiveActivity.swift
+++ b/LiveActivity/LiveActivity.swift
@@ -128,9 +128,7 @@ private extension LiveActivityAttributes.ContentState {
             tempTargetTarget: 120,
             widgetItems: LiveActivityAttributes.LiveActivityItem.defaultItems,
             minForecast: [],
-            maxForecast: [],
-            forecastLines: [],
-            forecastDisplayType: "cone"
+            maxForecast: []
         )
 
     // 0 is the widest digit. Use this to get an upper bound on text width.
--- a/LiveActivity/Views/LiveActivityChartView.swift
+++ b/LiveActivity/Views/LiveActivityChartView.swift
@@ -38,7 +38,7 @@ struct LiveActivityChartView: View {
 
         let isOverrideActive = additionalState.isOverrideActive == true
         let isTempTargetActive = additionalState.isTempTargetActive == true
-        let hasForecast = !additionalState.minForecast.isEmpty
+        let hasForecast = !additionalState.minForecast.isEmpty || !additionalState.forecastLines.isEmpty
 
         let calendar = Calendar.current
         let now = Date()
@@ -47,7 +47,11 @@ struct LiveActivityChartView: View {
         let endDate: Date = {
             let baseEnd = calendar.date(byAdding: .minute, value: isWatchOS ? 5 : 0, to: now) ?? now
             guard hasForecast, let anchorDate = state.date else { return baseEnd }
-            let predictionEnd = anchorDate.addingTimeInterval(TimeInterval(additionalState.minForecast.count * 300))
+            let forecastCount = max(
+                additionalState.minForecast.count,
+                additionalState.forecastLines.max(by: { $0.values.count < $1.values.count })?.values.count ?? 0
+            )
+            let predictionEnd = anchorDate.addingTimeInterval(TimeInterval(forecastCount * 300))
             return max(baseEnd, predictionEnd)
         }()
 
@@ -94,7 +98,11 @@ struct LiveActivityChartView: View {
             }
 
             if hasForecast, let anchorDate = state.date {
-                drawForecastCone(anchorDate: anchorDate, isMgdL: isMgdL, maxValue: maxValue)
+                if additionalState.forecastDisplayType == "lines" {
+                    drawForecastLines(anchorDate: anchorDate, isMgdL: isMgdL)
+                } else {
+                    drawForecastCone(anchorDate: anchorDate, isMgdL: isMgdL, maxValue: maxValue)
+                }
             }
 
             drawChart(yAxisRuleMarkMin: yAxisRuleMarkMin, yAxisRuleMarkMax: yAxisRuleMarkMax)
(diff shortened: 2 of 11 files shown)


kristinkand:
I built it and changed between cone and lines and back to the cone. Both work well. 
Important to keep in mind is that the size for the LA widget is very limited and this influences also, how well you can e.g. differentiate the lines vs the cone.
The way of prediction shown in the LA widget follows the setting chosen for the main screen.

A possible idea could be to be able to select different styles for the widget and the main screen.

Great work @t1dude 🚀

t1dude:
> I built it and changed between cone and lines and back to the cone. Both work well. Important to keep in mind is that the size for the LA widget is very limited and this influences also, how well you can e.g. differentiate the lines vs the cone. The way of prediction shown in the LA widget follows the setting chosen for the main screen.

Thanks for testing! :)

> 
> A possible idea could be to be able to select different styles for the widget and the main screen.

I considered that, but I know that there is a consensus of trying to keep the amount of settings and options to a minimum. Among other things, this is based on how AndroidAPS ended up with too many toggles, making onboarding new users more difficult. Milos is working on removing and cleaning up for AAPS v4, and he's getting lots of pushback. Once a setting or feature is out there, it's hard to take it away.

Another approach could be to visualize the forecast in a totally different way, easier to see in the LA widget. However, that would break with how it is done in the main app.

> 
> Great work @t1dude 🚀

❤️

LiveActivity/Views/LiveActivityChartView.swift
```diff
@@ -83,6 +97,14 @@ struct LiveActivityChartView: View {
                 drawActiveTempTarget()
             }
 
+            if hasForecast, let anchorDate = state.date {
+                if additionalState.forecastDisplayType == "lines" {
```
marv-out: if we are touching this anyway I would suggest adding an enum to make the comparison more robust. Sth. like: 

```
 private enum ForecastDisplayTypeKey {
      static let lines = "lines"
      static let cone = "cone"
  }
```

LiveActivity/LiveActivity.swift
```diff
@@ -126,7 +126,11 @@ private extension LiveActivityAttributes.ContentState {
             tempTargetDate: Date().addingTimeInterval(-1800),
             tempTargetDuration: 60,
             tempTargetTarget: 120,
-            widgetItems: LiveActivityAttributes.LiveActivityItem.defaultItems
+            widgetItems: LiveActivityAttributes.LiveActivityItem.defaultItems,
```
marv-out, line 129: Just a side note: we have to be a little bit cautious here to not make ContentState any bigger as there is a size limit in ActivityKit and the LA would probably crash if we are exceeding that

LiveActivity/Views/LiveActivityChartView.swift
```diff
@@ -147,6 +169,72 @@ struct LiveActivityChartView: View {
...
+        return ForEach(0 ..< points.count, id: \.self) { i in
+            let point = points[i]
+            LineMark(
+                x: .value("Time", point.date),
+                y: .value("Value", point.value),
+                series: .value("Type", point.series)
+            )
+            .foregroundStyle(colorMap[point.series] ?? Color.gray)
+            .lineStyle(.init(lineWidth: 1.5))
+            .interpolationMethod(.linear)
```
marv-out, line 241: in `ForecastView.swift` in `private func drawForecastsCone() -> some ChartContent` we are using `.interpolationMethod(.catmullRom)` instead of `linear`. And in `private func drawForecastsLines() -> some ChartContent ` in `ForecastView.swift` we're not using any interpolationMethod. Here you're using `linear` for both functions. Is that on purpose? Otherwise I would suggest adapting that to what we're doing in the main chart. It probably won't be a very obvious change though...

LiveActivity/Views/LiveActivityChartView.swift
```diff
@@ -22,9 +22,13 @@ struct LiveActivityChartView: View {
...
         let maxThreshhold: Decimal = isWatchOS ? 220 : 300
 
-        // Determine scale
-        let minValue = min(additionalState.chart.min(by: { $0.value < $1.value })?.value ?? 39, 39)
-        let maxValue = max(additionalState.chart.max(by: { $0.value < $1.value })?.value ?? maxThreshhold, maxThreshhold)
+        // Determine scale, accounting for both glucose history and prediction values
+        let chartMin = additionalState.chart.min(by: { $0.value < $1.value })?.value ?? 39
+        let chartMax = additionalState.chart.max(by: { $0.value < $1.value })?.value ?? maxThreshhold
+        let forecastMin = additionalState.minForecast.min().map { Decimal($0) } ?? chartMin
+        let forecastMax = additionalState.maxForecast.max().map { Decimal($0) } ?? chartMax
```
marv-out: I could imagine that simply using the max forecast value could potentially result in reduced readability when there is a forecast with a very high max whereas the glucose values are in range, especially for such a small chart like the LA chart. I remember that y axis scaling was a pain for the main chart... maybe this needs some sort of clamping, but we need to test

marv-out (changes requested):
Made some small comments. Overall looking good  :)
2 things to notice: 

1. the zero temp predictions didn't show up when I tested that in my sim. Maybe a problem with the color asset that is not available for the LA target? Idk, haven't looked into that
2. I could imagine that we will be running into problems with update frequency and stale forecasts. Hence we made the decision a while ago to not add this feature to the small LA. But we'll see xD

Good job @t1dude!

t1dude:
Thanks for the comments @marv-out! I'll have a look into your suggestions, but first, let's discuss an alternative way to get the information we are looking for: the expected glucose forecast.

Personally I get more accurate information if I can see **glucose**, **delta**, **trend** arrow, **eventualBG** and **IOB**. That tells me more about what's going to happen than any lines or cones.

Here's a draft of what that could look like. Should we move on with that, or maybe add it as an alternative view option? I'm not trying to flood this with new features, I'm just trying to find out what information would help us on the lock screen widget.
Alternative_LA_Widget

marv-out:
Personally, I don't use the forecasted values at all. I'm using iob, cob, glucose and delta glucose. We can of course include `eventualBG` to the `LiveActivityItem` enum and just let the user choose whether he wants to see that or not. I think thats a good option. 

(But I wouldn't include the glucose bobble (again))

dnzxy:
I'd suggest we address the change-request comments Marvin has put here (based on the proposed code) and move forward with this.

t1dude:
> Made some small comments. Overall looking good :) 2 things to notice:
> 
> 1. the zero temp predictions didn't show up when I tested that in my sim. Maybe a problem with the color asset that is not available for the LA target? Idk, haven't looked into that
> 2. I could imagine that we will be running into problems with update frequency and stale forecasts. Hence we made the decision a while ago to not add this feature to the small LA. But we'll see xD
> 
> Good job @t1dude!

Yes, there are issues with the colors - I'm writing them out in the LA target now to get them to work properly. Thanks for pointing me in the right direction :)

LiveActivity/Views/LiveActivityChartView.swift
t1dude: Got it, thanks!

LiveActivity/LiveActivity.swift
t1dude: Yes, I've had it crash a couple of times when I tried to add too much data. The limitation does not seem to be documented anywhere but I got the amount of data down to something that (according to what Google can tell me) is within "normal".

LiveActivity/Views/LiveActivityChartView.swift
t1dude: I kept .linear for both intentionally. Widget extensions run under tighter resource constraints than the main app, and catmullRom takes more to render. Since the lock screen chart is small enough that you can't really tell the difference anyway, it's not worth the extra CPU cycles. That said, I'm okay with changing it to fit to what the main app is doing. Let me know what you think @marv-out .

LiveActivity/Views/LiveActivityChartView.swift
t1dude: This is a good point. For now I will make sure the predictions are limited to the existing limitation of the chart.

t1dude:
Please have another look at this @marv-out and @dnzxy . I think it might be good-to-go.

marv-out:
Code LGTM. Can test it tomorrow again

marv-out:
Not sure about hiding the iob forecast though. Does it work like that on the phone? Not sure as I use the cone 😂 @dnzxy?

marv-out:
I will approve and dan will give this a test today. But you need to address the merge conflicts @t1dude  xD

marv-out (approved):

t1dude:
> I will approve and dan will give this a test today. But you need to address the merge conflicts @t1dude xD

Rebased onto dev now. Let's see it through unit tests.

dnzxy (approved):

PR #1199, opened 2026-06-10
Use shaded CircleShape

EJM0:
Noticed ZStack still had an object in it from a commit three days earlier. If this was intended, nvm… 
image


dnzxy:
That shape has been there since 2023, when @marv-out build this fragment :)

image

Closing, because intended.

Issue #392, opened 2024-08-21
Add option for alarm after long suspension


LiroyvH:
## Is your feature request related to a problem? Please describe.

Hit a jackpot: a sensor failure and personal stupidity of forgetting to restart the pump after shower before bed. Result: no basal and no glucose values; and thus no alarms. BG 18+. Usually one or the other is no issue, in this case its a crappy scenario with a double "failure". 

## Describe the solution you'd like

I figured it'd be nice if Trio would have warned me after a while that the pump was still suspended. (It does the "inactive for 20 minutes", but that was already sent during the shower, drying, etc.) It'd be nice if this can be enabled with a "after ... hours" selector and independent selector for playing an alarm sound or not. 

## Describe alternatives you've considered

Don't be an idiot and resume insulin after shower. Alas, nobody is perfect. However, this issue can also occur during a pump error leading to suspension. 

## Additional context
No

## Technical Details
Alarm manager has to be modified. (Which wouldn't hurt to begin with anyway as its currently exceptionally limited.)

## User Impact
User has backup alarm in case of being a moron and two simultaneous issues and/or when pump suspends itself for whatever reason.

PR #1203, opened 2026-06-12
Refactor Trio Alert/Notification Handling – Prepare O5 and Libre 3/3+ Integration [1/3]

dnzxy:
## Summary

Refactors Trio's entire alert / notification handling. Routes all device + glucose alerts through a single `TrioAlertManager` (serves as LoopKit `AlertIssuer`/`AlertResponder`), introduces a user-configurable glucose alarm list (LoopFollow-style), tiered device-alarm list, a shared Day/Night window, and unifies snooze handling across phone / watch / banner / lock-screen quick-action.

**Basic functionality:**
If user has Trio in the foreground, we display an in-app banner. If Trio is backgrounded, we push a UserNotification.

## Changes
- Adds glucose alarms via user-defined list (urgent-low / low / high / forecasted-low / carbs-required), each with own threshold, schedule, sound, per-window enable, snooze. Hierarchical suppression (urgent-low silences low; low family silences forecasted-low).
- Adds `CGMManagerAlertOwnership` gate: when the active CGM provides its own glucose alerts (Dexcom G5 / G6 / One / G7 / One+, plus xDrip4iOS via the `.xdrip` App Group source), reading-driven alarms move into a dedicated "Handled by CGM App" section and the coordinator suppresses them. The "Use CGM App Alerts" toggle is auto-off + disabled when the active CGM has no companion app (e.g. LibreLoop), so users see clearly that Trio owns the alerting. The footer names the responsible app and renders the name as a deep link when a URL scheme is known.
- Adds device alarms with three severity tiers (Critical / Time-Sensitive / Normal). Each tier holds multiple configs keyed by Day/Night/Always; pump alerts pick the matching variant. Snooze is per-tier.
- Adds `AlertCatalogVendor` protocol + `AlertCatalogRegistry` — a per-pump catalog (Omnipod / Minimed / Dana / Medtrum, incl. Omnipod's dynamic `Fault Event Code 0xNN: …` hex parser) that maps each plugin-emitted `Alert.Identifier` to its `(concept, interruptionLevel, title, category)`. Replaces the previous substring classifier; concept-based dedupe collapses cross-plugin duplicates ("Reservoir Empty" etc.) in the Settings UI.
- CGM lifecycle alerts (sensor failure / expiry / signal loss) bypass the catalog and pass through with the plugin's chosen interruption level.
- Adds configurable Day & Night Window that is shared across alarms; default day starts 06:00, night starts 22:00 (or the AM/PM equivalent).
- Adds sound library of 17 bundled `.caf` tones; per-alarm picker with audible preview, override-Silence-&-Focus toggle. Ported from Loop's `next-dev` branch (thanks @ps2). Alarm row surfaces sound on/off and Overrides-Focus as separate inline tags so "silent but haptic-pierces-Focus" is visible.
- Adds critical-audio fallback: `CriticalAlertAudioPlayer` plays via `AVAudioSession(.playback)` + `MPVolumeView` for builds without the Critical Alerts entitlement; iOS otherwise silently downgrades and DnD / silent switch mute the sound. Partially taken from PR #392, inspired by both LoopFollow (thanks @bjorkert) and Loop's `next-dev` (thanks @ps2).
- Adds `com.apple.developer.usernotifications.time-sensitive` entitlement so `.timeSensitive` alerts pierce Focus modes.
- Replaces the legacy "Not-Looping" monitor with a re-armed delayed alert: every loop success cancels and re-schedules a `delayed(20m)` alert via LoopKit's `Alert.Trigger`. Background-survivable; retracts cleanly when loop and/or app recovers. Skipped automatically when closed loop is off, a manual temp basal is active, or the pump is suspended — those states imply the user isn't expecting auto-enactment.
- Adds **Carbs Required** as a configurable `GlucoseAlert` type alongside the existing low/high/forecasted family. Configured under `Settings → Glucose Alarms → Carbs Required` with its own threshold (grams), schedule, sound, snooze, and Override Silence & Focus toggle. The alarm threshold is **independent** from the home-screen tab badge's threshold (which lives in `Settings → Features → UI → Carbs Required Threshold`), so users can have a low-threshold passive badge and a higher-threshold active interrupt. Existing installs get the new alarm backfilled on launch alongside any other types added by later releases.
- Add-Alarm picker enforces a one-alarm-per-window constraint per type. Types whose Day & Night coverage is fully occupied (either a single `.always` alarm or both `.day` + `.night` set) show up under an `Unavailable` section in the picker with a footer hint pointing at the existing alarm. The editor's "Active During" dropdown is filtered to options that wouldn't collide.
- Adds new in-app banner in iOS-push style; stacked deck with critical-always-on-top, tap or swipe up to snooze 15min by default, long-press for context-menu to snooze (15m / 1h / 3h / 6h).

## Test coverage
- Per-pump alert-emission + classification: `Omnipod`, `Minimed`, `Dana`, `Medtrum`, `Tandem`, `RileyLink`.
- Catalog: `AlertCatalogRegistryEntriesTests`, `AlertCatalogRegistryOmniFaultTests` (hex-prefix parser).
- Glucose alarms: `GlucoseAlertTests`, `GlucoseAlertCoordinatorTests` (incl. carbs-required predicate guards), `GlucoseAlertConfigurationTests` (incl. DST spring-forward / fall-back), `ForecastedGlucoseEvaluatorTests`, `GlucoseAlertsStoreTests` (seed + backfill + Add-picker window gating).
- Ownership lookup: `CGMManagerAlertOwnershipTests` (manager + `.xdrip` source paths + deep-link mapping).
- Plumbing: `AlertMuterTests`, `DeviceAlertsStoreTests` (variant precedence + per-tier snooze), `TrioAlertCategoryTests`, `TrioModalAlertSchedulerTests`, `NotLoopingMonitorTests`.

## Screenshots
| Critical Alert | Time-Sensitive Alert | Info Alert | Snoozing |

| Day & Night Windows | Glucose Alarms | Device Alarms | Sound Config (w/ Play) |

| Device Alarm Form | Device Alarm Form | Glucose Alarm Form | Glucose Alarm Addition |
| image | image | image | image|

## Additional notes
- Old per-category settings are gone. Alarms now live in the new modules (`Settings → Glucose Alarms / Device Alarms / Day & Night Window`).
- Pump alerts no longer auto-acknowledge; they surface through the unified in-app banner + UserNotification pipeline until the user (or the device) clears them.
- Critical interruption level pierces snooze; Time-Sensitive / Active are fully suppressed during a snooze window.


Trio/Sources/APS/APSManager.swift
```diff
@@ -8,6 +8,11 @@ import Swinject
 
 protocol APSManager {
     func heartbeat(date: Date)
+    /// Mark the next loop attempt as user-initiated (e.g. force-loop button).
```
jwoglom, line 11: Smart, i like this

Trio/Sources/APS/APSManager.swift
```diff
@@ -714,10 +735,9 @@ final class BaseAPSManager: APSManager, Injectable {
             throw APSError.apsError(message: "Pump not set")
         }
 
-        // Check if pump is suspended and abort if it is
         if pump.status.pumpStatus.suspended {
-            info(.apsManager, "Skipping enactDetermination because pump is suspended")
-            return // return without throwing an error
+            debug(.apsManager, "Skipping enactDetermination because pump is suspended")
```
jwoglom, line 764: Doesn't have to be in this PR, but since we were talking about improving the “loop history/performance” stats page, seems like this is where we would save additional state on why a loop cycle didn't occur

Trio/Sources/APS/APSManager.swift
```diff
@@ -1208,9 +1228,133 @@ final class BaseAPSManager: APSManager, Injectable {
...
+
+    private func issueAlertForCategory(_ category: TrioAlertCategory, title: String, body: String) {
+        let content = Alert.Content(
+            title: title,
+            body: body,
+            acknowledgeActionButtonLabel: String(localized: "OK")
+        )
+        let alertId: String
+        switch category {
+        case .bolusFailed: alertId = "bolusFailed"
```
jwoglom: Can we have the string category value be an attribute stored on the enum to avoid the case statement here (which will probably need to be repeated somewhere else as well)?

Trio/Sources/APS/APSManager.swift
```diff
@@ -1208,9 +1228,133 @@ final class BaseAPSManager: APSManager, Injectable {
...
+    private func describeForAlert(
+        _ error: Error,
+        category _: TrioAlertCategory
+    ) -> (alertId: String, title: String, body: String) {
+        if let apsError = error as? APSError {
+            switch apsError {
+            case let .pumpError(inner):
+                return (
+                    "pumpError",
+                    String(localized: "Pump Error"),
```
jwoglom, line 1348: Title will have “Error” included on it for only some of these types. Should we add an “Error” / “Alert” / etc suffix on all these?

Trio/Sources/APS/APSManager.swift
```diff
@@ -1208,9 +1228,133 @@ final class BaseAPSManager: APSManager, Injectable {
...
+            trigger: .immediate,
+            interruptionLevel: category.interruptionLevel
+        )
+        trioAlertManager?.issueAlert(alert)
+    }
+
+    private func describeForAlert(
+        _ error: Error,
+        category _: TrioAlertCategory
+    ) -> (alertId: String, title: String, body: String) {
```
jwoglom: if alertId is a category should it be a string? It looks like you are only using this variable as an internal id within Alert.Identifier but still maybe worth returning an enum and then just stringifying it when passing it as a literal id

Trio/Sources/Models/GlucoseAlerts/DeviceAlertSeverity.swift
```diff
@@ -0,0 +1,54 @@
...
+        case .critical: return String(localized: "Critical")
+        case .timeSensitive: return String(localized: "Time-Sensitive")
+        case .normal: return String(localized: "Normal")
+        }
+    }
+
+    var blurb: String {
+        switch self {
+        case .critical:
+            return String(localized: "Overrides Silence & Focus Mode. Always audible. For hazardous failures.")
```
jwoglom: Nit: “for situations requiring immediate attention”

Trio/Sources/Models/GlucoseAlerts/PumpAlertCategory.swift
```diff
@@ -0,0 +1,94 @@
...
+        case .deviceExpirationReminder: return String(localized: "Device Expiration Reminder")
+        case .deviceExpired: return String(localized: "Device Expired")
+        case .podShutdownImminent: return String(localized: "Shutdown Imminent")
+        case .suspendTimeExpired: return String(localized: "Suspend Time Expired")
+        case .glucoseDataStale: return String(localized: "Glucose Data Stale")
+        }
+    }
+
+    /// Maps this category to one of the three tier configs the user edits.
+    var defaultSeverity: DeviceAlertSeverity {
```
jwoglom: Nit: within each category can you have them sorted here matching their definition above. Eg occlusion,  hardwareFault are listed first above so have them also in that order here

Trio/Sources/Models/TrioSettings.swift
```diff
@@ -213,44 +205,6 @@ extension TrioSettings: Decodable {
             settings.delay = delay
         }
 
-        if let notificationsPump = try? container.decode(Bool.self, forKey: .notificationsPump) {
```
jwoglom, line 217: Will we want to migrate any of these existing options on upgrade

Trio/Sources/Modules/AlarmWindows/View/AlarmWindowsRootView.swift
```diff
@@ -0,0 +1,114 @@
...
+                                .font(.footnote)
+                                .foregroundColor(.secondary)
+
+                            Spacer()
+                            Button(
+                                action: {
+                                    hintLabel = String(localized: "Day and Night Window")
+                                    selectedVerboseHint = AnyView(
+                                        VStack(alignment: .leading, spacing: 10) {
+                                            Text("Default: Day starts 06:00, Night starts 22:00.").bold()
```
jwoglom: Nit: may want to show am/pm 12hr time in locales which dont use 24hr

Trio/Sources/Modules/GlucoseAlerts/View/Components/AlarmAudioSection.swift
```diff
@@ -0,0 +1,168 @@
...
+            try AVAudioSession.sharedInstance().setCategory(.playback, mode: .default, options: [.mixWithOthers])
+            try AVAudioSession.sharedInstance().setActive(true, options: [])
+            let p = try AVAudioPlayer(contentsOf: url)
+            p.delegate = previewDelegate
+            p.prepareToPlay()
+            guard p.play() else { return }
+            player = p
+            playing = filename
+        } catch {
+            // best-effort preview — ignore
```
jwoglom: Might want to log the audio issue if it occurs

Trio/Sources/Modules/GlucoseAlerts/View/Components/AlarmBGSection.swift
```diff
@@ -0,0 +1,54 @@
...
+    var body: some View {
+        Section(
+            header: Text(header),
+            footer: footer.map { Text($0) }
+        ) {
+            VStack(spacing: 0) {
+                HStack {
+                    Text(title)
+                    Spacer()
+                    Text(valueMgDL.formatted(for: units))
```
jwoglom, line 23: This should use mmol in the ui when enabled i suspect?

Trio/Sources/Modules/Main/MainStateModel.swift
```diff
@@ -200,49 +200,6 @@ extension Main {
             SwiftMessages.show(config: config, view: view)
         }
 
-        /*
```
jwoglom, line 203: If im reading this right,  looks like this was a rileylink specific hack to only alert on communication failure lasting 10 mins or more, which we are already doing by default in this pr across all pump managers

Trio/Sources/Modules/PumpConfig/PumpConfigStateModel.swift
```diff
@@ -49,7 +51,32 @@ extension PumpConfig {
...
+        }
+
+        func fireTestAlert(critical: Bool) {
+            let identifier = Alert.Identifier(
+                managerIdentifier: "Trio.test",
+                alertIdentifier: "test-\(UUID().uuidString.prefix(8))"
+            )
+            let content = Alert.Content(
+                title: critical ? "Critical test alert" : "Test alert",
+                body: "If you see this banner in-app, the foreground modal scheduler is wired. Background it before tapping to test the UN push path.",
```
jwoglom: Might be semi intentional but the message text here is kind of technobabble if this is meant to be kept as a way of testing what alerts look like

Trio/Sources/Modules/Settings/SettingItems.swift
```diff
@@ -322,24 +322,44 @@ enum SettingItems {
...
-                "Always Notify CGM",
-                "Always Notify Carb",
-                "Always Notify Algorithm",
-                "Show Glucose App Badge",
-                "Glucose Notifications",
-                "Add Glucose Source to Alarm",
-                "Low Glucose Alarm Limit",
-                "High Glucose Alarm Limit"
+                "Urgent Low Glucose",
+                "Low Glucose",
```
jwoglom, line 329: if these are keywords for settings search might wanna put “cgm” in here somewhere

Trio/Sources/Services/Alerts/DeviceAlertsStore.swift
```diff
@@ -0,0 +1,145 @@
...
+    private static func severityRank(_ severity: DeviceAlertSeverity) -> Int {
+        switch severity {
+        case .critical: return 0
+        case .timeSensitive: return 1
+        case .normal: return 2
+        }
+    }
+
+    private static func activeRank(_ option: ActiveOption) -> Int {
+        switch option {
```
jwoglom, line 158: Does this ordering mean a custom day or night setting is overridden by an always setting? should this be flipped or am i reading it wrong. Since severity critical is rank 0 and normal 2 i would think always should have value greater than day or night

Trio/Sources/Services/Alerts/GlucoseAlertCoordinator.swift
```diff
@@ -0,0 +1,303 @@
...
+        switch alarm.type {
+        case .low,
+             .urgentLow:
+            return String(
+                format: String(localized: "Glucose %1$@ at or below %2$@."),
+                valueString, limitString
+            )
+        case .forecastedLow:
+            return String(
+                format: String(localized: "Forecast +20 min %1$@, at or below limit %2$@."),
```
jwoglom: Use the defaultHorizonWindow const here or if it should be overridable, pass the applied horizon min amount here instead of hardcoding 20min

Trio/Sources/Services/Alerts/GlucoseAlertCoordinator.swift
```diff
@@ -0,0 +1,303 @@
...
+
+    private func bodyText(for alarm: GlucoseAlert, valueMgDL: Decimal) -> String {
+        let units = settingsManager.settings.units
+        let valueString = valueMgDL.formatted(withUnits: units)
+        let limitString = alarm.thresholdMgDL.formatted(withUnits: units)
+        switch alarm.type {
+        case .low,
+             .urgentLow:
+            return String(
+                format: String(localized: "Glucose %1$@ at or below %2$@."),
```
jwoglom: Nit:  “at or below/above” is super verbose. Why not just say below when it is below and special case the equals case

Trio/Sources/Services/Alerts/GlucoseAlertCoordinator.swift
```diff
@@ -0,0 +1,303 @@
...
+        switch alarm.activeOption {
+        case .always: return true
+        case .day: return !configuration.isNight(at: date)
+        case .night: return configuration.isNight(at: date)
+        }
+    }
+
+    private func fetchLatestReading() -> GlucoseStored? {
+        do {
+            let cutoff = Date().addingTimeInterval(-20 * 60)
```
jwoglom: does this mean we can treat a cgm reading 19min old as current from the perspective of the cgm notifications? Would we auto dismiss an alert after it is 20min out of date? Want to make sure this isnt a magic constant which bites us later or isnt factored into edge case handling

Trio/Sources/Services/Alerts/GlucoseAlertsStore.swift
```diff
@@ -0,0 +1,98 @@
...
+        var forecasted = GlucoseAlert(type: .forecastedLow)
+        forecasted.isEnabled = false
+        let low = GlucoseAlert(type: .low)
+        let high = GlucoseAlert(type: .high)
+        return [urgent, low, forecasted, high]
+    }
+
+    private func bind() {
+        $alerts
+            .dropFirst()
```
jwoglom, line 66: Not sure what dropFirst does here

Trio/Sources/Services/Alerts/TrioAlertCategory.swift
```diff
@@ -0,0 +1,254 @@
...
+    case glucoseLow
+    case glucoseForecastedLow
+    case glucoseHigh
+    case glucoseDataStale
+    case algorithmError
+    case commsTransient
+    case other(String)
+
+    var isAlertWorthy: Bool {
+        switch self {
```
jwoglom, line 39: Nit: this really is whether an alert is worthy of triggering a notification immediately so the name is a bit confusing. I would rename

Trio/Sources/Services/Alerts/TrioAlertCategory.swift
```diff
@@ -0,0 +1,254 @@
...
+             .deliveryUncertain,
+             .glucoseUrgentLow,
+             .hardwareFault,
+             .notLooping,
+             .occlusion,
+             .reservoirEmpty:
+            return .critical
+        case .batteryLow,
+             .bolusFailed,
+             .deviceExpired,
```
jwoglom, line 110: Some of these are kind of borderline or maybe the category slugs can be improved. For instance is device expired that the device HAS expired or will expire soon. Sensor failure might also be one worth clarifying into separate temporarySensorError and permanentSensorFailure for instance (for g7 first would be ”---” state while second would be ”sensor failed replace now“)

Trio/Sources/Services/Alerts/TrioAlertCategory.swift
```diff
@@ -0,0 +1,254 @@
...
+    ///     MedtrumKit (MedtrumKit bypasses LoopKit `AlertIssuer` today —
+    ///     submodule change required to route through this classifier).
+    ///   - CGMs: CGMBLEKit (Dexcom G5/G6), G7SensorKit, LibreTransmitter,
+    ///     EversenseKit, NightscoutRemoteCGM, dexcom-share-client-swift.
+    static func categorize(alertIdentifier: String) -> TrioAlertCategory {
+        let id = alertIdentifier.lowercased()
+
+        // Glucose family — owned by GlucoseAlertCoordinator, return early so the
+        // pump/device interception in TrioAlertManager doesn't apply tier config.
+        if id.contains("glucose.urgentlow") || id.contains("glucoseurgentlow") { return .glucoseUrgentLow }
```
jwoglom: The (lack of) indenting here is weird and makes this harder to read

Trio/Sources/Services/Alerts/TrioAlertCategory.swift
```diff
@@ -0,0 +1,254 @@
...
+             .commsTransient,
+             .deviceExpirationReminder,
+             .other:
+            return .active
+        }
+    }
+}
+
+enum TrioAlertClassifier {
+    /// Classify a LoopKit alert identifier coming from a pump or CGM manager.
```
jwoglom: We need unit testing with the actual slugs used across the drivers for this logic

Trio/Sources/Services/Alerts/TrioAlertCategory.swift
```diff
@@ -0,0 +1,254 @@
...
+            return .hardwareFault
+        }
+
+        // Sensor / session failure — N7 (CGM-side).
+        if id.contains("sensorfailed") || id.contains("sensor.failed") || id.contains("sensorstopped")
+            || id.contains("sensorerror") || id.contains("invalidsensor") || id.contains("encryptedsensor")
+            || id.contains("sensortemperature") || id.contains("sensorlowtemperature")
+            || id.contains("readertemperature") || id.contains("sensorretirement") || id.contains("nosensordetected")
+            || id.contains("transmitterdisconnected") || id.contains("glucosesuspended")
+            || id.contains("sensorconnection")
```
jwoglom: Want us to check to make sure some of these arent picking up other non failed states

Trio/Sources/Services/Alerts/TrioAlertCategory.swift
```diff
@@ -0,0 +1,254 @@
...
+        {
+            return .sensorFailure
+        }
+
+        // Device expired — N6 (pod, sensor, transmitter end-of-life).
+        if id.contains("podexpired") || id.contains("podexpiring") || id.contains("sensorexpired")
```
jwoglom: Wouldnt podexpiring be deviceExpirationReminder

Trio/Sources/Services/Alerts/TrioAlertManager.swift
```diff
@@ -0,0 +1,447 @@
...
+            muted: muted,
+            soundURL: soundLoader.url(for: effective)
+        )
+        playCriticalAudioFallbackIfNeeded(effective, muted: muted)
+    }
+
+    private func applyDeviceSeverityConfig(to alert: Alert, category: PumpAlertCategory) -> Alert? {
+        let severity = category.defaultSeverity
+        let now = Date()
+        let isNight = GlucoseAlertsStore.shared.configuration.isNight(at: now)
```
jwoglom: Could someone ever want some hours of the day not expressed in either night or day hours? I dont think so but wanted to check

Trio/Sources/Services/Alerts/TrioAlertManager.swift
```diff
@@ -0,0 +1,447 @@
...
+
+// MARK: - Muter
+
+final class AlertMuter: ObservableObject {
+    @Published private(set) var startDate: Date?
+    @Published private(set) var duration: TimeInterval = 0
+
+    static let allowedDurations: [TimeInterval] = [
+        30 * 60,
+        60 * 60,
```
jwoglom, line 394: Seems reasonable to think someone might want to set some other intermediate values or even longer potentially

Trio/Sources/Services/Alerts/TrioModalAlertScheduler.swift
```diff
@@ -0,0 +1,350 @@
...
+        active.removeAll { $0.identifier == identifier }
+    }
+
+    private func sortByPriority() {
+        active.sort { Self.priority(of: $0.interruptionLevel) > Self.priority(of: $1.interruptionLevel) }
+    }
+
+    private static func priority(of level: LoopKit.Alert.InterruptionLevel) -> Int {
+        switch level {
+        case .critical: return 2
```
jwoglom: I think the other criticality sorting earlier in this pr had 0 ranked first. May want to ne consistent

Trio/Sources/Services/Alerts/TrioModalAlertScheduler.swift
```diff
@@ -0,0 +1,350 @@
...
+            DragGesture(minimumDistance: 12)
+                .onChanged { value in
+                    let dx = value.translation.width
+                    dragOffset = abs(dx) > abs(value.translation.height) ? dx : 0
+                }
+                .onEnded { value in
+                    let dx = value.translation.width
+                    // Swipe right past threshold → dismiss / acknowledge.
+                    // Swipe left past threshold → quick snooze (20m) when
+                    // the alert isn't critical. Otherwise reset.
```
jwoglom: What does reset mean here

Trio/Sources/Services/Alerts/TrioUserNotificationAlertScheduler.swift
```diff
@@ -0,0 +1,100 @@
...
+            identifier: alert.identifier.value,
+            content: content,
+            trigger: alert.trigger.unTrigger
+        )
+    }
+
+    private func sound(for alert: Alert, muted: Bool, soundURL: URL?) -> UNNotificationSound? {
+        let isCritical = alert.interruptionLevel == .critical
+        if muted {
+            return isCritical ? .defaultCriticalSound(withAudioVolume: 0) : nil
```
jwoglom, line 62: What does withAudioVolume: 0 do here

Trio/Sources/Services/UserNotifications/UserNotificationsManager.swift
```diff
@@ -570,23 +296,10 @@ extension BaseUserNotificationsManager: alertMessageNotificationObserver {
             content.title = message.title
         }
         switch message.subtype {
-        case .pump:
-            if message.type == .info || message.type == .error {
```
jwoglom, line 574: Why are pump and algorithm being removed here with carb and glucose staying? Does it have to do with who (Trio or the pump manager) being responsible for firing the alert?

Trio/Sources/APS/APSManager.swift
dnzxy: I like it!

Trio/Sources/APS/APSManager.swift
dnzxy: May be a good idea. Haven't thought that far. Let me see to it.

Trio/Sources/Models/TrioSettings.swift
dnzxy: I genuinely woldn't know how, it does not really translate. Before this refactor, we basically had two states: no notifications at all (other than critical) or all of them. This does not translate to the severity-type wiring we now have for device alarms. 

I'd suggest we just notify user about new release (once this ships) and then they have to just re-config their alarms. They will want to do that anyways.

Trio/Sources/Modules/AlarmWindows/View/AlarmWindowsRootView.swift
dnzxy: Yes!

Trio/Sources/Modules/GlucoseAlerts/View/Components/AlarmBGSection.swift
dnzxy: That's what `.formatted(for: units)` does. It internally grabs the unit, optionally parses to mmol/L, and outputs a (parsed) string.

Trio/Sources/Modules/Main/MainStateModel.swift
dnzxy: Yes, hence why it was dropped.
This will need testing, but I hope it goes well!

Trio/Sources/Modules/PumpConfig/PumpConfigStateModel.swift
dnzxy: Just for testing – will remove this before we merge.
Let me add a `FIXME`.

Trio/Sources/Services/Alerts/GlucoseAlertCoordinator.swift
dnzxy: It's a hard-set 20 min (4th forecasted value in each of the predBG arrays) window. That is the applied horizon.

Trio/Sources/Services/Alerts/GlucoseAlertCoordinator.swift
dnzxy: Theoretically, if you were to force quit the app, there hasn't been an alert yet (for whatever reason), and there's nothing snoozed, then this could potentially lead to a glucose alert for a 19min and 59 sec old reading, yes. Unless I am reading this wrong myself here now. 

I was planning on adding unit tests for all this, so lets cover this via testing and see how it behaves, potentially finetune.

Trio/Sources/Services/Alerts/TrioAlertCategory.swift
dnzxy: I wanted to combine them, maybe I have done a bit of a too good job here. 
deviceExpired would be a patch or pod expired alarm.

Trio/Sources/Services/Alerts/TrioAlertManager.swift
dnzxy: I asked myself the same thing, but this now follows LoopFollow and also how the official Dexcom app lets you define alerts for two "sections" of day. I think that's enough without making it overly complex.

Trio/Sources/Services/Alerts/TrioAlertManager.swift
dnzxy: We have had this in place already, and context menus on lock screens only allow for 4 options, hence these 4.

Trio/Sources/Services/Alerts/TrioModalAlertScheduler.swift
dnzxy: It clears/acks it.

Trio/Sources/Services/UserNotifications/UserNotificationsManager.swift
dnzxy: Good question. Lemme see.

Trio/Sources/Services/Alerts/TrioUserNotificationAlertScheduler.swift
dnzxy: `. defaultCriticalSound()` returns the system critical alert sound but lets us scale playback volume (0.0–1.0). 0 means play silently, so we use it for critical alerts during a snooze (or for critical alerts that were issued with .vibrate/no sound) so the notification still bypasses Silent + Focus. Without a critical-class sound iOS demotes the notification to regular tier and DnD/Focus would suppress it, I'll look into having this overriden, too.

Trio/Sources/Services/Alerts/GlucoseAlertsStore.swift
dnzxy: $alerts fires immediately on subscribe with the current value. Without dropFirst(), the bind would re-encode the just-loaded state back to UserDefaults on app launch (pointless write, also overwrites if the load failed/returned defaults). dropFirst() skips that initial emission, so we only persist on actual user changes. At least that's the idea here, lol

jwoglom: weirdly named but ok makes sense

jwoglom:
fyi on the latest 2 commits:

Trio
/Users/james/repos/Trio-tandem/Trio/Sources/Services/Alerts/TrioModalAlertScheduler.swift
/Users/james/repos/Trio-tandem/Trio/Sources/Services/Alerts/TrioModalAlertScheduler.swift:251:49 Ambiguous use of operator '/'

Found this candidate in module 'Swift'
Found this candidate in module 'CoreFoundation'

Trio/Sources/Services/Alerts/AlertCatalogRegistry.swift
```diff
@@ -143,7 +143,14 @@ private extension AlertCatalogRegistry {
...
             .insulinLimitReached
         ),
+        // MedtrumKit currently emits the occlusion alert with a misspelled
+        // identifier ("patch-occlussion", double "s") — see MedtrumKit's
+        // NotificationManager. We register both the misspelled identifier (what
+        // ships today) and the corrected spelling so the alert escalates to
+        // critical regardless of which MedtrumKit is bundled. Drop the
+        // misspelled entry once the typo is fixed upstream.
         addEntry("Medtrum", "com.nightscout.medtrumkit.patch-occlussion", .critical, "Occlusion", "Delivery", .occlusion),
+        addEntry("Medtrum", "com.nightscout.medtrumkit.patch-occlusion", .critical, "Occlusion", "Delivery", .occlusion),
```
jwoglom, line 153: there's no harm keeping the entry though for now, right? in case someone was running an older MedtrumKit for whatever reason?

trioneer-dev:
image

jwoglom (approved):
this PR is getting large, but I've been testing this branch for the past week or so and outside of minor formatting nitpicks, the core seems solid. Let's get this out to O5 testers.

PR #981, opened 2026-02-19
Add CGM status highlight

bastiaanv:
With CGM's like Eversense and Accu Chek, it is important to be aware of any required actions, like calibrations or critical issues. LoopKit provides the `cgmStatusHighlight` for this use case. Currently, only the G7 CGM has implemented this, but I believe this should be implemented in every CGMKit and Trio should show this highlight.

Current proposal (when clicking on the badge the CGM sheet opens, design is copied from the pump timezone warning):

IMG_1385
IMG_1386

If this proposal gets approved, the following items need to be done:
- [ ] Are we cool with the current design?
- [ ] Should we show all highlights or only warning/critical? -> Should we show `Searching for sensor` [link](https://loopkit.github.io/loopdocs/loop-3/add-cgm/#dexcom-g7-or-one)
- [ ] Do we want to show the image?
- [ ] Implement `cgmStatusHighlight` on other CGMKit's, like LibreTransmitter, G6/G7, EversenseKit

Issue #304, opened 2024-06-14
Show CGM life left on main screen


JohnKitching:
In the same way that life left of a pump is displayed in the main screen it would be very handy to see the same for the CGM. 

I have no opinion on how it's displayed, just that I'd like to see it. 

It would help with planning CGM changes around a busy life. I'm guessing this would help most people.

LiroyvH:
Not sure how much life the current design has left in it, its a bit short on space. But maybe it'd be doable to implement a countdown bar thingy like xDrip4iOS has. You'd sacrifice a bit of graph space, but I think we have plenty of that. Alternatively the BG-label in the legend could be dropped and the eventual glucose moved to top right (left: basalrate, right: eventual glucose) to free up some space at the bottom. Get's crowded tho.

dnzxy:
Like the idea a lot. How do you guys imagine to handle Anubis and extended sensors in this type of display? 🤔

JohnKitching:
I'd say that if it's possible, if there's a known life then display it, if there isn't (Anubis etc) then display nothing.

bjornoleh:
Here are some samples of how xDrip4iOS handles Anubis (G6) before and ten days. 

Before:
<img src="https://github.com/nightscout/Trio/assets/63544115/604e317c-e2be-4c22-b870-9f73d4db296e" width="300">


After ten days:
<img src="https://github.com/nightscout/Trio/assets/63544115/0b7d95ac-57ce-4368-b200-7e765ec42203" width="300">

LiroyvH:
I'd suggest that for Anubis either we show nothing or indeed just keep the counter going after 10 days. Anubis can be detected easily by the 180 days in the txAge, stock Dexcom won't ever have that. For my Anubis I'm typically more interested in the VoltB than SAGE. :P

bjornoleh:
This could still be an interesting feature. I set the `long-term` label now.

dnzxy:
Maybe this would be something down your lane and you'd want to have a look at this, once public beta drops @aidanlane 😎

aidanlane:
Thanks for thinking of me @dnzcy, but I now have another feature that I’d like to focus on first 😉

LiroyvH:
@Sjoerd-Bo3 That's horrible for OCD with the alignment and mismatching amount of rows haha :) Might it be possible to move the adjustments info section up a tiny little bit and then below that put a countup or countdown bar for sensor age? (Like how xDrip4iOS has it implemented).

[...]

PR #1205, opened 2026-06-14
Add CGMStatusHighlight Handling: CGM sensor lifetime indicator (arc + tag) – Prepare Libre 3/3+ Integration [2/3]

dnzxy:
## Summary

Adds a sensor lifetime indicator to the home-screen glucose bobble:
- **Outer arc** — drains clockwise from 12 o'clock over the sensor's life, sourced from `cgmLifecycleProgress`. Green → orange → red based on `progressState`.
- **Tag below the bobble** — short label showing time remaining (`5d 14h`, `22h left`) or a status word (`warming up`, `calibrate`, `sensor expired`, `sensor error`).
- Six sensor states: normal / expiring soon / warmup / calibration required / expired / sensor failed. Warmup + failure hide the arc and mute the glucose number to `– –`.
- **G7 grace period** — bobble countdown falls back to `sensorEndsAt` once `sensorExpiresAt` passes (was collapsing to `<1m` while the arc still showed ~12h).

‼️ **CAVEAT**
This does for users using CGMBLEKit (Dexcom G5/G6/One) through a fork and an addition to CGMBLEKit, not via the current version of the submodule as found in `dev` and `main`.
LibreTransmitter has not been tested, should work though.
xDrip4iOS will get an update providing us with CGM lifecycle and status highlight _like_ data, but that has not been shipped or checked in to xDrip4iOS yet (thanks to @paulplant for the collab on this)

Some Follow-up PRs like needed.

---

## How sensor lifetime is resolved

`HomeStateModel.sensorExpiresAt(for:lifecycle:)` resolves the expiration date in this order:
1. **Direct read** when the manager exposes it (e.g., `sensorExpiresAt`). For G7, falls back to `sensorEndsAt` once the primary lifetime has passed so the grace window is counted down instead of clamped to zero.
2. **Reverse-derive** from `activatedAt + percentComplete`: `totalLifetime = elapsed / percentComplete`, `expiresAt = activatedAt + totalLifetime`. Picks up `activatedAt` via the same type-check pattern already used in `PluginSource.readCGMResult`
3. **Fallback** — `XX%` label when neither is available.

## Extended CGM Simulator: Scenario picker

The glucose simulator exposes `cgmStatusHighlight` + `cgmLifecycleProgress` so the indicator is exercisable without real CGM hardware. CGM Settings → Glucose Simulator → "Sensor Lifecycle Scenario" picker flips between all six states.

To match the data shape a real CGM would deliver in each state:
- Non-active scenarios (warmup / calibration / expired / failed) **stop emitting fresh readings**, so the bobble's stale path fires and the highlight's `imageName` surfaces.
- Picker change also **drops `GlucoseStored` rows inside the 12 min freshness window**, so the bobble flips immediately instead of waiting 12 minutes for organic aging.
- `HomeStateModel`'s 30 s refresh timer reads `cgmLifecycleProgress` / `cgmStatusHighlight` from the simulator source when no live `CGMManager` is set, so the same code path serves both real and simulated sensors.

## Screenshots

| Nominal | Expiring soon | Expired |
|--------|--------|--------|
| image | image | image |

| Warmup | Calibration | Error |
|--------|--------|--------|
| image> | image | image |

Addresses #304 , picks up where #981 left off (thanks for the previous work @bastiaanv ).


bastiaanv:
I've tested this using the Eversense 365 demo unit.
It shows the progress in percent, most likely because there is no mapping yet for EversenseKit.

I like the UI, but a lot of whitespace between the carbs & percentage and not too sure why you would want to use the custom message mapping.

Showing the `CGMManager.cgmStatusHighlight.localizedMessage` and let the CGMManager handle the warning/error messages (does require an update for LibreTransmitter).

Also not entirely convinced that the progressbar is working (see screenshot 2).

IMG_1709

Screenshot 2026-06-14 at 17 17 41

dnzxy:
> I've tested this using the Eversense 365 demo unit. It shows the progress in percent, most likely because there is no mapping yet for EversenseKit.

This is only hooking into the CGM Manager like your previous PR #981 did @bastiaanv 
I'm not sure what else needs to be mapped out yet? Could you point me to that, please?

> I like the UI, but a lot of whitespace between the carbs & percentage and not too sure why you would want to use the custom message mapping.

Good finding. I haven't added any extra whitespace there, it should just use the space that was there before. It does slightly push the rest down, so let me see about that.

> Showing the `CGMManager.cgmStatusHighlight.localizedMessage` and let the CGMManager handle the warning/error messages (does require an update for LibreTransmitter).

And CGMBLEKit, it does not expose the transmitter lifecycle. I've already asked @ps2 about it :) 
@dabear are you open to extending LibreTransmitter for this?

> Also not entirely convinced that the progressbar is working (see screenshot 2).

Let me try and find why it's not working, but I am working off of "mocks" here.

dnzxy:
@bastiaanv okay so I have slightly reworked the UI representation and also when to propagate the status, but the wiring remained the same. It should work out of the box for Eversense (following PR #981). I did however add a setter / refresh logic for the cgm state so the UI gets updated. 

Locally, in-vivo, running with the changes proposed via https://github.com/loopandlearn/CGMBLEKit/pull/6, this looks as follows (with a fresh G6 sensor paired via G6 app and CGMBLEKit CGM manager for G6)

image

t1dude:
Tested with G7. Works as expected, including grace period.

t1dude (approved):
This PR now works as expected. Test rig:

Phones: iPhone Air (17), iPhone 17 Pro Max
iOS version: 26.5.1
CGM: Dexcom G7 using Dexcom native app

Trio/Sources/APS/CGM/GlucoseSimulatorSource.swift
```diff
@@ -49,6 +54,57 @@ final class GlucoseSimulatorSource: GlucoseSource {
...
+        if !scenario.deliversFreshGlucose {
+            clearRecentSimulatorReadings()
+        }
+        publishSimulatedState()
+    }
+
+    /// Deletes GlucoseStored rows newer than the home view's 12 min
+    /// freshness window. Dev-only — only invoked from the simulator's
+    /// scenario picker, which is itself gated to simulator mode.
+    private func clearRecentSimulatorReadings() {
```
kingst, line 90: Is there a way to make sure that these glucose readings are all from the simulator?

Trio/Sources/APS/CGM/GlucoseSimulatorSource.swift
```diff
@@ -49,6 +54,57 @@ final class GlucoseSimulatorSource: GlucoseSource {
...
+            do {
+                let recent = try context.fetch(request)
+                for row in recent {
+                    context.delete(row)
+                }
+                if context.hasChanges {
+                    try context.save()
+                }
+            } catch {
+                print("GlucoseSimulatorSource: clearRecentSimulatorReadings failed: \(error)")
```
kingst, line 105: Should we use a log here instead of print?

kingst (commented):
It looks like this was already merged so these comments are just FYI. When I ran this with the G7 I happened to have "1d 12h" left and there was no arc, and I believe that we intended to have one during this period.

PR #1206, opened 2026-06-14
Feat/bolus initiating

bastiaanv:
This PR updates the bolus progress view to show to the user that the bolus is initiating.
This is required for the upcomming TandemKit driver, because the Tandem Mobi takes ~5s between accepting a bolus command and actually progressing the bolus.

We could also add the last bolus state: canceling. But this is not done in this PR

Demo video:

<video src="https://github.com/user-attachments/assets/2c22ff5c-d203-4305-bb1f-c64019df2192" />


dnzxy:
On first observation, looks great, thanks for this!

We may want to bring this to the treatment view, too, so the user knows what's going on when they are between a state of "Trying to enter a meal and bolus for it" and "SMB was just initiated by algorithm, I need to wait." Currently, they can only see when a bolus is already running (so it would be stuck at 0 of xx Units), potentially may be nice to also see the same "Initiating…" message.

bastiaanv:
I messed up the commit history :)

Will open a new PR

PR #1211, opened 2026-06-15
Feat: Bolus initiating

bastiaanv:
(duplicate of #1206)
(provided feedback from previous PR https://github.com/nightscout/Trio/pull/1206#issuecomment-4702981172)

This PR updates the bolus progress view to show to the user that the bolus is initiating.
This is required for the upcomming TandemKit driver, because the Tandem Mobi takes ~5s between accepting a bolus command and actually progressing the bolus.

We could also add the last bolus state: canceling. But this is not done in this PR

Demo video:

<video src="https://github.com/user-attachments/assets/2c22ff5c-d203-4305-bb1f-c64019df2192" />


Trio/Sources/Models/BolusStatus.swift
```diff
@@ -0,0 +1,7 @@
+import LoopKit
+
+enum BolusStatus {
+    case noBolus
+    case initiating
+    case inProcess
```
dnzxy: Nit: Can we name this `inProgress` please?

Trio/Sources/Modules/Home/View/HomeRootView.swift
```diff
@@ -810,22 +810,32 @@ extension Home {
                         Spacer()
 
                         VStack {
-                            Text("Bolusing")
-                                .font(.subheadline)
-                                .frame(maxWidth: .infinity, alignment: .leading)
+                            if state.bolusStatus == .inProcess {
```
dnzxy: Nit: Instead of if-else-ing the same view elements, you could conditionally assign a `var bolusStateLabel`, and then just render the same UI control once, and pass `Text(bolusStateLabel)`

dnzxy (commented):
Two smaller comments. 
This looks good, and I think will also be an improvement for users on Medtronic or Eros with Riley/OrangeLink.

Trio/Sources/Models/BolusStatus.swift
bastiaanv: Much better :D

bastiaanv:
Not sure why I get this CI error:
`error: Clang dependency scanner failure: fatal error: file '/Applications/Xcode_26.2.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator26.2.sdk/usr/include/AvailabilityVersions.h' has been modified since the module file '/Users/runner/Library/Developer/Xcode/DerivedData/ModuleCache.noindex/KUWMVTRDF6PJ/_AvailabilityInternal-FJX0ABVEZB4I.pcm' was built: mtime changed (was 1782208545, now 1782799679)`

dnzxy (approved):

bastiaanv:
@dnzxy I merged latest dev into this branch, fixed the merge conflicts and the CI is happy now :)

dnzxy (approved):

dnzxy (approved):

PR #1214, opened 2026-06-15
Enable text selection to copy SHA

### LINKED ISSUE #1210: Settings: add button to copy Trio version+hash to clipboard
Trio 0.8.1.8 (4a7029c)

_Suuurre_ would be handy to be able to copy the Trio version + SHA into clipboard as a single, concatenated string. :-)

MikePlante1:
resolves #1210 

Allows users to easily copy the SHA of Trio or of a specific submodule.

### DIFF
--- a/Trio/Sources/Modules/Settings/View/SettingsRootView.swift
+++ b/Trio/Sources/Modules/Settings/View/SettingsRootView.swift
@@ -100,30 +100,49 @@ extension Settings {
             }
         }
 
+        private func copyVersionInfo(_ text: String) {
+            UIPasteboard.general.string = text
+            UINotificationFeedbackGenerator().notificationOccurred(.success)
+            withAnimation { showCopiedToast = true }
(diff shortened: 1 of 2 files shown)


dnzxy:
May I suggest to make it into a normal tap and add a `document.on.document.fill` SF Symbol?
Or add a hint somewhere (?) Long-press gestures without any hint are just hidden from the user. 

Edit: Maybe to the left of the "Branch" label?

MikePlante1:
Thanks for the suggestion, @dnzxy. The code and screenshot have been updated.

jwoglom:
@MikePlante1 would it be better maybe if this was instead available once you have tapped into the Trio "versions" page, so that you don't have conflicting touch targets between the copy icon above the Trio button section and the button itself? plus the copy icon adds just a bit more to the busyness of this page

dnzxy (approved):
I think this is good. 

Are you okay with this, too? @jwoglom :)

PR #1219, opened 2026-06-19
Trio v0.8.3 Hotfix: Fix Delivery Limit Regression on Silent PumpManagerRawState Migration

AndreasStokholm:
Stand in committing for Dan as he's having GitHub account issues at the moment.

Should solve the `.invalidSetting` failures we've been seeing.

PR #1221, opened 2026-06-19
Sync delivery limits to pump manager on init

jwoglom:
When a pump manager is instantiated from persisted raw state rather than through the pump setup flow, it can come up with a stale or default delivery limit from its deserialized state. With pump managers that enforce the limit on temp basals (e.g. OmnipodKit), this silently rejects in-range temp basals and fails the loop.

Re-sync the user's configured max basal/bolus into the pump manager whenever one is set, ahead of the branches that may return early. Add unit tests covering the settings-to-limits derivation, including regression guards that the user's values are not collapsed to the PumpInitialSettings defaults.


dnzxy:
Closing in favor of #1219, unit tests pulled over.

Issue #1196, opened 2026-06-07
ISF value swift oref defaulting to profile isf and not adjusting


EJM0:
Hi, I wanted to sniff a bit on the cutting edge of the oref Trio verse.
So I merged the current Swift-oref branch, since @marv-out peer approved.
After merging into my hippie branch of Trio, I noticed that the Isf Chart I’m using for logarithmic setting eval. seemed to be 0ing out.

<img width="241" height="524" src="https://github.com/user-attachments/assets/f2f65018-06e0-4f04-a2bf-140ef30639dc" />

So I took a dive into it. I noticed the issue only occurs when using APSSwift, not APSjs.
The loop status also is not affected since, as I understand it, scimming the code, it only fetches the JSON status created from APSSwift and APSjs, and that seems to be right in either one.
While fixing, I additionally noticed that this issue also appears to be present in the calculation for meals. (Treatment view)

<img width="241" height="524" alt="Image" src="https://github.com/user-attachments/assets/4d3ed8d3-2dea-4d4d-a25c-70c830014977" />
<img width="241" height="524" alt="Image" src="https://github.com/user-attachments/assets/e8b8f490-6fb6-448f-8dd7-7303d0f2c683" />

What fixed it for me is changing https://github.com/nightscout/Trio/blob/caca90f8af3dbb0c6e588787d0455ecb152e697a/Trio/Sources/APS/OpenAPSSwift/DetermineBasal/DetermineBasalGenerator.swift#L380
 to: `adjustedSensitivity.jsRounded(),`

I think internally in the oref-swift implementation everything is fine, only when needed externally, e.g. for the treatment calc is defaults to the not adjusted isf set in profile.

@kingst  @dnzxy

PR #1222, opened 2026-06-20
Report adjusted ISF in determination output

bjorkert:
The Swift oref left `isf` as nil in the determination, so the ISF field uploaded to Nightscout came out as 0. This sets it to the adjusted sensitivity, the same value already shown in the reason string.

Dan found the fix, I'm just the secretary.


kingst (approved):
LGTM. There is a broader issue around running values through a `convertGlucose` function that will impact this value, but that should be handled in a separate PR that is more broad than this ISF property.

dnzxy:
Looking over our open issues, this had actually been reported in a different context by @EJM0 via https://github.com/nightscout/Trio/issues/1196, so two dummies, same solution 😂 

Thanks for @EJM0 for the initial report (and I guess solution) and to @bjorkert for providing this fix.
Should be good to go. Providing based on code review and having received test feedback on LNL Slack for this.

dnzxy (approved):

PR #1223, opened 2026-06-20
Remove unused convertGlucose function

kingst:
This PR removes an unused function from the core determine basal logic in oref-swift: `convertGlucose`. In the original JS implementation, several output properties and reason string values are run through a `convert_bg` function that converts the glucose value to `mmol` based on the `profile.outUnits` value.

However, `profile.outUnits` is always in `.mgdl`, and should always be in `.mgdl` so we should remove this function. The path goes:
- The convert function branches based on the value of `profile.outUnits`
- `profile.outUnits` is set from `bgTarget.userPreferredUnits`
- all of the write paths for the file that saves the `bgTarget` value store these values as mg/dl and hard code `userPreferredUnits` to `.mgdl`.

Unit tests still pass with this change.


dnzxy (approved):
Approving from code review and having discussed this via DM.

outUnits is hard-coded, so this never gets used, nothing gets converted.

PR #1226, opened 2026-06-21
APSManager thread safety fixes (Part 2)

marv-out:
Closes the loop-serialization race and reworks the background-task
lifecycle around the loop's async scope. Sequel to Part 1, split out
of the original https://github.com/nightscout/Trio/pull/1145 to discuss LoopGuard + background-task handling separately from the bolus-progress actor.

**LoopGuard**

  The old check-then-act pattern read `isLooping.value`, returned from
  `canStartNewLoop()`, then later (after a suspension point) called
  `isLooping.send(true)`. Two concurrent `recommendsLoop` events could
  both observe `false` before either had sent `true`, so loop() could
  run twice. Replaced by a private `LoopGuard` actor that performs the
  check-and-mark-running step as a single atomic operation. isLooping
  becomes a downstream notification mechanism.

**Background-task lifecycle**

Removed the Task in the expiration handler which didn't provide a guarantee to be executed. Aligned with the example in https://developer.apple.com/documentation/uikit/extending-your-app-s-background-execution-time. Side note: we probably need to refactor this when moving to Swift 6 as the expiration handler closure is Sendable and captures and mutates the var backgroundTaskID which isn't allowed when strict concurrency checking is on.

**Loop lifecycle consolidation**

  `canStartNewLoop`, `setupLoop`, `loopCompleted` are folded into
  loop() and a new single-exit-point `finalizeLoop`. `executeLoop` no
  longer calls finalize itself; the caller owns lifecycle. Open and
  closed loop paths in executeLoop now share the stats-recording
  block, with `enactDetermination` gated by `settings.closedLoop` at
  the end.

  `calculateLoopInterval` now takes `loopStartDate` as a parameter
  instead of reading the instance `lastLoopStartDate` — keeps the
  helper independent of the surrounding mutation order in loop().

### DIFF
--- a/Trio/Sources/APS/APSManager.swift
+++ b/Trio/Sources/APS/APSManager.swift
@@ -267,8 +267,18 @@ final class BaseAPSManager: APSManager, Injectable {
                 return
             }
 
-            // Local background-task identifier — no shared mutable state.
-            let bgTask = await UIApplication.shared.beginBackgroundTask(withName: "Loop starting")
+            // Start background task
+            // we probably need to refactor this when implementing Swift 6 due to mutation of a captured var in an async context
+            var taskID: UIBackgroundTaskIdentifier = .invalid
+            taskID = await UIApplication.shared.beginBackgroundTask(withName: "Loop starting") {
+                // closure runs on the Main Thread
+                // removed the Task that provided no guarantee to end the background task
+                if taskID != .invalid {
+                    UIApplication.shared.endBackgroundTask(taskID)
+                    taskID = .invalid
+                }
+            }
+
             isLooping.send(true)
 
             let loopStartDate = Date()
@@ -297,16 +307,17 @@ final class BaseAPSManager: APSManager, Injectable {
                 debug(.apsManager, "\(DebuggingIdentifiers.failed) Failed to complete Loop: \(error)")
             }
 
-            // End the background task on the async path itself
-            if bgTask != .invalid {
-                await UIApplication.shared.endBackgroundTask(bgTask)
+            // End the background task
+            if taskID != .invalid {
+                await UIApplication.shared.endBackgroundTask(taskID)
+                taskID = .invalid
             }
         }
     }
 
     private func executeLoop(loopStatRecord: inout LoopStats) async throws {
         try await determineBasal()
-        
+
         // Closed loop: also enact the determination.
         if settings.closedLoop {
             try await enactDetermination()
--- a/Trio/Sources/APS/APSManager.swift
+++ b/Trio/Sources/APS/APSManager.swift
@@ -306,16 +306,16 @@ final class BaseAPSManager: APSManager, Injectable {
 
     private func executeLoop(loopStatRecord: inout LoopStats) async throws {
         try await determineBasal()
+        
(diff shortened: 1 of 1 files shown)


Trio/Sources/APS/APSManager.swift
```diff
@@ -233,116 +256,81 @@ final class BaseAPSManager: APSManager, Injectable {
...
+                lastLoopStartDate: lastLoopStartDate
+            ) else {
+                debug(.apsManager, "Loop skipped (already running or too soon)")
+                return
+            }
 
-            // Setup loop and background task
-            var (loopStatRecord, backgroundTask) = await self.setupLoop()
+            // Local background-task identifier — no shared mutable state.
+            let bgTask = await UIApplication.shared.beginBackgroundTask(withName: "Loop starting")
```
kingst: It seems strange to me to not have an expiration handler, although I have to admit that I don't fully understand how this might impact us in practice. Can you explain what will happen if the OS invokes this and we don't have a call to `endBackgroundTask` within the expiration handler?

Trio/Sources/APS/APSManager.swift
```diff
@@ -233,116 +256,81 @@ final class BaseAPSManager: APSManager, Injectable {
...
-            loopStatRecord.loopStatus = "Success"
-            await loopCompleted(loopStatRecord: loopStatRecord)
-            return
-        }
-
-        // Handle closed loop
-        try await enactDetermination()
-        loopStatRecord.end = Date()
-        loopStatRecord.duration = roundDouble((loopStatRecord.end! - loopStatRecord.start).timeInterval / 60, 2)
+        let endDate = Date()
```
kingst: This isn't including `enactDetermination` within the `endDate` timing, which is different from the original code. Is this on purpose?

kingst (commented):
Overall this PR looks great, it's a real improvement. The LoopGuard is a nice addition.

I do have questions around the background task expiration handler, which I'd like to understand in more detail.

Trio/Sources/APS/APSManager.swift
marv-out: No, I've changed this back

Trio/Sources/APS/APSManager.swift
marv-out: It probably won't impact us as the loop should execute before we have to end the background task. But you're right, we should have an expiration handler just in case. But the Task in there was the problem that didn't guarantee the execution either. I tried several different more and less elegant things to refactor this. But at the end I think its probably still the best to stick with what Apple recommends in here https://developer.apple.com/documentation/uikit/extending-your-app-s-background-execution-time. Hence my last commit.

But we might get a problem in the future when migrating to Swift 6 due to the mutation of the the var taskID in the Sendable expiration handler closure

kingst (commented):
LGTM from a code review perspective, let me test it before giving it a 👍 . 

For the background task what you did looks good to me, I will approve this. But if you wanted something Swift 6 compatible you could probably refactor it into a small actor, but not needed for this PR.

kingst (approved):
Tested on my iPhone in vivo. It's hard to test this rigorously given the nature of the changes, but it was working great.

PR #1227, opened 2026-06-21
APSManager thread safety fixes (Part 3)

marv-out:
Closes the remaining mutable-state race in APSManager around bolus
progress reporting. Final piece split out of the original https://github.com/nightscout/Trio/pull/1145 to have its own architectural discussion.

**BolusProgressState**

  The old `bolusReporter` property was a plain mutable `DoseProgressReporter?`
  mutated from a Combine sink on processQueue (bolusTrigger transitions)
  and concurrently from `cancelBolus` (async context). No isolation.

  Replaced with a private `BolusProgressState` actor that owns the
  reporter and serializes add/remove-observer calls.

**Reorder race in the bolusTrigger sink**

  Pre-change, the sink called `createBolusReporter()` or `clearBolusReporter()`
  directly, both of which were sync. Making them async opens a window
  where two unstructured Tasks (one per emission) could reorder against
  each other at the actor — unstructured Tasks enter an actor in
  scheduler order, not the order they were spawned. A fast bolus
  end-then-start transition (real in practice) could land setReporter
  before its matching clear and wipe the fresh reporter.

  Both transitions are now funneled through a single Task per sink
  emission, awaiting in-order at the actor.

**doseProgressReporterDidUpdate**

  The `DoseProgressObserver` protocol method is sync, so the call to
  `clearBolusReporter()` (now async) is wrapped in a Task. The race that
  would normally make this dangerous is exactly what the generation
  token in BolusProgressState protects against.

### DIFF
--- a/Trio/Sources/APS/APSManager.swift
+++ b/Trio/Sources/APS/APSManager.swift
@@ -92,6 +92,32 @@ private actor LoopGuard {
     }
 }
 
+// MARK: - Thread-safe bolus progress state
+
+/// Isolates the mutable `DoseProgressReporter` reference to prevent data
+/// races between Combine sinks (on `processQueue`) and async contexts
+/// (e.g. `cancelBolus`)
+private actor BolusProgressState {
+    private var reporter: DoseProgressReporter?
+    private var generation: UInt64 = 0
+
+    func setReporter(_ newReporter: DoseProgressReporter?, observer: any DoseProgressObserver) {
+        generation &+= 1
+        reporter?.removeObserver(observer)
+        reporter = newReporter
+        reporter?.addObserver(observer)
+    }
+
+    func clear(observer: any DoseProgressObserver) -> UInt64 {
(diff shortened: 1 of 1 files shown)


kingst (commented):
This is a real bug that we should fix, but I think an actor with generation tokens is too much complexity here. Here's my suggestion:
- Put the `cancelBolus` mutation on the `processQueue` to eliminate the race
- Drop the whole 500ms delay. I tested it in the simulator without the delay and the UX is fine, I think we should drop it.

If we really do want the 500ms delay, we should implement that at the UI layer, not in the bolus reporter, IMO.

marv-out:
After re-thinking that I think you're right. The actor approach was probably too complex for this situation. I've removed that and put the cancelBolus method on the processQueue.

marv-out:
Are you okay with my latest changes @kingst ?

kingst (approved):
LGTM, nice update

PR #1238, opened 2026-06-25
fix(treatments): keep focused input visible above the keyboard

### LINKED ISSUE #1115: Change treatments UI default behaviour/scroll so you can always see what you're entering

I'm torn between bug and feature request, but opted for the latter. Whilst typing #1114, I realised the treatments UI keyboard placement can potentially be dangerous. Mistyping something, whether due to lag or just fat-fingering it or whatever the reason is, may go unnoticed by the user, especially when they're a bit rushed.

When its of sufficient difference to exceed maxBolus: no problem, it'll block. But when for example maxBolus is 15 and you accidentally input (or it accidentally registers) "11" instead of "1": this will go through and give you 1000% more insulin than you intended if you don't see your mistake.

The bolus entry field should always be visible when the keyboard is on screen. It must preferably never fall behind the keyboard and be hidden by it, but at the very least the keyboard shouldn't go on top of it by default. 

It currently looks like this by default (on an iPhone 16 Pro) when tapping the bolus field: 

image

As you can see the bolus field is obscured by the keyboard - so during typing you don't see what got put in there. But imho, you *must* be able to see what you're typing at all times to reduce chances of missing a mistake. This may be able to get fixed by simply making the UI "scroll up" a bit when the keyboard appears. You see, after conjuring the keyboard you can manually scroll up to reveal the field so you can see what you're typing: 

image

I think, for safety, it should scroll up like that by itself so the user can always see what they've entered - which reduces mistakes/unnoticed higher entries than intended. 

There may be different ways to solve this in the UI, but I don't know of them. 

## User Impact

User will always see the bolus field, so the user can watch what they're entering without having to manually scroll after conjuring the keyboard. This increases the chance a user notices an error/typo.

Sjoerd-Bo3:
The treatments entry list relies on the system's default keyboard avoidance to keep the focused field visible. This is unreliable: the list height changes with user settings (e.g. the Reduced Bolus / Super Bolus toggles add a row), and on some devices the bolus field ends up hidden behind the keyboard. A user can then type "11" instead of "1" without seeing it, which is a safety concern.

Wrap the list in a ScrollViewReader and explicitly scroll the focused input row into view (centered above the keyboard) whenever focus changes, so the user can always see what they are entering regardless of which optional rows are shown.

Fixes #1115

### DIFF
--- a/Trio/Sources/Modules/Treatments/View/TreatmentsRootView.swift
+++ b/Trio/Sources/Modules/Treatments/View/TreatmentsRootView.swift
@@ -202,190 +202,225 @@ extension Treatments {
             }
         }
 
+        /// Scrolls the currently focused input row so it stays visible above the on-screen keyboard.
+        ///
+        /// The treatments list grows or shrinks depending on user settings (e.g. the Reduced Bolus /
+        /// Super Bolus toggles add an extra row), so relying on the system's default keyboard avoidance
+        /// is unreliable: on some devices/configurations the bolus field ends up hidden behind the
+        /// keyboard. Explicitly scrolling the focused row into view guarantees the user can always see
+        /// what they are entering, regardless of which optional rows are shown. See issue #1115.
+        ///
+        /// - Parameters:
+        ///   - field: The field that just gained focus, or `nil` when focus was cleared.
+        ///   - proxy: The enclosing `ScrollViewReader`'s proxy used to perform the scroll.
+        private func scrollFocusedFieldIntoView(_ field: FocusedField?, proxy: ScrollViewProxy) {
+            guard let field else { return }
+
+            // Fat and Protein share a single row, which is tagged with `.fat`.
+            let targetID: FocusedField = (field == .protein) ? .fat : field
+
+            // Delay slightly so the keyboard's presentation animation has begun and the List has
+            // resized its scrollable area before we scroll, otherwise the target can land back
+            // underneath the keyboard.
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.3) {
+                withAnimation {
+                    proxy.scrollTo(targetID, anchor: .center)
+                }
+            }
+        }
+
         var body: some View {
             ZStack(alignment: .center) {
                 VStack {
-                    List {
-                        Section {
-                            ForecastChart(state: state)
-                                .padding(.vertical)
-                        }.listRowBackground(Color.chart)
+                    ScrollViewReader { proxy in
+                        List {
+                            Section {
+                                ForecastChart(state: state)
+                                    .padding(.vertical)
+                            }.listRowBackground(Color.chart)
 
-                        Section {
-                            carbsTextField()
+                            Section {
+                                carbsTextField()
+                                    .id(FocusedField.carbs)
 
-                            if state.useFPUconversion {
-                                proteinAndFat()
+                                if state.useFPUconversion {
+                                    proteinAndFat()
+                                        .id(FocusedField.fat)
 
-                                if showFatProteinOrderBanner {
+                                    if showFatProteinOrderBanner {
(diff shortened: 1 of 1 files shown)


Sjoerd-Bo3:
@LiroyvH olease stamp this with a approval if it works

EJM0:
While the issue was also present in the local iPhone 16 Pro SIM, defining a separate .safeAreaPadding(.bottom: 50) for the List at line 398 seems to fix this. Also, this elevates the text input a bit further from the keyboard close button. Seems to be device consistent, see screenshots of iPhone 16 Pro SIM and iPhone 17 Pro.
Maybe the padding could be increased even more to circumvent miss clicks on the close button of the keyboard.

I only tested this for the 16 and 17 pro so i don't know if that fix is also consistent on smaller iPhone models, here maybe the original fix of @Sjoerd-Bo3 would be better.
But i think adding the specific .safeAreaPadding triggers iOS to definitely keep the safe area for focused fields....

16:
image
17:
image

Sjoerd-Bo3:
Defining a pixelike area is tricky when dealing with a lot of devices and aspect ratios. It will not work somewhere. So thats why I went the builtin way instead of that. I agree, yours looks cleaner on your device, but that may vary unfortunately...

EJM0:
Okay good to know. That's where I did not knew, if these type of areas are consistent for different devices.Considering that using a scrollview is way more consistent. 👍

Sjoerd-Bo3:
Could be, but you are hand defining that 50px, so that probably doesnt scale i GUESS..

LiroyvH:
@EJM0 With it scrolled down as like in your screenshot, you cannot see the warning text for dangerous low glucose as its dropped behind the keyboard. Enter a high amount of units (without carbs to compensate) to see what I mean.

dnzxy:
Since this is entirely vibed, before I spend time reviewing it:
@LiroyvH @EJM0 does this work?

LiroyvH:
@dnzxy 
> Since this is entirely vibed, before I spend time reviewing it: @LiroyvH @EJM0 does this work?

The first version (cfe02c5 or a42620e depending on where you look) works, yup - have been using it ever since (also in latest dev, 112fe88). Screenshots are here: https://github.com/nightscout/Trio/issues/1115#issuecomment-4804198883

The second version (I think 810da2a) didn't work and actually made matters worse. I see now there appears to be another modification shown here in this PR, 5add693, that I have not tested and I'm not sure which iteration (1st or 2nd) it augments. Seems to be a minor modification but I'm unsure to what and which version.

Unfortunately I'm not quite able to tell which version actually ended up being submitted for PR now... Sorry.  Hopefully @Sjoerd-Bo3 can clarify that.

Sorry for being vague but I'm not entirely sure which one has been submitted. if it's the first version then I can confirm it works fine. Any other version: I'd have to testdrive it.

dnzxy:
Thanks for adding some context @LiroyvH .
@Sjoerd-Bo3 please clean up the PR and then let's get it tested one more round.

Sjoerd-Bo3:
Version 1/3 were the same, cleaned up the PR and removed commit 2/3. So we are back where @LiroyvH and I tested for 2 weeks.

LiroyvH:
@Sjoerd-Bo3 @dnzxy It would seem the latest dev version fails to build with these changes.
Scherm­afbeelding 2026-07-12 om 23 43 56

To exclude this as the culprit: 
- Cloned a fresh copy of Trio dev and made no modifications: build succeeds.
- Clean build folder, close XCode, replace TreatmentsRootView.swift with the version from this PR and go again: build fails again. 

It built fine before this, but I cannot get it to build this at all anymore with the latest dev version.

trioneer-dev:
Closing in favor of #1365 .

PR #1243, opened 2026-06-26
Move "Snooze All" to Settings and Clean up Legacy Alerting; Follow-Up to #1203 – Prepare O5 and Libre 3/3+ Integration [3/3]

trioneer-dev:
NEEDS TO BE MERGED AFTER #1203 LANDS IN `feat/o5` BRANCH.

## Summary

- **Snooze All** moves from a hidden long-press on the home glucose bobble to a prominent button in **Settings → Notifications**. Tapping it opens an inline sheet with the four canonical durations (15 min / 1 hr / 3 hrs / 6 hrs), each calling `TrioAlertManager.applySnooze(for:)` directly.
- **Snooze module deleted.** No more module / router round-trip, no obsolete `GlucoseAlarm` enum, no stray 20 min option.
- **Legacy SwiftMessages alert pipeline removed.** `Router.alertMessage`, `MessageContent`, `MessageType`, `MessageSubtype`, `NotificationAction`, `alertMessageNotificationObserver`, `MainStateModel`'s SwiftMessages dispatch, and the dead `.pumpConfig` UN-tap deep-link are all gone. `SwiftMessages` SwiftPM dependency dropped.
- **4 real callers migrated** to `TrioAlertManager`:
  - Calibration: "Glucose is stale for calibration"
  - GarminManager: "Garmin Connect must be installed"
  - CGM Settings (×2): "Unable to open the app" / "No URL available"
- **`Logger.info()`** is now log-only — matches `warning` / `debug`. Banner side-effect removed.
- **`TimeInterval` convenience helpers** (`init(hours:)`, `init(minutes:)`, getters, statics) extracted from the deleted Snooze module into `Trio/Sources/Helpers/TimeInterval+Convenience.swift`, since callers outside the alert pipeline rely on them.

Net effect: every user-facing alert in the app flows through `TrioAlertManager`. No more parallel SwiftMessages path.

Follow-up to PR #1203.


Trio/Sources/Modules/Home/View/HomeRootView.swift
```diff
@@ -133,24 +133,15 @@ extension Home {
...
-                cgmStatus: state.cgmDisplayState,
-                cgmSensorExpiresAt: state.cgmSensorExpiresAt,
-                cgmWarmupEndsAt: state.cgmWarmupEndsAt
-            )
+                glucose: state.latestTwoGlucoseValues
+            ).scaleEffect(0.9)
                 .onTapGesture {
                     if !state.cgmAvailable {
                         showCGMSelection.toggle()
                     } else {
```
jwoglom: So this will remove a secret long press on the cgm details to snooze, is the suggested alternative to go to the settings page? That seems worse, why not keep both

Trio/Sources/Modules/Settings/View/Subviews/NotificationsView.swift
```diff
@@ -33,58 +35,50 @@ struct NotificationsView: BaseView {
     var body: some View {
         List {
             Section(
-                header: Text("Manage iOS Preferences"),
+                header: Text("Mute Trio Alarms"),
                 content: {
```
jwoglom, line 38: It seems off to me to perform a one time operation alongside the settings for alerts themselves. Like, if I wanted to configure how a snooze would work, it would make sense to go to preferences, but if I just want to snooze what alerts exist right now, I'm not going to go into preferences

Trio/Sources/Modules/Home/View/HomeRootView.swift
dnzxy: It removes a secret long-press on what we call the "glucose bobble", essentially the big centered element on the home screen. We _can_ keep it, @MikePlante1 has also noted this, but as you say, it was very much secret, and not only not documented anywhere, but for a long time also not really properly functional (got fixed as recently as #807 ).
I'm open to keeping it, if we truly thing it should be kept, but do vote in favor of not doing it via the router and UIKit mainModel view modifier but just my calling it in a sheet, like we now do in notification settings view. 

Another idea I had contemplated was to add a 🔔 (and when snoozed 🔕) somewhere on home view, that would toggle this — just _really_ not sure where to best place that. (I digress).

Trio/Sources/Modules/Settings/View/Subviews/NotificationsView.swift
dnzxy: You are doing that for example in the official Dexcom G7 app if I am not completely mistaken, and the "Mute All" button for DIY Loop is also situated in a similar context. It would add some cohesiveness across systems. 
Alternative ideas?

Trio/Sources/Modules/Home/View/HomeRootView.swift
Sjoerd-Bo3: I would like to access snoozing somewhere from the homescreen. Have to agree with @dnzxy that having secret things is kinda weird. But removing the function entirely from the homescreen is also not the solution. Maybe a snooze button could be the solution indeed.

jwoglom:
Did we add back the long press?

jwoglom (approved):
I updated my build to include the latest 1243 commits yesterday, and have been running prior versions of the PR for several days before that. No concerns here.

dnzxy (approved):

PR #1248, opened 2026-06-29
Add a Quick Bolus feature

t1dude:
Inspired by @bjorkert 's work on Loop Follow, I have added a Quick Bolus feature.

**Why?**
Users who pre-bolus and/or do not register carbs still have to go through the Treatments sheet, tap the bolus field, add amount, and confirm. When you "just want to bolus", the Treatments sheet is a bit complex.

**How does it work?**
Tap and hold the + button to open the Quick Bolus feature. It will look at your bolus history for the last 90 days and suggest up to 5 different amounts that you usually give at the same time-of-day. It will distinguish between weekdays and weekends, and give less weight to older bolus data.

**Screenshots**
If you have no bolus history, it will let you know:

Simulator Screenshot - iPhone 17 Pro Max - 2026-06-29 at 08 50 46

Bolus history of 1U and 2U:

Simulator Screenshot - iPhone 17 Pro Max - 2026-06-29 at 08 54 26

Quick Bolus suggests 1U and 2U:

Simulator Screenshot - iPhone 17 Pro Max - 2026-06-29 at 08 54 34

More historic values:

Simulator Screenshot - iPhone 17 Pro Max - 2026-06-29 at 08 59 02

More suggestions:

Simulator Screenshot - iPhone 17 Pro Max - 2026-06-29 at 08 59 13

After selecting a value, you are asked to confirm. Biometric control is applied after confirmation.

Simulator Screenshot - iPhone 17 Pro Max - 2026-06-29 at 08 54 58

Information/help sheet:

Simulator Screenshot - iPhone 17 Pro Max - 2026-06-29 at 08 54 47


Sjoerd-Bo3:
I like the idea a lot as a FCL-er. i will try it in my build and report back!

Sjoerd-Bo3:
Maybe this should live under a flag in settings like superbolus etc? I dunno if everyone is going to use it.

t1dude:
> Maybe this should live under a flag in settings like superbolus etc? I dunno if everyone is going to use it.

I’m okay with adding a flag if that’s what the team wants :) On the other hand it’s not like this feature is in the way of anything else. The way it is implemented now you’d have to know that it’s there.

Sjoerd-Bo3:
> > Maybe this should live under a flag in settings like superbolus etc? I dunno if everyone is going to use it.
> 
> 
> 
> I’m okay with adding a flag if that’s what the team wants :) On the other hand it’s not like this feature is in the way of anything else. The way it is implemented now you’d have to know that it’s there. 

Unsure, was just thinking aloud. As i would like it in there. But I can feel that for default use a hidden action after a long press is not nice default ux

dnzxy:
Really like the general proposal. After all, it came from a nifty feature you had for your personal build and some conversation around it potentially adopting Jonas' great feature from Loop Follow. 👏 

Just a few comments on the UI:
- Instead of using a `List` with `Section`s as rows, potentially show the recent 3 boluses suggestions in capsule / pill shapes in a single `HStack` – saves space and makes it look a bit more "Trio"-esque (I think)
- The wording seems slightly odd around the confirmation dialog. In TreatmentView, we use "Enact Bolus", never "Deliver".
- The confirmation dialog uses a `.destructive` role for the "Deliver" button which renders the button in a red tint. This sets the wrong "intent" here for that button.
- Typically, all our "help sheets" in Trio come with a "Got It!" button in the bottom of the sheet, not a `toolbarItem` button labeled "Done".
- While I really love the overall idea, I find the workflow to be slightly odd from a UX perspective. I am not certain if the following is a _good_ idea, but want to bring it up for discussion:
   - Currently, we have a half-sheet open, then the user has to select a choice, which opens a confirmation dialog, which the user then confirms, which then prompts faceId / touchId
   - How about we have a single (half) sheet, with 3 pills for the 3 suggestions, and on the bottom of the sheet there's a slide-to-confirm control (similar to the "Insert Cannula" control for Omnipod pods), which is an _intentional_ enough gesture to potentially even omit faceId / touchId? 
   - That way, it would be long-press, tap, slide in a _single_ view, instead of long-press, tap, tap in subsequent views, plus look-at-phone / hold-finger-over-homebutton.
   - We will probably have to discuss whether the slider is sufficient to omit the authentication. 
   - Edit to add a little mockup (courtesy of Claude based on the above description):
   - image

From a code perspective, our `HomeRootView` is already enormous. Please refactor your proposed additions so that they new view elements live in small fragment files (either proper view `struct`s or potentially just as extensions to `HomeRootView`), and try to keep it as D-R-Y as possible. 

As for Sjoerds suggestion for a toggle setting for this proposal: 
- I agree, it will not be in the way and just be a "secret" element.
- The fact it is "secret" though, may actually speak _for_ adding it as a setting.
- Having a setting – we don't already have enough settings anyways 🙃 – gives us the opportunity to provide more explanations around how this works (the sheet you show when there is not enough data), and the user to "find" this feature and "learn" about it by browsing Trio. 
- Also, we could make it so that if the user is not enabling this setting, we do not constantly fetch data and perform the scoring in `HomeStateModel` every time there are updates to data.

Overall, really think this as a great feature! Thank you for the invested time @t1dude

Sjoerd-Bo3:
@dnzxy I agree we have too many settings🤣 but sometimes its allowed😅

I like your UX idea! I found it also a bit clunky when using. The slider could be a good solution. i'm all for your suggestions.

t1dude:
Thanks for a thorough review and good suggestions @dnzxy ! I like the mockup, I’ll see what I can do. 

I still think we should keep FaceID as a safety net to avoid the unlikely edge case of someone giving a bolus to someone else.

I was trying to avoid adding yet another setting, but it makes sense as a way to introduce the feature to users who do not know about it.

Refactoring this later today.

t1dude:
Refactored based @dnzxy 's great feedback. Here's a video showing how it works on my test phone (iPhone 13 Mini)

https://github.com/user-attachments/assets/72cbbea5-8329-4ed5-97f5-d69d9aa4173c


Let's discuss if the FaceID part is necessary with the introduction of swipe-to-enact.

AndreasStokholm:
This looks great! 👏

Just wanted to add my 2 cents on the slider warranting not having the authentication, which IMO is a no.

There is a difference between acknowledging that you want to take an action (sliding the slider) and the app verifying you are allowed to take that action (FaceID).

We have FaceID after the normal flow, and IMO this is no different from a safety perspective.

Sjoerd-Bo3:
Refactor looks great! Imo having the slider and faceid both is pretty great! Seems like it works seamless.

Sjoerd-Bo3:
@t1dude great already. First comment: I would like a haptic feedback when i long press invoke the quick bolus like the force looping also has.

dnzxy:
Generally speaking, great! I have a few nits. 

I suggest to combine the sections top 3 setting navigation rows under a single navigation row "Treatments".
image

- Help sheet button (the ℹ️ icon) should be a `(?)` icon, to be consistent with all other views.
- Justified text is very hard to read, especially for low-literacy and dyslexic users. Let's please keep it left-aligned.
- From a "view hierarchy" point, the space between the various shapes and texts is a bit too big
   - Slightly increase the sheet title font size
   - Reduce space below sheet title
   - Reduce space below slider button
   - Increase space between hint text and slider button
   - Increase the space between the unit recommendation shapes
   - Slightly decrease the horizontal padding of the recommendation shapes. Ideally, the relation between vertical and horizontal padding is 1:2
   - Make the unit label the same size, feel free to keep it `.secondary`
   - When there are only 1 or 2 suggestions, I'd suggest to give the shape some form of max-width maybe? 
   - Sheets typically come with a presentation drag indicator. It's missing here for both the help sheet and the quick bolus sheet itself.

Untitled

As for the slide button and subsequent authentication:
- It is the de-facto standard as per IEC standards that 2 intentional action that trigger a high risk action. That'd mean after selecting the dose, it's 1. tap a button (or use the slider), 2. bio authenticate via FaceID / ToucID
- The slider may add friction. It's not present for the normal treatment view and there are users that like it, and those that dislike it. We can get it into wider user testing, but we may (have to) end up replacing the slider for a normal "Enact Bolus" button

I haven't done a thorough code review yet, mostly UI/UX.

Overall great feature.

Sjoerd-Bo3:
@dnzxy thoughts about haptic feedback after longpress? Like we have for force loop? 

Also feels like to me that its a standard apple behaviour for long presses.

claude says:
> Short answer: it’s a strong iOS convention, not a formal HIG “rule.” Here’s the accurate picture:
What Apple actually says
	•	Apple’s HIG treats haptics as part of feedback in general — “every action should have a reaction,” haptics should be “subtle and meaningful — never constant or annoying,” and you should pair haptics with visual/other feedback rather than rely on them alone, and stay consistent with system-provided haptics. It does not contain a line mandating “a long press must fire a haptic.”
	•	Where a “rule” effectively exists is that iOS itself plays the haptic for you when a system long-press affordance triggers — i.e. a UIContextMenuInteraction / SwiftUI .contextMenu firing on long-press produces the standard “thunk” automatically. Apple’s guidance is to respect standard gestures and their haptic patterns and not override them.

t1dude:
> @dnzxy thoughts about haptic feedback after longpress? Like we have for force loop?
> 

I’m looking into it.

t1dude:
@Sjoerd-Bo3 see latest commit

t1dude:
@dnzxy Thanks for the detailed feedback!

> I suggest to combine the sections top 3 setting navigation rows under a single navigation row "Treatments".

When opening "Treatments", should we have 3 submenus for each row, or just one loooong section with all the settings from Bolus Calculator, Quick Bolus and Meal Settings?

dnzxy:
@Sjoerd-Bo3 Sure. I haven’t addressed it as I didn’t read through the issue but just built the status quo and went from there typing my notes. 

@t1dude three submenus. Make sure to adjust all the search item paths accordingly 👀

Sjoerd-Bo3:
> @Sjoerd-Bo3 see latest commit

Works great thanks!

MikePlante1:
Should we pivot to a different name since “Quick Bolus” is what Mobi calls the button on their pump and it might confuse users having two features called the same thing?

dnzxy:
I‘d argue that the Mobi name is semantically wrong and we‘re quite literally doing a quick bolus here, but how about Smart Bolus?

t1dude:
Thanks @MikePlante1 ! I don't worry too much about what Mobi does, but I agree we can discuss Quick Bolus vs Smart Bolus.

bjorkert:
LoopFollow uses Quick-Pick [Boluses]. Naming things is the toughest, just wanted to stay away from anything saying that the bolus itself is good or recommended in any way.. it just a faster way to bolus something you bolused before.

image

t1dude:
> Generally speaking, great! I have a few nits.
>
> "All kinds of feedback."
> 
> I haven't done a thorough code review yet, mostly UI/UX.
> 
> Overall great feature.

@dnzxy I have addressed all of this in recent commits. Please re-build and let me know if I need to tweak any further. Happy to keep adjusting to get this right.

t1dude:
> LoopFollow uses Quick-Pick [Boluses]. Naming things is the toughest, just wanted to stay away from anything saying that the bolus itself is good or recommended in any way.. it just a faster way to bolus something you bolused before.
> 

Thanks, that is why I would like to keep any "recommendations" or "smartness" out of this. The feature is just enabling users to give a Quick Bolus. We can do a poll if we think it is necessary, or just ask a select group of experienced users what they think.

MikePlante1:
I think “Quick-Pick Boluses” would be great and the consistency between the two apps would be nice.

t1dude:
> I think “Quick-Pick Boluses” would be great and the consistency between the two apps would be nice.

Let’s do that!

dnzxy:
Made a few small adjustments after discussing them with @t1dude around spacing. 
Looks as follows now:

| ❌ Before Changes | ✅ 1 Bolus in History | ✅ 2 Entries | ✅ 3 or more |

t1dude:
> Made a few small adjustments after discussing them with @t1dude around spacing. Looks as follows now:
> 

Thanks @dnzxy ! :) I think we are ready for a code review?

dnzxy (commented):
I think this is good, code-wise. 

Please add the new setting to the setting export processing, so it shows up, and adjust the export for the now moved views (meal settings, bolus calculator).

trioneer-dev (approved):

PR #1251, opened 2026-06-29
upgrade looping animation and remove looping text

EJM0:
This updates the looping indicator form using the default progress spinner, to being the progress spinner.
Also it disables the localisation for the looping text if loop is active ... see comment.

spinner in action 1. longer loop 4s ,  2. shorter loop 1s

https://github.com/user-attachments/assets/be6f1ca3-bda4-4ac7-8814-7a6819b686fa


t1dude:
I’m not sure I agree on the translation block for «looping». 

Animation is cool 😎

EJM0:
sadly without locking "looping" the updatePerimeter is a bit messedup and on resizing, the spinner stutters a bit and provokes lagging.

emulated 1s loop

https://github.com/user-attachments/assets/261e739a-2a38-430c-af60-6892130fc8b6

P.S. I think I misunderstood. 😅
I reckon blocking localisation could be a good way to tidy up the UI for languages with longer words. Or, we could use looping in the translation as a helpful shortener for languages that might otherwise have translations that are too big.

dnzxy:
Moving this to draft. Will wire this in when #1332 lands in `feat/home-refactor`.

bjornoleh:
So there's still time to argue a bit more about the no -localisation of "Looping" 😂

Do we need any wording there at all, with the thing looping around as an animation? Most other strings like this are being replaced with symbols anyways?

dnzxy:
> Most other strings like this are being replaced with symbols anyways?

Huh?

bjornoleh:
We used to have IOB and COB spelled out, and I think some other strings are being done away with in the upcoming refactored UI, such as "i" replacing "i Info".

EJM0:
so sould i change the code to skip the displacement of the looping text, and instead just show the animation of the border and the icon? @bjornoleh

EJM0:
Also a thing to overlook is the display of manual mode, quite nicely translated to "Handbuch“ ( Handbook) in German. I’d suggest only changing the colour/ icon and displaying  the remaining time, how long the manual basal rate is scheduled for.

| example |
| --- |
| image |

bjornoleh:
@EJM0 , this isn't up to me, but I suspect that some understandable icons or symbols could potentially do a better job indeed. 

Cute German translation of manual mode! It nicely shows how little gamification is needed to make people spend their time to speedrun translations by only clicking on the first auto-translate suggestion without thinking. And Crowdin has very little in the way of gamification in their design 😂

bjornoleh:
But first of all, we shouldn't have to duplicate the info about manual basal both in the pump info and in the loop status.

EJM0:
Instead of using the ProgressView's spinner as a persistent loop indicator, the newly added SpinnerCapsule component is now used to show loop progress on the home screen, while the spinner within the pulldown persists for as long as the loop is active.

https://github.com/user-attachments/assets/efacc06a-7a29-4523-8d38-98d6e27cef9b

EJM0:
Finished up to review.

AndreasStokholm (approved):
Tested and reviewed. Looks good, and the animation is very cool! 👏

PR #594, opened 2025-05-27
feat: Enhanced insulin type visualization and filtering in treatment history

Sjoerd-Bo3:
## Summary
- Added treatment type filtering with multi-select checkboxes
- Filter menu allows viewing specific treatment types
- Fixes #592

## Problem / Solution
**Problem**: The treatment history list is cluttered with temp basal and SMB entries, making it difficult to find specific treatment types like manual boluses.

**Solution**: 
- Added filter button with dropdown menu containing checkboxes
- Users can select multiple treatment types to display (OR logic)
- Filter categories: Bolus, SMB, External, Temp Basal, Pump State
- Icon changes color when filters are active

## Test plan
- [x] Verify checkbox filtering works for each treatment type
- [x] Test multiple selections show combined results
- [x] Confirm empty selection shows no treatments
- [x] Check that filter state is visually indicated
- [ ] Test in both light and dark modes
- [ ] Verify no regression in treatment deletion/swipe actions

### Filter Implementation
| Filter Menu with Checkboxes | Multiple Types Selected | Filtered View |
|---|---|---|
| [Add screenshot] | [Add screenshot] | [Add screenshot] |

| Bolus Filter | SMB Filter | External Filter |
|---|---|---|
| [Add screenshot] | [Add screenshot] | [Add screenshot] |

Issue #102, opened 2024-04-14
Inconsistent ExternalInsulin type enum and isExternalInsulin attribute


avouspierre:
With the #93 the model PumpHistoryEvent is modified to add external insulin Type and a isExternalInsulin boolean : 

| struct | type |
|--------|--------|
| ![CleanShot 2024-04-14 at 11 20 24](https://github.com/nightscout/Open-iAPS/assets/4339604/661312e6-6464-47f8-a663-c667c746a21b) | ![CleanShot 2024-04-14 at 11 20 54](https://github.com/nightscout/Open-iAPS/assets/4339604/477f2393-1800-41e2-9495-673363790f1b) | 

However, when a external insulin is added with the interface, only the externalInsulin bool is defined to true (the type = bolus and no externalInsulin) : 
![CleanShot 2024-04-14 at 11 24 50](https://github.com/nightscout/Open-iAPS/assets/4339604/0a728382-f8f1-4c04-90d8-e8a8bb71fb19)

I also found a function never used by any code where isExternalInsulin = true 
![CleanShot 2024-04-14 at 11 22 52](https://github.com/nightscout/Open-iAPS/assets/4339604/64f8d4d2-4592-45b7-bec6-93aa569589bc)

The externalInsulinType is used by NS (and only by NS) to build a nightscoutTreatement object with a function to deduce the type : https://github.com/nightscout/Open-iAPS/blob/8ebd52a69612977cf705c5843d71250eb23bb009/FreeAPS/Sources/APS/Storage/PumpHistoryStorage.swift#L260 

My proposal will be to remove the externalInsulin Type in Event Type and create a specific type enum for NS, avoid to have inconsistent between type and isExtrernalInsulin attribue.

PR #1252, opened 2026-07-01
Tag bolus origin and show it in Nightscout

bjorkert:
Tags remote, watch, manual and shortcut boluses with their origin using the LoopKit bolusReference, and shows it in Nightscout as a plain note plus a machine-readable bolusOrigin field. SMB is left as is since it already shows as SMB.

Needs LoopKit/LoopKit#594 and loopandlearn/OmnipodKit#102.

While building/testing, the LoopKit and OmnipodKit submodules point at bjorkert branches (the #594 and #102 branches). Repoint them at merged loopandlearn commits before merging.

The Loop side uses the same shared kit changes: LoopKit/Loop#2455, LoopKit/NightscoutKit#3, LoopKit/NightscoutService#27.


dnzxy:
Moving this to draft, as there's discussion with other contributors around the upstream changes.
Very excited for this to land in Trio and the OS-AID landscape on iOS 🫶

PR #1254, opened 2026-07-01
Remove core data fetch limits (already time predicate limited)

dnzxy:
We had a few fetch limits in place that were potentially leading to fetch results, especially for pump history.

E.g. `getPumpHistory()` uses `pumpHistoryLast24h` but `fetchLimit: 288`.  With TBRs every 5-15 min (96-288/day) plus SMBs and manual user-initiated boluses, there are well over 288 events per day. This could potentially lead to the most‑recent 288 fetch result truncating pump eventing, covering less than 24h.

Similar for fetching glucose, where the predicate did not exclude potential manual glucose entries, leading to a false assumption of one reading per 5 minutes (288(day).

### DIFF
--- a/Trio/Sources/Modules/Home/View/HomeRootView.swift
+++ b/Trio/Sources/Modules/Home/View/HomeRootView.swift
@@ -139,18 +139,18 @@ extension Home {
                 cgmSensorExpiresAt: state.cgmSensorExpiresAt,
                 cgmWarmupEndsAt: state.cgmWarmupEndsAt
             )
-                .onTapGesture {
-                    if !state.cgmAvailable {
-                        showCGMSelection.toggle()
-                    } else {
-                        state.shouldDisplayCGMSetupSheet.toggle()
-                    }
+            .onTapGesture {
+                if !state.cgmAvailable {
+                    showCGMSelection.toggle()
+                } else {
+                    state.shouldDisplayCGMSetupSheet.toggle()
                 }
-                .onLongPressGesture {
-                    let impactHeavy = UIImpactFeedbackGenerator(style: .heavy)
-                    impactHeavy.impactOccurred()
-                    state.showModal(for: .snooze)
-                }
+            }
+            .onLongPressGesture {
+                let impactHeavy = UIImpactFeedbackGenerator(style: .heavy)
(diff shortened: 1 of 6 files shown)


Trio/Sources/Modules/Home/View/HomeRootView.swift
```diff
@@ -139,18 +139,18 @@ extension Home {
...
-                    impactHeavy.impactOccurred()
-                    state.showModal(for: .snooze)
+            .onTapGesture {
+                if !state.cgmAvailable {
+                    showCGMSelection.toggle()
+                } else {
+                    state.shouldDisplayCGMSetupSheet.toggle()
                 }
+            }
+            .onLongPressGesture {
```
kingst, line 149: Do you have context on this change? I'm not sure if this is related to this PR...

Trio/Sources/Services/BolusCalculator/BolusCalculationManager.swift
```diff
@@ -196,8 +196,7 @@ final class BaseBolusCalculationManager: BolusCalculationManager, Injectable {
             onContext: glucoseFetchContext,
             predicate: NSPredicate.glucose,
```
kingst, line 197: Not really related to this PR, but we should consider removing this predicate and being explicit, e.g., `.glucoseFor24Hours` instead of `glucose`

kingst (commented):
The code looks great, I am very much in favor of this change. Just one question and one very optional comment that you are welcome to ignore.

kingst (approved):
Looks like the unrelated change was just linting, confirmed that the logic is the same. LGTM

Trio/Sources/Modules/Home/View/HomeRootView.swift
dnzxy: Just a linting change from a previous merge conflict fix :)

Trio/Sources/Services/BolusCalculator/BolusCalculationManager.swift
dnzxy: Yes, will push change this in another subsequent PR

PR #1255, opened 2026-07-01
Shorten Omnipod selection string in Onboarding to avoid break

dnzxy:
Change `All Omnipod Types` to `Omnipod` for the Onboarding selection to avoid UI breakage.
We don't specifically list the Medtrum and different RS models for Dana, or the different reservoir-size models for Medtrum.

image


kingst (approved):
String change only, LGTM

PR #1256, opened 2026-07-01
Bump OmnipodKit, fixes #1236

### LINKED ISSUE #1236: Omnipod error not registered during bolus
## Describe the bug
As well as some other users, I'm suffering from a lot of Omnipod Dash errors with error code 203. For me, this usually happens when I administer a bolus, but it's not limited to manual bolus injections as these errors occur with SMBs as well.
 
Previously, when this happened, the bolus progress would stop and Trio would notify me about the error via a notification. The pod status in the overview would then also be updated and of course the pod screams at me for as long as I didn't stop it.

This only started to happen after one of the most recent updates (to either v0.8.3, v0.8.2 or v0.8.1).
Previously the bolus progress would stop right away on a pod error and I also received the notification right away.

**This might also be especially dangerous, since Trio will most likely calculate the IOB based on what it thinks how far the bolus progressed instead of the actual value before the error.**

## Attach a Log
See attached file.

[log_prev.txt](https://github.com/user-attachments/files/29334004/log_prev.txt)
[log.txt](https://github.com/user-attachments/files/29334003/log.txt)

## To Reproduce
It's kinda hard to reproduce this behavior on purpose as it only happens when the pod reaches an error state.
However, what is always common:
1. Start a bolus
2. Pod reaches error state
3. Bolus keeps on progressing according to Trio

## Expected behavior
Bolus progress stops on a pod error and send notification immediately.

### Smartphone:
* Hardware: iPhone 17 Pro
* OS Version: iOS 26.5.1

### Pump:
* Manufacturer: Insulet
* Model: Omnipod Dash

### CGM:
* Device: *Dexcom G7
* Manager app: *Dexcom App

### Trio Version:
* Version Number: 0.8.3
* Repo: *nightscout/trio*
* Git Reference: 7b8db13

## Technical Details
See attached logs.

## Additional context
Nothing else I can share upfront. Feel free to reach out to me if you need anything else.

[log.txt](https://github.com/user-attachments/files/29334032/log.txt)

[log_prev.txt](https://github.com/user-attachments/files/29334048/log_prev.txt)

dnzxy:


marionbarker (approved):
Approve by code review and test.

Repeated the test cases of Bolus, Fault, Cancel bolus and Bolus, Fault. Both showed the improved response time to update status after the bolus was interrupted by the fault.

PR #1273, opened 2026-07-07
Use Swift Oref by default

kingst:
This change updates the oref swift setting to be an oref javascript
setting instead, effectively defaulting all users into the Swift
implementation.

By defaulting people into the Swift implementation we will get a
chance to find any lingering issues with the implementation before
removing the Javascript code from the codebase. If there are any
issues, people can switch back to JS without needing to recompile.

### DIFF
--- a/Trio/Sources/APS/APSManager.swift
+++ b/Trio/Sources/APS/APSManager.swift
@@ -184,7 +184,7 @@ final class BaseAPSManager: APSManager, Injectable {
             if wasParsed {
                 Task {
                     do {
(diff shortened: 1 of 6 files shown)


MikePlante1 (approved):
LGTM.
Code looks straightforward and good.
Built to a sim via Xcode 26.6 to confirm no linting.

t1dude (approved):
LGTM (also been running this for a month)

marv-out (approved):

PR #1284, opened 2026-07-10
[Part 1 of 6] Remove the Javascript oref implementation

kingst:
This commit removes the Javascript algorithm from the code
base. Changes include removing the JS resources, removing the setting
`useJavascriptOref`, and removing all swift code that uses or supports
running Javascript.

For now, we left the JSON marshalling in place but we will remove this
in subsequent commits / PRs.

Tested using the simulator.


dnzxy (approved):
LGTM 👏

PR #1285, opened 2026-07-10
Fix Spurious Pump Alerts

### LINKED ISSUE #1135: Dash Pod notifications coming out delayed
## Describe the bug
There are times when the Pod upcoming expiration notifications come out at a much later time sometimes at Pod expiration notice.  I suspect these are notifications that haven't been acknowledge or displayed previously but not positive on that as some of my acknowledgments are done pretty mindlessly.  In this latest instance I got notification of a low reservoir and the upcoming pod expiration notification was sent at the same time.  My Pod expiration reminder time is set to 24 hours and got the low reservoir message just before getting a 203 error.  Per 203 error design guidelines that typically comes out 2 hours prior to pod expiration:).

## Attach a Log
Screen shot  showing the two notifications and the fact that they came out at the same time will be attached if I don't mess it up.

## To Reproduce
Unsure as it doesn't always seem to happen but may have something to do with not 'acknowledging' the first pod expiration reminder.

## Expected behavior
Notification should come out at the expiration reminder time and not 22 hours later.

## Screenshots
Attached I think.

## Setup Information (please complete the following information):
Just setup an Omnipod with Trio and don't acknowledge the upcoming expiration reminder.

Image

### Smartphone:
* Hardware: iphone 14 Pro
* OS Version: iOS 26.4.1

### Pump:
* Manufacturer: Insulet
* Model: Omnipod Dash

### CGM:
* Device: Dexcom G7
* Manager app: Dexcom App

### Trio Version:
* Version Number: v0.7.0
* Repo: nightscout/trio
* Git Reference: Too lazy to look up right now but lmk if needed.

## Technical Details
As above nothing else I can think of.

## Additional context
Just everyday Trio use as far as I can think of.

dnzxy:
## Summary

The alerting rework routes pump-thrown alerts through the LoopKit protocol as intended. Tapping a notification calls acknowledgeAlert on the PM, no auto-ack anywhere. But it missed one spot: dismissing the in-app banner (and swiping away the UN) routed to snooze only, so the ack was never sent to the PM. With the app foregrounded, a pod alert could stay unacknowledged on the pod indefinitely unless the user also tapped the notification, leaving pods beeping on their repeat schedule (15 min expiration reminder, hourly low reservoir/expired) and alerts stuck as unacknowledged.

Dismissal is the user's response, so requestSnooze now also forwards the ack to the issuing device manager via acknowledgeOnDevice(), extracted from handleAcknowledgement. The forward is guarded on a registered AlertResponder – only device managers register (currently just the pump manager) – so Trio-internal alerts (glucose for Libre 3 and other CGms that require it, not-looping, algorithm) are unaffected, and PMs with no-op ack's (Medtrum, Dana, Minimed) see no behavior change.

User-facing behavior is unchanged: swipe/tap away still auto-snoozes for 15 min and dismisses both presentations, **but** pump alerts are now properly acknowledged at the PM layer (e.g. AcknowledgeAlerts reaches the pod and silences its beeps).

Fixes #1135.

### DIFF
--- a/Trio/Sources/Services/Alerts/TrioAlertManager.swift
+++ b/Trio/Sources/Services/Alerts/TrioAlertManager.swift
@@ -226,8 +226,18 @@ final class BaseTrioAlertManager: TrioAlertManager, Injectable {
         queue.async {
             self.liveAlerts.removeValue(forKey: identifier)
         }
-        let responder = queue.sync { responders[identifier.managerIdentifier]?.ref as? AlertResponder }
-        responder?.acknowledgeAlert(alertIdentifier: identifier.alertIdentifier) { error in
+        acknowledgeOnDevice(identifier: identifier)
+    }
+
+    /// Forwards the user's response to the issuing device manager — pump alerts
+    /// need this to also clear the alert on the device itself (e.g. pod beeps).
+    /// Only device managers register as responders (currently just the pump
+    /// manager), so Trio-internal alerts (glucose, not-looping, algorithm)
+    /// never match and bail out here.
+    private func acknowledgeOnDevice(identifier: Alert.Identifier) {
+        guard let responder = queue.sync(execute: { responders[identifier.managerIdentifier]?.ref as? AlertResponder })
+        else { return }
+        responder.acknowledgeAlert(alertIdentifier: identifier.alertIdentifier) { error in
             if let error = error {
                 debug(.service, "AlertManager ack failed for \(identifier.value): \(error)")
             }
@@ -364,6 +374,8 @@ final class BaseTrioAlertManager: TrioAlertManager, Injectable {
 
 extension BaseTrioAlertManager: TrioModalAlertResponder, TrioUserNotificationAlertResponder {
     func requestSnooze(identifier: Alert.Identifier, duration: TimeInterval) {
+        // Dismissal is the user's response — ack device-issued alerts at the PM layer too.
+        acknowledgeOnDevice(identifier: identifier)
         let untilDate = duration > 0 ? Date().addingTimeInterval(duration) : .distantPast
         if let glucoseType = GlucoseAlertType(slug: identifier.alertIdentifier) {
             broadcaster.notify(GlucoseSnoozeObserver.self, on: .main) { (observer: GlucoseSnoozeObserver) in


marionbarker:
## Test

Test started 2026-07-12 at 15:46 PDT. Will return to report results later.

### Configuration

Build fix-spurious-pump-alerts branch, commit e66bc689e, onto an SE 2nd gen phone running iOS 18.

### Test Narrative

2026-07-12 at 15:46 PDT
* Pair, prime and insert Omnipod 5 Pod, PH1K04182331, seq 0690007
   * Firmware, 6.0.0
   * BLE Firmware, 56.0.5
   * Card ID: -029C
   * Dexcom G6 only
* Configure expiration reminder 24 hours ahead
* Configure low reservoir reminder to 50 U

marionbarker:
### Updated Configuration

I checked with @itsmojo and we can modify the expiration reminder time to be more than 24 hours before the expiration.

I changed this line:
* https://github.com/loopandlearn/OmnipodKit/blob/main/OmnipodKit/OmnipodCommon/Pod.swift#L85

From:
`static let expirationReminderAlertMaxHoursBeforeExpiration = 24`
To:
`static let expirationReminderAlertMaxHoursBeforeExpiration = 72`

Rebuild the app onto the phone. Note that the code only allows you to modify the expiration reminder for the current pod to a time that is in the future.  So chose "Today at 17:46". It is currently 17:12.

Set a timer to remind me to return before the alert is expected.

marionbarker:
## Test Summary

✅ successful for pod expiration

### Test Narrative Continued

Trio app is in the foreground.
Phone is unlocked.
17:45: I hear the pump beep. (note it happened 3 times while waiting).

Wait for Trio alert (it's been 2 minutes since last loop so it may be a while).
Looping and I get the alert.
I touched it and heard the pod beep.
Go to pod diagnostics and check triggered alerts and they are all blank.

### Screenshot

This is what the alert looks like.

IMG_7283

marionbarker (approved):
I believe this should not be targeted to dev because this has the feat/o5 code in it. However, that is a decision for the devs to make.

I approve from test when directed to the correct branch.

dnzxy:
I‘ll change the OmnipodKit SHA. It will make its way back to feat/O5 through syncing dev in.

dnzxy:
@marionbarker changed the SHA to latest `OmnipodKit:main`. Thanks for the pointer :)

marionbarker (approved):
approve from code review and previous test. Thanks for updating the OmnipodKit SHA>

marionbarker (changes requested):
removing approval while I do more testing.

marionbarker:
I have a pod connected to Trio and I'm getting a repeated expiration alert that I cannot clear. Sorry - please use the Silent Pod trick to clear alerts that are not clearing.

marionbarker:
### Test Narrative

2026-07-13, PST time zone (UTC -7 hours).

I modified the Notifications to set the expiration alert to the next time slot available.
But when it alerted (at 11:47), I decided to tap on the pod icon to see if I could trigger the Trio notification and clear the pod alert sooner.  I never got the Trio notification.

After clearing the alert by setting it again to notify at 12:47, I rebuilt to make completely sure I have the right code and confirmed the code changes in this PR are in Xcode. (I also did a clean build folder and fresh build.)

The only difference between my build and this PR is I'm using the dev branch of OmnipodKit because I have an Omnipod 5 pod connected and I allow the expiration reminder to be set much sooner than is typical (72 hours instead of 24 hours).

The next alert is planned for Today 12:47 pm

This time - I plan to repeat what I did with my first test, which is do not disturb Trio and leave it open with the phone unlocked. I will confirm that I do get the Pod Expiration Reminder alert (see screenshot in prior comment).

> Note - I did check Trio notification style and it was on temporary banners instead of persistent. I changed it to persistent.

marionbarker:
## Report 1

The alert does appear when I don't mess with the phone.

### Next Test

Set alert to Today at 13:47, unplug the phone and lock it.

marionbarker:
## Test Status

I configured for more expiration reminder alerts, left the phone locked, waited for the pod to beep and then opened the Trio app and was always presented with the notification to acknowledge the alert.

I don't know why the first test I did today did not get an alert.

I'm uploading the logs from yesterday and today for later evaluation.

[Trio-PR1285_log.zip](https://github.com/user-attachments/files/29988546/Trio-PR1285_log.zip)

marionbarker (approved):
I am approving this. It worked just about every time. Maybe someone can figure out from the log what I did wrong the one time I could not get the banner to clear the alert.

I'm happy to do more testing tomorrow if desired but need to quit now. This is probably worth merging even if we need another update.

PR #1287, opened 2026-07-11
[Part 2 of 6] Remove JSON marshalling for makeProfile

kingst:
This commit removes the JSON marshalling for makeProfile. Note however that most of the data that makeProfile consumes is stored in JSON and its result is also stored in JSON, so this commit is largely about separation of responsibilities and making our algorithm functions use native types even though the JSON encoding / decoding is still happening.

We did get a minor win by removing double encoding / decoding for the Preferences struct and we will get big benefits later when we take on glucose / pump events carbs. But for now this change is mostly a refactor.

### DIFF
--- a/Trio/Sources/APS/OpenAPS/OpenAPS.swift
+++ b/Trio/Sources/APS/OpenAPS/OpenAPS.swift
@@ -625,17 +625,15 @@ final class OpenAPS {
         async let getCR = loadFileFromStorageAsync(name: Settings.carbRatios)
         async let getTempTargets = loadFileFromStorageAsync(name: Settings.tempTargets)
         async let getModel = loadFileFromStorageAsync(name: Settings.model)
-        async let getTrioSettingDefaults = loadFileFromStorageAsync(name: Trio.settings)
 
-        let (pumpSettings, bgTargets, basalProfile, isf, cr, tempTargets, model, trioSettings) = await (
+        let (pumpSettings, bgTargets, basalProfile, isf, cr, tempTargets, model) = await (
             getPumpSettings,
             getBGTargets,
             getBasalProfile,
             getISF,
             getCR,
             getTempTargets,
-            getModel,
-            getTrioSettingDefaults
+            getModel
         )
 
         // Retrieve user preferences, or set defaults if not available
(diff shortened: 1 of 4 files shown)


Trio/Sources/APS/OpenAPS/OpenAPS.swift
```diff
@@ -625,17 +625,15 @@ final class OpenAPS {
         async let getCR = loadFileFromStorageAsync(name: Settings.carbRatios)
         async let getTempTargets = loadFileFromStorageAsync(name: Settings.tempTargets)
         async let getModel = loadFileFromStorageAsync(name: Settings.model)
-        async let getTrioSettingDefaults = loadFileFromStorageAsync(name: Trio.settings)
 
-        let (pumpSettings, bgTargets, basalProfile, isf, cr, tempTargets, model, trioSettings) = await (
+        let (pumpSettings, bgTargets, basalProfile, isf, cr, tempTargets, model) = await (
```
dnzxy: Nit: you changed cr to carbRatio (singular) further down, but left it here as cr. Probably should be plural too. 

Also, not changed in this PR but just pointing out: if we use descriptive variable names it probably should be insulinSensitivities.

Trio/Sources/APS/OpenAPS/OpenAPS.swift
```diff
@@ -680,31 +678,37 @@ final class OpenAPS {
...
         let clock = Date()
         do {
-            let pumpProfile = try makeProfile(
+            // Decode the raw settings into native models. The bundled-defaults
+            // fallback still happens in loadFileFromStorageAsync above, so decoding
+            // here preserves the same behavior it previously had inside makeProfile.
+            let pumpSettings = try JSONBridge.pumpSettings(from: pumpSettings)
+            let bgTargets = try JSONBridge.bgTargets(from: bgTargets)
+            let basalProfile = try JSONBridge.basalProfile(from: basalProfile)
+            let isf = try JSONBridge.insulinSensitivities(from: isf)
```
dnzxy: Nit: same comment in ISF here (existed previously like this).

Trio/Sources/APS/OpenAPSSwift/OpenAPSSwift.swift
```diff
@@ -2,44 +2,27 @@ import Foundation
...
-        isf: JSON,
-        carbRatio: JSON,
-        tempTargets: JSON,
-        model: JSON,
-        trioSettings: JSON,
+        preferences: Preferences,
+        pumpSettings: PumpSettings,
+        bgTargets: BGTargets,
+        basalProfile: [BasalProfileEntry],
+        isf: InsulinSensitivities,
```
dnzxy: Nit: same here

Trio/Sources/APS/OpenAPSSwift/OpenAPSSwift.swift
```diff
@@ -2,44 +2,27 @@ import Foundation
...
-            return (try .success(JSONBridge.to(profile)))
-        } catch {
-            return (.failure(error))
-        }
+    ) throws -> Profile {
+        try ProfileGenerator.generate(
+            pumpSettings: pumpSettings,
+            bgTargets: bgTargets,
+            basalProfile: basalProfile,
+            isf: isf,
```
dnzxy: Nit: same here.

dnzxy (approved):
LGTM overall. 

Left nit commands about variable (and function signature) naming. Feel free to adjust or leave as-is, entirely optional.

kingst:
I updated the variable names, thanks for the suggestion @dnzxy . Merging now...

PR #1290, opened 2026-07-11
Fix duplicate Nightscout treatment uploads by serializing per-pipeline runs

bjorkert:
## Problem

Trio occasionally uploads the same treatment to Nightscout twice — most visibly Exercise entries for overrides enacted via Shortcuts or remote commands. Reproduced with an APNS remote command producing two documents with millisecond-identical `created_at` and consecutive Mongo `_id`s, i.e. two near-simultaneous POSTs:

- Enacting an override via App Intents / remote control triggers uploads twice: the Core Data save fires the upload pipeline, and `.willUpdateOverrideConfiguration` fires a direct, unthrottled upload. Both fetch the same rows still marked `isUploadedToNS == false` before either has marked them uploaded.
- Nightscout's treatment upsert is a non-atomic find-then-insert (no unique index), so concurrent identical POSTs both insert. Sequential re-posts are deduped server-side; only concurrent ones slip through. (API v3 has the same race, so switching API versions wouldn't help.)
- The 2s throttle didn't prevent this: it only spaced out task starts within one trigger path, never waited for the previous upload to finish, and with `latest: false` silently dropped triggers arriving inside its window. Carb edits had a third, entirely unthrottled path (`HistoryStateModel` calls `uploadCarbs()` directly).

## Solution

Replace the per-pipeline subject/throttle machinery with `NightscoutUploadSerializer`, an actor that coalesces and serializes upload runs per pipeline:

- An idle pipeline starts a run immediately; while a run is in flight, exactly one follow-up is queued and further requests coalesce into it.
- No two runs of a pipeline ever overlap: a follow-up re-fetches only after the previous run marked its rows uploaded, so it finds nothing left to send. Queued runs double as retry attempts after network failures.
- The upload routine is provided once at init; `request(_:)`/`run(_:)` take only the pipeline, so every entry point (Core Data triggers, App Intents, remote control, direct calls) is covered with no call-site changes.
- App Intents semantics are preserved: `uploadOverrides()`/`uploadTempTargets()` are awaitable, so `.didUpdateOverrideConfiguration` still posts only after the upload attempt finished.

## Benefits

- No more duplicate treatments, regardless of how many trigger paths fire for the same change.
- Uploads start immediately — the fixed 2-second throttle delay is gone.
- No dropped triggers (the old throttle discarded a second change arriving within 2 s; it stayed unuploaded until the next unrelated trigger).
- `uploadDeviceStatus()` no longer throws; its only caller was the deleted pipeline dispatch, and errors are logged in `performUpload(for:)`.

Unit tests cover the serializer: no overlap under concurrency, queued-run ordering, burst coalescing into a single follow-up, completion semantics, and pipeline independence.

### DIFF
--- a/Trio/Sources/Services/Network/Nightscout/NightscoutManager.swift
+++ b/Trio/Sources/Services/Network/Nightscout/NightscoutManager.swift
@@ -13,10 +13,7 @@ protocol NightscoutManager: GlucoseSource {
     func deleteCarbs(withID id: String) async
     func deleteInsulin(withID id: String) async
     func deleteGlucose(withID id: String, withDate date: Date) async
-    func uploadDeviceStatus() async
-    func uploadGlucose() async
     func uploadCarbs() async
-    func uploadPumpHistory() async
     func uploadOverrides() async
     func uploadTempTargets() async
     func uploadProfiles() async throws
@@ -47,8 +44,9 @@ final class BaseNightscoutManager: NightscoutManager, Injectable {
     /// Coalesces and serializes upload runs so no two runs of the same pipeline overlap.
     /// Runs execute `performUpload(for:)`, provided at init. The public `upload*()`
     /// methods and `requestUpload(_:)` go through the serializer; the `performUpload*()`
-    /// bodies must never call those entry points, or a same-pipeline run deadlocks
-    /// behind itself.
+    /// bodies must not call the awaitable `upload*()` entry points. The serializer
+    /// asserts on such a call in debug builds and downgrades it to a fire-and-forget
+    /// request in release.
     private var uploadSerializer: NightscoutUploadSerializer!
 
     /// Request an upload for a pipeline (enqueue work). Safe to call from anywhere.
@@ -110,14 +108,11 @@ final class BaseNightscoutManager: NightscoutManager, Injectable {
 
     // MARK: - Upload triggers
 
-    //
     // Each upload pipeline is driven by an NSFetchedResultsController whose predicate is the
     // "not yet uploaded to Nightscout" set for that entity. The controller fires whenever
     // un-uploaded items appear (or drop out after a successful upload), which we map to a
-    // `requestUpload(pipeline)` call (coalesced and serialized per pipeline). Bound to the viewContext, they
-    // also pick up batch-inserted glucose via the persistent history merge in CoreDataStack —
-    // replacing the previous changedObjects publisher plus the glucoseStorage.updatePublisher
-    // fallback.
+    // `requestUpload(pipeline)` call (coalesced and serialized per pipeline). Bound to the viewContext,
+    // they also pick up batch-inserted glucose via the persistent history merge in CoreDataStack.
 
     let determinationUploadControllerDelegate = FetchedResultsControllerDelegate()
     lazy var determinationUploadController: NSFetchedResultsController<OrefDetermination> = {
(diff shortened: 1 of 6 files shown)


dnzxy (approved):

bjornoleh (approved):
Approving based on live testing for a few days. No more duplicate Exercise events from LF remote overrides. 

Thx @bjorkert for fixing this!

dnzxy (dismissed):

Trio/Sources/Services/Network/Nightscout/NightscoutManager.swift
```diff
@@ -13,7 +13,7 @@ protocol NightscoutManager: GlucoseSource {
     func deleteCarbs(withID id: String) async
     func deleteInsulin(withID id: String) async
     func deleteGlucose(withID id: String, withDate date: Date) async
-    func uploadDeviceStatus() async throws
+    func uploadDeviceStatus() async
```
dnzxy: After this PR, `uploadDeviceStatus()` has no callers left . The new `performUpload(for:)` calls `performUploadDeviceStatus()` directly. Rather than changing its signature, could it be dropped from the protocol.

Same applies to `uploadGlucose()` and `uploadPumpHistory()`, which are also caller-less now; only `uploadCarbs()`, `uploadOverrides()` and `uploadTempTargets()` are still called directly.

Trio/Sources/Services/Network/Nightscout/NightscoutManager.swift
```diff
@@ -44,56 +44,32 @@ final class BaseNightscoutManager: NightscoutManager, Injectable {
...
-    /// Subjects used to request an upload pipeline. The pipeline applies a throttle so
-    /// close calls don’t double-upload.
-    var uploadPipelineSubjects: [NightscoutUploadPipeline: PassthroughSubject<Void, Never>] = {
-        var d: [NightscoutUploadPipeline: PassthroughSubject<Void, Never>] = [:]
-        NightscoutUploadPipeline.allCases.forEach { d[$0] = PassthroughSubject<Void, Never>() }
-        return d
-    }()
+    /// Coalesces and serializes upload runs so no two runs of the same pipeline overlap.
+    /// Runs execute `performUpload(for:)`, provided at init. The public `upload*()`
+    /// methods and `requestUpload(_:)` go through the serializer; the `performUpload*()`
```
dnzxy, line 46: > performUpload* must never call the public entry points

This is not enforced by any guarding; violating it is a silent potential deadlock for this pipeline.
Cannot tell, if this is potentially an issue, just noting.

TrioTests/NightscoutUploadSerializerTests.swift
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        func wait() async {
+            if isOpen { return }
+            await withCheckedContinuation { waiters.append($0) }
+        }
+    }
+
+    /// Polls until `condition` is true, giving up after ~2 seconds so a broken
+    /// serializer fails the test instead of hanging it.
```
dnzxy: doc comment says a timeout "fails the test", but on timeout `waitUntil` returns silently and the failure only surfaces later as a generic #expect mismatch

dnzxy (approved):
Overall looking awesome. Few nit comments / questions.

dnzxy (approved):
Few nit comments, overall great.

Trio/Sources/Services/Network/Nightscout/NightscoutManager.swift
bjorkert: Dropped uploadDeviceStatus(), uploadGlucose() and uploadPumpHistory() from the protocol and removed the wrappers.

Trio/Sources/Services/Network/Nightscout/NightscoutManager.swift
bjorkert: Good point, better to have it guarded by code than comments. The serializer now tracks the active pipeline in a @TaskLocal: a re-entrant run() trips an assertion in debug builds and degrades to a fire-and-forget request in release, so it can no longer hang silently. Covered by a new test. I am a bit unsure if it is worth the extra complexity but a silent deadlock would be hard to troubleshoot later on.

TrioTests/NightscoutUploadSerializerTests.swift
bjorkert: True, waitUntil now throws a TimeoutError after the polling loop, so the test fails right at the wait site and the doc comment matches the behavior.

dnzxy (approved):

PR #1291, opened 2026-07-11
[Part 3 of 6] Remove JSON encoding for glucose

kingst:
This commit removes the encoding and decoding of glucose between
CoreData and the oref algorithm. Previously we:
* CoreData -> AlgorithmGlucose -> JSON -> BloodGlucose -> oref

but with this change we now:
* CoreData -> BloodGlucose -> oref

There are a few subtleties in this conversion that were being handled
by the encode/decode phases of the previous pipeline, so I added unit
tests based on the old code to confirm that the conversion is
identical.

One semantic change is that the previous version would serialize dates
to ISO strings, but instead here we just copy the dates over directly
so the dates have sub MS precision.


TrioTests/GlucoseSmoothingTests.swift
```diff
@@ -336,22 +335,8 @@ import Testing
...
-        let decoder = JSONDecoder()
-        decoder.dateDecodingStrategy = .custom { decoder in
-            let container = try decoder.singleValueContainer()
-            let dateDouble = try container.decode(Double.self)
-            return Date(timeIntervalSince1970: dateDouble / 1000)
-        }
-
-        return try decoder.decode([AlgorithmGlucose].self, from: data)
+    private func runFetchAndProcessGlucose(smoothGlucose: Bool) async throws -> [BloodGlucose] {
+        try await glucoseStorage.getGlucoseForAlgorithm(shouldSmoothGlucose: smoothGlucose, fetchHours: 24)
```
dnzxy: Nit: This function is not really necessary. The `bool` parameter is just being passed on, and `glucoseStorage.getGlucoseForAlgorithm()` does also `throw`. 

Totally fine to let it stay.

dnzxy (approved):
LGTM.
Thanks for addressing the move of fetching and processing to GlucoseStorage.

TrioTests/GlucoseSmoothingTests.swift
kingst: Done.

PR #1293, opened 2026-07-12
add swipe action to undo snooze

EJM0:
This simply adds a button to undo the snooze action.
image

### DIFF
--- a/Trio/Sources/Views/SnoozeAlertsSheetView.swift
+++ b/Trio/Sources/Views/SnoozeAlertsSheetView.swift
@@ -17,7 +17,7 @@ struct SnoozeAlertsSheetView: View {
         NavigationStack {
             List {
                 if snoozeUntilDate > Date() {
-                    Section(footer: Text("Swipe left or right to end snooze.")) {
+                    Section(footer: Text("Swipe left to end snooze.")) {
                         HStack {
                             Image(systemName: "moon.zzz.fill").foregroundStyle(.tint)
                             Text(String(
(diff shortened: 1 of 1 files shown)


dnzxy:
Snoozes are ended, not undone. I'd suggest "End Snooze".
We don't do buttons with icons.

dnzxy:
What's the use case to have this present? 
Is a simple swipe action &/or a long press context enough?

EJM0:
Okay I'll adjust the button just red text with a chevron?
No context Menu, it's just present beneath the snooze until ... text.
Uscase: to end a snooze if user has a sundden change of mind or situation, where he she does need to be alerted.

https://github.com/user-attachments/assets/28aa9a6c-eca8-41c5-8858-e254b10f6752

dnzxy:
We never have a half section with text and then the other half being a button — that’s why I asked / suggested whether a row in a Section may simply either swipe or be a context menu (or ideally both), instead of this button that looks very out of place for Trio.

EJM0:
mb, i oriented on the g7kit. yes swiping gesture would be a great solution. id rather stick to only adding the swiping gesture, since this is a action and not a editable list item? and also add a footer introducing the action to the user.

image

EJM0:
https://github.com/user-attachments/assets/78c2891e-1915-48b2-a530-856cbd9dfb7e

dnzxy:
>  oriented on the g7kit

No worries. That's a submodule not done in "our" design unfortunately 😅 

Deletion swipes are usually right-to-left swipes only (on lists, but still). I'd suggest keep it only available in one direction. 
You _may_ consider using an SF Symbol `"hand.draw.fill"` similar to Adjustments view with it, although @g5WS probably would have a comment about that 😅

dnzxy (approved):

PR #1294, opened 2026-07-12
[Part 4 of 6] Remove JSON encoding for carbs

kingst:
This PR replaces the old CoreData -> JSON string -> [CarbsEntry]
pipeline with one that goes directly from CoreData to
[CarbsEntry]. The conversion is almost identical with two changes
coming from (1) including IDs (the old way didn't copy IDs over to the
CarbsEntry objects) and (2) using full precision dates instead of
encoded / decoded ISO string dates.

### DIFF
--- a/Trio/Sources/APS/Storage/CarbsStorage.swift
+++ b/Trio/Sources/APS/Storage/CarbsStorage.swift
@@ -371,15 +371,12 @@ final class BaseCarbsStorage: CarbsStorage, Injectable {
         }
     }
 
-    /// Converts a `Double` to the exact `Decimal` the old JSON round-trip produced.
-    /// The problem is that the CarbEntryStored values are Double, not Decimal so JSON
-    /// was doing some conversion for us.
+    /// Converts a `Double` to a `Decimal` using JSON style conversion
     static func algorithmDecimal(_ value: Double) -> Decimal {
         Decimal(string: value.description) ?? Decimal(value)
     }
 
-    /// Maps a `CarbEntryStored` to the `CarbsEntry` the algorithm consumes, reproducing the coercions
-    /// the old `CarbEntryStored` → JSON → `JSONBridge.carbs` round-trip applied
+    /// Converts CoreData stored carb entries into a struct that the oref algorithm can use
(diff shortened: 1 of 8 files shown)


Trio/Sources/APS/Storage/CarbsStorage.swift
```diff
@@ -344,6 +345,75 @@ final class BaseCarbsStorage: CarbsStorage, Injectable {
...
+
+            if let additionalCarbs = additionalCarbs {
+                entries.append(Self.additionalCarbsEntry(carbs: additionalCarbs, date: carbsDate ?? Date()))
+            }
+
+            return entries
+        }
+    }
+
+    /// Converts a `Double` to the exact `Decimal` the old JSON round-trip produced.
```
dnzxy: Nit: no one needs these prosa comments.

Trio/Sources/APS/Storage/CarbsStorage.swift
```diff
@@ -344,6 +345,75 @@ final class BaseCarbsStorage: CarbsStorage, Injectable {
...
+    /// The problem is that the CarbEntryStored values are Double, not Decimal so JSON
+    /// was doing some conversion for us.
+    static func algorithmDecimal(_ value: Double) -> Decimal {
+        Decimal(string: value.description) ?? Decimal(value)
+    }
+
+    /// Maps a `CarbEntryStored` to the `CarbsEntry` the algorithm consumes, reproducing the coercions
+    /// the old `CarbEntryStored` → JSON → `JSONBridge.carbs` round-trip applied
```
dnzxy: Nit: we don't need before-after comments

TrioTests/CarbsNativeConversionTests.swift
```diff
@@ -0,0 +1,318 @@
+import CoreData
+import Foundation
+import Testing
+
+@testable import Trio
+
+/// Golden tests certifying that the native `CarbEntryStored` → `CarbsEntry` mapping
```
dnzxy, line 7: Nit: let's also cut back comments here.

dnzxy (approved):
Functionally LGTM. 
Subjective nit, but let's please cut back the prosa-like comments and the before/after explanations.

PR #1295, opened 2026-07-12
feat(smoothing): Adaptive Smoothing — a level+rate CGM glucose smoother (selectable, off by default)

tim2000s:
Adds **Adaptive Smoothing** as a selectable CGM glucose smoother in *Settings → CGM → Glucose Smoother*, alongside `None` and the existing `Exponential`. **Off by default** — unless a user picks it, nothing here runs and behaviour is byte-identical to today.

## What it is
A level **and** rate estimator (`x = [G, Ġ]`), so it tracks real trends with less lag while rejecting noise. It's built on an Unscented Kalman Filter core but is much more than a textbook Kalman filter — the useful behaviour comes from the layers around it:
- online-**learned** measurement noise (adapts to *this* sensor's noise, no tuning);
- Huber-style outlier rejection (an isolated spike barely moves the estimate);
- real-trend detection (accelerates onto a genuine fast move instead of lagging);
- an IOB-gated compression-low guard (a suspected compression dip is down-weighted, capped so a real low isn't masked);
- gap-aware segmentation + a Rauch–Tung–Striebel backward pass for the displayed history;
- fail-safe throughout (degenerate cases copy raw, floored at 39).

The core (`GlucoseSmoothingCore`, a dependency-free Swift package) is a 1:1 port of AndroidAPS `UnscentedKalmanFilterPlugin.kt`, validated **bit-exact** against a reference Python implementation and 9 golden vectors. Full details: `GlucoseSmoothing/README.md`.

## How it integrates
When selected, `FetchGlucoseManager` runs the exponential smoother first (proven baseline / fail-safe fallback), then overwrites `smoothedGlucose` where Adaptive Smoothing produced a valid value. `smoothedGlucose` feeds oref (when smoothing is on) and the chart's smoothed line; the primary glucose number and dots always remain the **raw** reading.

## Two bugs this exposed (both fixed here)
Soak-testing the running app against live Nightscout data surfaced two issues, both fixed in this branch:
- **Ordering:** `fetchGlucose` returns readings oldest-first, but the filter requires newest-first; fed backwards it copies raw and goes inert. Now reversed, guarded by a regression test.
- **Live display refresh:** the view context (`automaticallyMergesChangesFromParent = false`) served cached objects, so smoothed values written in another context only surfaced after a relaunch — this affected **both** smoothers (exponential's lag was just invisible because its line hugs the raw dots). Fixed in the shared `setupGlucoseArray`.

## Testing
- `GlucoseSmoothingCore`: golden vectors + Swift↔Python bit-exact parity.
- `UkfOrderingRegressionTest` (TrioTests): guards the ordering bug.
- Verified end-to-end in the iOS Simulator against real Nightscout data (smooths live; the smoothed line refreshes live for both smoothers).

## Scope / safety
A **sensing** change only — no dosing, time-in-range, or glycaemic-outcome claim. Gated behind a default-off setting; with the smoother set to `None`/`Exponential`, none of this runs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tim2000s:
Recreating, replacing exponential rather than running alongside

PR #1300, opened 2026-07-12
Finalized Pump Events and Infer Scheduled Basal In-Memory

dnzxy:
## Summary

Adopts LoopKit's dose lifecycle for pump event storage and records scheduled basal delivery instead of inferring it per consumer. Scheduled basal rates (SBRs) are kept mostly for Trio-internal convenience.

  - **Upsert by dose `syncIdentifier`**: mutable (in-flight) events update in place; finalized reports freeze rows with actual `deliveredUnits` and run times. Previously we kept the commanded first report and dropped the finalized one. Also fixes same-timestamp bolus+TBR shadowing in dedup.
  - **Adopt `replacePendingEvents`**: unasserted mutable rows are revoked (LoopKit `DoseStore` semantics); this should avoid any orphaned doses. Uploads (NS/Health/Tidepool) now send finalized rows only, once.
  - **Scheduled basal recorded** (`isScheduledBasal`): a reconciler fills reporting gaps (open loop, connectivity loss, TBR expiry) from the basal schedule each loop tick; pump-reported data always wins and overlapping synthetic rows self-heal.
  - **Consumers**: TDD sums recorded rows and `deliveredUnits` (inference removed); chart plots SBR segments at reduced opacity; oref input and `currentTemp` intentionally exclude them (commands only); NS is not uploaded SBR rows.

**Migration:** additive optional attributes, lightweight (verified on a populated store with ~1k records of existing pump data). Legacy rows count as finalized, non-scheduled. Verified on pump+CGM sim: mutable and subsequent finalized on same row, 5.7h gap backfilled at schedule rates, TDD/chart/loop work / plot nominally with SBR rows present.

Follow-ups / Left-over TODOs: 
* Health/Tidepool pseudo-event consolidation
* Potentially address open-suspend at 24h look-back boundary (fetch 24 hrs of data, excluding suspend event)


kingst:
Overall looking good.

I'm not sure about storing profile basal rate in core data after all. We'd have to maintain mutable state, update events, etc. Maybe just calculating them in memory when needed (e.g., TDD) is better.

Do you think we can simplify the schedule basal inference? Here's what I'm thinking:

```swift
struct event {
    let start: Date
    let end: Date
    let type // .tempBasal, .suspend, .resume, .profileRateChange, .now
}

// Assume we have a list of events sorted that includes all tempBasal, suspend,
// resume, and profileRateChange events. For all events start == end except for
// temp basal. `profileRateChange` and `now` are fake events to help simplify
// the algorithm
let events = // build event list
for (curr, next) in zip(events, events.dropFirst) {
    let gap = next.start.timeIntervalSince(curr.end)
    if gap > minGapSeconds, curr.type != .suspend {
        // lookup rate        
        // create a scheduled basal that starts at curr.end and ends at next.start
    }
}
```

dnzxy:
@kingst I've addressed both points from your comment and our conversation via DM. I'll move this PR to ready. This will require a ton of in-vitro testing though.

Model/Classes+Properties/PumpEventStored+CoreDataProperties.swift
```diff
@@ -6,11 +6,16 @@ public extension PumpEventStored {
         NSFetchRequest<PumpEventStored>(entityName: "PumpEventStored")
     }
 
+    @NSManaged var actionDuration: NSDecimalNumber?
     @NSManaged var id: String?
+    @NSManaged var insulinType: Int16
```
kingst: Can we make this a String instead of an int? I know that it's a LoopKit type but we could create a small bridge. I'm worried about persisting numbers for enums as it restricts how we can change this enum in the future.

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
```diff
@@ -196,24 +189,70 @@ final class BasePumpHistoryStorage: PumpHistoryStorage, Injectable {
         }
     }
 
+    /// Finalized reports freeze the row with delivered values. Upload flags
```
kingst: So if we cancel a bolus NS won't know about it? This doesn't seem right. I did a simple test with Trio/NS and when I cancel a bolus it sees both the initial bolus and the updated value.

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
```diff
@@ -196,24 +189,70 @@ final class BasePumpHistoryStorage: PumpHistoryStorage, Injectable {
...
+            newPumpEvent.syncIdentifier = identifier
+            newPumpEvent.isMutable = false
             // restrict entry to now or past
             newPumpEvent.timestamp = timestamp > Date() ? Date() : timestamp
             newPumpEvent.type = PumpEvent.bolus.rawValue
             newPumpEvent.isUploadedToNS = false
             newPumpEvent.isUploadedToHealth = false
             newPumpEvent.isUploadedToTidepool = false
+            // insulin type unknown: dose is external to the pump
+            self.applyInsulinSnapshot(to: newPumpEvent, insulinType: nil)
```
kingst: This is going to store the pump insulin's values along with this external dose. But I think it's common for people to use different insulin when taken externally (e.g., Afrezza). Maybe for now we leave it blank and add the UI for people to pick the external insulin type at log time later?

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
```diff
@@ -288,8 +328,7 @@ final class BasePumpHistoryStorage: PumpHistoryStorage, Injectable {
             onContext: context,
             predicate: NSPredicate.pumpEventsNotYetUploadedToNightscout,
             key: "timestamp",
-            ascending: false,
-            relationshipKeyPathsForPrefetching: ["bolus", "tempBasal"]
```
kingst, line 292: what does removing this argument do?

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
```diff
@@ -46,140 +46,133 @@ final class BasePumpHistoryStorage: PumpHistoryStorage, Injectable {
...
+                if let syncIdentifier = newPumpEvent.syncIdentifier {
+                    bySyncIdentifier[syncIdentifier] = newPumpEvent
+                }
+                byTimestampAndType[fallbackKey] = newPumpEvent
+                assertedRows.insert(newPumpEvent.objectID)
+            }
 
-                default:
-                    continue
+            // Mutable rows are the pump's assertions (LoopKit contract): a
```
kingst, line 172: This is a really strange interface. So it basically says to delete all mutable events that are not part of this batch? This can't really be what the semantics are, I can't imagine when a pump manager would use this.

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
```diff
@@ -46,140 +46,133 @@ final class BasePumpHistoryStorage: PumpHistoryStorage, Injectable {
...
+                    }
+                    continue
                 }
+
                 let newPumpEvent = PumpEventStored(context: context)
                 newPumpEvent.id = UUID().uuidString
-                newPumpEvent.timestamp = timestamp
-                newPumpEvent.type = type.rawValue
+                // restrict entry to now or past
+                newPumpEvent.timestamp = min(event.date, Date())
```
kingst, line 112: I know I complain about this every time I see it, but we shouldn't change timestamps if we're using them for keys. Good news is that eventually we can drop the timestamp fallback with this PR :)

kingst (commented):
This PR looks great, I'm going to review this in two parts. This part is about the core mutable + extra properties logic and I'll do a second review on scheduled basal inference.

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
dnzxy: Insulin type being nil does that, doesn’t it? It leaves the type blank. Alternatively we can just apply the same type as the current instantiated PM has configured.

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
dnzxy: It‘s what happens on LoopKit. You have a pump interval, it’s essentially time
between one loop and the next, and all unfinalized doses are kept and then cleared upon next cycle. We can do things differently ofc.

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
dnzxy: I disagree. We need to ensure it’s the correct timestamp on the persisted event, which we aren’t changing, but setting here. 

The timestamp differs from the one the PM may be sending, but the one our CD layer touches is not changed.

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
kingst: Sort of, the function sets insulin type to -1 and pulls the duration and peak from settings and preferences and stores these along with the rest of the data. My suggestion would be for external insulin we leave type, duration, and peak as nil and create an issue that allows people to choose an insulin type for their external doses in the UI in the future.

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
kingst: I'm still confused about why a pump manager would do this. It should have 0 or 1 active temp basal and 0 or 1 active bolus events at any given time. How would we get another temp basal or bolus without finalizing the old one? When would we have a mutable event that we want to delete vs finalize?

This interface is unintuitive, we need to be rigorous with the pump managers and make sure that they use it in a consistent way.

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
kingst: Is your basic assumption that since it's running this code already we'll store the syncIdentifier and the timestamp isn't being used as a key? If so I retract my comment and I agree with your disagreement :) To be clear we are modifying the timestamp that is stored in the CD layer but it doesn't matter because we're not using it as a key to match events anymore.

Model/TrioCoreDataPersistentContainer.xcdatamodeld/TrioCoreDataPersistentContainer.xcdatamodel/contents
```diff
@@ -199,18 +200,26 @@
...
+        <attribute name="actionDuration" optional="YES" attributeType="Decimal"/>
         <attribute name="id" optional="YES" attributeType="String"/>
+        <attribute name="insulinType" optional="YES" attributeType="Integer 16" defaultValueString="-1" usesScalarValueType="YES"/>
+        <attribute name="isMutable" optional="YES" attributeType="Boolean" usesScalarValueType="YES"/>
         <attribute name="isUploadedToHealth" optional="YES" attributeType="Boolean" usesScalarValueType="YES"/>
         <attribute name="isUploadedToNS" optional="YES" attributeType="Boolean" usesScalarValueType="YES"/>
         <attribute name="isUploadedToTidepool" optional="YES" attributeType="Boolean" usesScalarValueType="YES"/>
         <attribute name="note" optional="YES" attributeType="String"/>
+        <attribute name="peakTime" optional="YES" attributeType="Decimal"/>
+        <attribute name="syncIdentifier" optional="YES" attributeType="String"/>
```
kingst, line 212: Should we add a uniqueness constraint at the CoreData layer for syncIdentifier? I don't know how that would work with an optional property and all of the events that don't have this set, but if it's clean it might be a nice defense in depth for our storage layer.

kingst (commented):
One more comment on the finalized pump event part of this PR...

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
dnzxy: Yes and no, I think we mean the same thing and are in agreement.

The `events: [NewPumpEvent]` are _not_ stored in CD. They are the PM emitted event(s), and when going through this workflow here, we persist them. The line in question, so `newPumpEvent.timestamp = min(event.date, Date())` ensures that if you make a meal entry with a carb entry, where you future-date carbs but issue a pump bolus, that the pump bolus is not stored in the future, which isn't possible here anyway, really, but it's a safety fallback.

If we are replacing events, it's an upsert, but the events we are going over here are typically PM given, not CD persisted. We are persisting within this process.

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
dnzxy: I've checked the other PMs and essentially true is the default and passing false is the odd case. 

I'm not super sure, but my understanding is that it's there for optimization, where if the PM sends an update for an existing event, or wants to add a new event, and we don't want resubmit unchanged ongoing events, so they are not emitted? I'm not sure frankly whether this is a realistic use case.

Model/TrioCoreDataPersistentContainer.xcdatamodeld/TrioCoreDataPersistentContainer.xcdatamodel/contents
dnzxy: I think it may cause issues, but can look into it. Conceptually it very much makes sense to me to have a uniqueness constraint on any kind of identifier.

dnzxy: @marv-out any insight on this?

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
kingst: Bottom line, the part that caught my attention is that there is a path where we are deleting insulin delivery events and it's behind an unintuitive interface. I don't know the right way to handle it, but I do know that this is something we need to make sure we get right. Can we get some PM developers to weigh in on this?

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
kingst: yeah, I think we are in agreement, resolving...

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
kingst: @marv-out this one too, I think you added the prefetching recently

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
kingst: We discussed via DM and I'm good with this. My only ask is to add a log statement when we delete an event using this interface

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
dnzxy: You had me scrambling for a moment here to find the answer, sorry for the delay. The behavior here is actually much nicer now. mutable rows = unfinalized events are excluded from all three upload paths, so NS (or TP or AH) only ever receives the finalized row. If say you cancel a  5 Ubolus, which while running will be unfinalized, uploads exactly once, as e.g. 2.4 U. 

I'm not sure what your test is showing. Sounds like the old immediate-upload behavior (?)

This _does_ potentially have a slight downside, i.e. a running temp basal won't show up on NS until it ends. We can either accept that or in a follow-up PR upload mutable events and then amend/upsert on finalization.

Model/TrioCoreDataPersistentContainer.xcdatamodeld/TrioCoreDataPersistentContainer.xcdatamodel/contents
dnzxy: I will definitely need @marv-out to confirm this, but I believe we can add the constraint without any issues **but** SQLite (so CoreData) would treat NULLs as distinct, so our optional sync identifiers here technically will work, but with the NSMergeByPropertyObjectTrumpMergePolicy* we have, a constraint violation may silently merges rows instead of throwing errors… is that correct, Marv?

I wonder if the implemented upsert behavior on the sync identifiers already enforces identity in code to a sufficient degree. 

Obviously dangerous half knowledge here.
*https://developer.apple.com/documentation/coredata/nsmergebypropertyobjecttrumpmergepolicy

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
kingst: Ok, I can see that change so yep, this is the right way to handle it.

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
dnzxy: I think we should restore this. This would remove the relational data sets from the pre-fetch.

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
dnzxy: Let me handle this as nil then. I think your suggestion sounds very sound!
For the time being, we can just hard-code it during storing of external doses as the pump insulin type (that was what was happening before basically, within IOB, as we only know a single type, DIA and peak), and once we add multi-model support also add a picker for external doses 👍

Model/TrioCoreDataPersistentContainer.xcdatamodeld/TrioCoreDataPersistentContainer.xcdatamodel/contents
kingst: I'm good either way you choose here.

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
```diff
@@ -46,140 +46,133 @@ final class BasePumpHistoryStorage: PumpHistoryStorage, Injectable {
...
-                default:
-                    continue
+            // Mutable rows are the pump's assertions (LoopKit contract): a
+            // complete pending report supersedes any it no longer contains.
+            if replacePendingEvents {
+                let request = PumpEventStored.fetchRequest() as NSFetchRequest<PumpEventStored>
+                request.predicate = NSPredicate(format: "isMutable == YES")
+                for orphan in try context.fetch(request) where !assertedRows.contains(orphan.objectID) {
+                    debug(.coreData, "Removing unasserted mutable event \(orphan.syncIdentifier ?? orphan.id ?? "-")")
+                    context.delete(orphan)
```
kingst, line 186: Do we need to delete this event from NS/Tidepool/Apple Health also?

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
```diff
@@ -203,18 +205,26 @@ final class BasePumpHistoryStorage: PumpHistoryStorage, Injectable {
             }
             bolus.isSMB = dose.automatic ?? true
             event.isMutable = dose.isMutable
+            if event.isUploadedToNS, bolus.amount != previousAmount {
```
kingst, line 220: Is there a race condition now on `isUploadedToNS`? If we get a mutable event that is sent to NS, then a finalized event, this check will prevent the clearing of the `isUploadedToNS` flag. When the mutable event returns it would write `isUploadedToNs` to true before it has a chance to upload the finalized event. Rare for sure, but I think it's possible.

kingst (commented):
A few more comments for Part 1, which is looking good. I'll move on to Part 2 now...

dnzxy:
To address your earlier concerns around NS / AH / TP, with [d6b74b5](https://github.com/nightscout/Trio/pull/1300/commits/d6b74b59f692868fbb5e46efa284f759c311c0c6) we are now also uploading mutable doses to NS (as its used for live monitoring), @kingst .

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
dnzxy: I don't think there is a way of testing this really, or knowing. Open for proposals. 
I'll try and run through the logic again and see if it can be addressed via delta comparison.

kingst: I think you have two options: (1) go back to only uploading immutable events or (2) add a new field that keeps track of a token or revision so that the NightscountManager only sets that flag if it was the event that set it.

To test we'd need to do some more advanced dependency injection to block mocked API calls. It would be messy but possible.

I'm leaning slightly towards going back to uploading immutable only to keep things simple, even though we'd be losing some information in Nightscout (which probably only matters for long running temp basal commands)

dnzxy: It would also matter for long running boluses. Anything over half a unit or a unit takes a while before it would get to NS. 

I‘ll try and see about it.

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
dnzxy: NS gets upserted, let me check if we need to delete. AH and TP only get finalized events.

Model/TrioCoreDataPersistentContainer.xcdatamodeld/TrioCoreDataPersistentContainer.xcdatamodel/contents
marv-out: I would add a constraint here on `syncIdentifier` as a **separate** field to our already existing constraint on `id`. And we need to do it in this PR _before_ it gets to dev (risk of constraint violations for users who built this PR without the constraint). But I would also change the MergePolicy of our two MOCs that are relevant here (in `storePumpEvents` and in `storeExternalInsulinEvent`) to `NSMergeByPropertyStoreTrumpMergePolicy` because it aligns with the intention of this PR (finalized rows never change). This is also done in Loopkits Persistent Store. Maybe we could also change it in our CoreDataStack, but it would of course enforce this  merge policy for **ALL** MOCs and I'm unsure about the consequences right now. Have to think about it, but out of scope for this PR and easy to change in the future...

We should also cover this with Tests

marv-out: Another thing: we should change the deletion rule (Parent -> child) to cascade instead of nullify. The constraint merge as well as the replacePendingEvents purge will leave the bolusStored/tempBasalStored childs orphaned in the current implementation and we don't want that

Model/TrioCoreDataPersistentContainer.xcdatamodeld/TrioCoreDataPersistentContainer.xcdatamodel/contents
```diff
@@ -230,8 +239,12 @@
         <attribute name="weightedAverage" optional="YES" attributeType="Decimal" defaultValueString="0.0"/>
     </entity>
     <entity name="TempBasalStored" representedClassName="TempBasalStored" syncable="YES">
+        <attribute name="deliveredUnits" optional="YES" attributeType="Decimal"/>
         <attribute name="duration" optional="YES" attributeType="Integer 16" defaultValueString="0" usesScalarValueType="YES"/>
+        <attribute name="endDate" optional="YES" attributeType="Date" usesScalarValueType="NO"/>
+        <attribute name="isScheduledBasal" optional="YES" attributeType="Boolean" usesScalarValueType="YES"/>
```
marv-out: I think the predicate changes you made (e.g. `NOT (tempBasal.isScheduledBasal == YES)` ) clauses have a side effect. For a relationship keypath that means a LEFT OUTER JOIN — so for any row that has no `tempBasal` at all (boluses, suspends, resumes, primes, alarms...) the joined column is `NULL`, and in SQL `NOT (NULL = 1)` is still `NULL`, which the `WHERE` clause drops. Same story for every pre-existing TBR row, because `isScheduledBasal` is optional with no default, so migration leaves it `NULL`.

That's a problem because `pumpHistoryLast1440Minutes` feeds oref: it would get a pump history with no boluses in it, so IOB is way understated. The upload predicates are hit too — boluses would never make it to NS/Health/Tidepool again (and `NOT (isMutable == YES)` has the same NULL trap for legacy rows). It's easy to miss in testing since the chart and history use `pumpHistoryLast24h`, and a stored `NULL` reads as `false`, so everything looks fine in code and in the sim.

To fix this we need to rewrite the clauses as `(tempBasal == nil OR tempBasal.isScheduledBasal == NO)`, and make `isScheduledBasal`/`isMutable` non-optional with default `NO` so migration backfills legacy rows instead of leaving `NULL` (otherwise `== NO` still misses them).

I tested that with the following Unit Tests: 

```
import CoreData
import Foundation
import Swinject
import Testing

@testable import LoopKit
@testable import Trio

/// Store-level regression tests for the pump-history fetch predicates.
@Suite("Pump Event Predicate Tests", .serialized) struct PumpEventPredicateTests: Injectable {
    @Injected() var storage: PumpHistoryStorage!
    let resolver: Resolver
    var coreDataStack: CoreDataStack!
    var testContext: NSManagedObjectContext!
    typealias PumpEvent = PumpEventStored.EventType

    init() async throws {
        coreDataStack = try await CoreDataStack.createForTests()
        testContext = coreDataStack.newTaskContext()

        let assembler = Assembler([
            StorageAssembly(),
            ServiceAssembly(),
            APSAssembly(),
            NetworkAssembly(),
            UIAssembly(),
            SecurityAssembly(),
            TestAssembly(testContext: testContext)
        ])

        resolver = assembler.resolver
        injectServices(resolver)
    }

    // MARK: - Helpers

    private func bolusEvent(date: Date) -> LoopKit.NewPumpEvent {
        LoopKit.NewPumpEvent(
            date: date,
            dose: LoopKit.DoseEntry(
                type: .bolus,
                startDate: date,
                value: 1.0,
                unit: .units,
                deliveredUnits: 1.0,
                description: nil,
                syncIdentifier: "predicate-bolus",
                scheduledBasalRate: nil,
                insulinType: .lyumjev,
                automatic: false,
                manuallyEntered: false,
                isMutable: false
            ),
            raw: Data("predicate-bolus".utf8),
            title: "Test Bolus",
            type: .bolus
        )
    }

    private func scheduledBasalEvent(start: Date, end: Date) -> LoopKit.NewPumpEvent {
        LoopKit.NewPumpEvent(
            date: start,
            dose: LoopKit.DoseEntry(
                type: .basal,
                startDate: start,
                endDate: end,
                value: 0.8,
                unit: .unitsPerHour,
                deliveredUnits: nil,
                description: nil,
                syncIdentifier: "predicate-sbr",
                scheduledBasalRate: nil,
                insulinType: .lyumjev,
                automatic: true,
                manuallyEntered: false,
                isMutable: false
            ),
            raw: Data("predicate-sbr".utf8),
            title: "Scheduled Basal",
            type: .basal
        )
    }

    /// A temp basal row as it exists after lightweight migration from a
    /// pre-PR store: the new columns were added via ALTER TABLE, so
    /// isScheduledBasal and isMutable are NULL, not false.
    private func insertLegacyTempBasal(timestamp: Date) async throws {
        try await testContext.perform {
            let event = PumpEventStored(context: testContext)
            event.id = UUID().uuidString
            event.timestamp = timestamp
            event.type = PumpEvent.tempBasal.rawValue
            event.isUploadedToNS = false
            event.setValue(nil, forKey: "isMutable")

            let tempBasal = TempBasalStored(context: testContext)
            tempBasal.rate = 1.5 as NSDecimalNumber
            tempBasal.duration = 30
            tempBasal.tempType = TempType.absolute.rawValue
            tempBasal.setValue(nil, forKey: "isScheduledBasal")
            tempBasal.pumpEvent = event

            try testContext.save()
        }
    }

    private func fetchTypes(_ predicate: NSPredicate) async throws -> [String] {
        try await testContext.perform {
            try testContext.fetch(PumpEventStored.fetch(predicate, ascending: false))
                .compactMap(\.type)
        }
    }

    // MARK: - Tests

    @Test("Bolus rows survive the scheduled-basal exclusion in the oref fetch window") func testBolusSurvivesOrefPredicate() async throws {
        let date = Date().addingTimeInterval(-5.minutes.timeInterval)
        try await storage.storePumpEvents([bolusEvent(date: date)], replacePendingEvents: false)

        // Rows without a tempBasal relation must not be dropped by
        // NOT (tempBasal.isScheduledBasal == YES)
        let types = try await fetchTypes(NSPredicate.pumpHistoryLast1440Minutes)
        #expect(types.contains(PumpEvent.bolus.rawValue), "Bolus must be part of the oref pump history input")
    }

    @Test(
        "Legacy temp basals with NULL isScheduledBasal survive the exclusion clauses"
    ) func testLegacyTempBasalSurvivesOrefPredicate() async throws {
        let date = Date().addingTimeInterval(-10.minutes.timeInterval)
        try await insertLegacyTempBasal(timestamp: date)

        let types = try await fetchTypes(NSPredicate.pumpHistoryLast1440Minutes)
        #expect(
            types.contains(PumpEvent.tempBasal.rawValue),
            "Pre-migration temp basal rows must not vanish from the oref pump history input"
        )
    }

    @Test("Bolus rows are still selected for the Nightscout upload") func testBolusSurvivesNightscoutUploadPredicate() async throws {
        let date = Date().addingTimeInterval(-5.minutes.timeInterval)
        try await storage.storePumpEvents([bolusEvent(date: date)], replacePendingEvents: false)

        let types = try await fetchTypes(NSPredicate.pumpEventsNotYetUploadedToNightscout)
        #expect(types.contains(PumpEvent.bolus.rawValue), "Boluses must be uploaded to Nightscout")
    }

    @Test("Scheduled-basal rows are excluded, regular temp basals are kept (control)") func testScheduledBasalExclusionStillWorks(
    ) async throws {
        let start = Date().addingTimeInterval(-30.minutes.timeInterval)
        try await storage.storePumpEvents(
            [scheduledBasalEvent(start: start, end: start.addingTimeInterval(15.minutes.timeInterval))],
            replacePendingEvents: false
        )

        let types = try await fetchTypes(NSPredicate.pumpHistoryLast1440Minutes)
        #expect(
            !types.contains(PumpEvent.tempBasal.rawValue),
            "Pump-reported scheduled basal must stay excluded from the oref input"
        )
    }
}


```

Trio/Sources/APS/Storage/PumpHistoryStorage.swift
dnzxy: Addressed this in latest commit.

Model/TrioCoreDataPersistentContainer.xcdatamodeld/TrioCoreDataPersistentContainer.xcdatamodel/contents
dnzxy: Added this test and the respective changes.

Model/TrioCoreDataPersistentContainer.xcdatamodeld/TrioCoreDataPersistentContainer.xcdatamodel/contents
dnzxy: Changed the default and introduced both new constraint.

Trio/Sources/APS/DeviceDataManager.swift
```diff
@@ -594,9 +597,18 @@ extension BaseDeviceDataManager: PumpManagerDelegate {
                     return $0.dose?.unitsPerHour ?? 0 <= Double(settingsManager.pumpSettings.maxBasal)
                 }
                 debug(.deviceManager, "Storing \(events.count) new pump events: \(events)")
-                try await pumpHistoryStorage.storePumpEvents(events)
+                let purgedUploadedIds = try await pumpHistoryStorage.storePumpEvents(
+                    events,
+                    replacePendingEvents: replacePendingEvents
+                )
                 lastEventDate = events.last?.date
                 completion(nil)
```
kingst, line 629: I don't really know the semantics of pump managers but I would assume that we should not complete this call until after the purge. The problem I'm concerned about is getting another pump event and having the purge execute in parallel with it since this whole logic is executed in a task. So two things to check:

(1) Make sure that we don't get new pump events until after we call complete. If this is true then move the complete to the bottom

(2) If we can get another pump even before we complete, then we need some type of synchronization to make sure that we serialize the handling of pump events on the Trio side

Trio/Sources/APS/Storage/TDDStorage.swift
```diff
@@ -284,272 +299,44 @@ final class BaseTDDStorage: TDDStorage, Injectable {
...
-                }
-            }
-
-            // Record gap if exists
-            if event.timestamp > lastEndTime {
-                gaps.append((start: lastEndTime, end: event.timestamp))
-            }
-
-            lastEndTime = currentEndTime
+        let reported = scheduledBasalEvents.reduce(into: Decimal(0)) { totalInsulin, event in
```
kingst, line 315: Should we have a check to see if end > Date()? The previous code had it. If we do need this check then let's pass that in as a parameter rather than calling Date() to keep things testable and deterministic.

Trio/Sources/APS/APSManager.swift
```diff
@@ -448,16 +448,12 @@ final class BaseAPSManager: APSManager, Injectable {
...
-            [BasalProfileEntry](from: OpenAPS.defaults(for: OpenAPS.Settings.basalProfile)) ??
-            [] // OpenAPS.defaults ensures we at least get default rate of 1u/hr for 24 hrs
+        let basalProfile = storage.retrieve(OpenAPS.Settings.basalProfile, as: [BasalProfileEntry].self) ?? []
 
-        // Calculate TDD
+        // Calculate TDD; uncovered gaps are inferred from the basal profile in memory
         let tddResult = try await tddStorage.calculateTDD(
             pumpManager: pumpManager,
-            pumpHistory: pumpHistory,
+            pumpHistory: pumpHistoryStorage.getPumpHistory(),
```
kingst, line 456: getPumpHistory doesn't include suspends and resumes

Trio/Sources/APS/Storage/TDDStorage.swift
```diff
@@ -217,11 +226,17 @@ final class BaseTDDStorage: TDDStorage, Injectable {
     ) -> Decimal {
         guard !tempBasalEvents.isEmpty else { return 0 }
 
+        // Finalized rows carry the pump-reported total (suspends and pulse
+        // quantization included); timeline math covers only the rest.
+        let reportedInsulin = tempBasalEvents.compactMap(\.deliveredUnits).reduce(0, +)
```
kingst, line 267: Is this going to be a problem after an upgrade? So if there are a bunch of old events without deliveredUnits set?

Trio/Sources/APS/Storage/ScheduledBasalInference.swift
```diff
@@ -0,0 +1,104 @@
...
+        to end: Date,
+        profile: [BasalProfileEntry]
+    ) -> [TimelineEvent] {
+        let calendar = Calendar.current
+        var rateChanges: [TimelineEvent] = []
+        var day = calendar.startOfDay(for: start)
+
+        while day < end {
+            for minutes in profile.map(\.minutes) {
+                let boundary = day.addingTimeInterval(TimeInterval(minutes * 60))
```
kingst: start of day for spring forward days might cause issues, maybe set the boundary directly? I'm not 100% sure so double check, but we might need to set minutes directly rather than adding.

kingst (commented):
A few comments on the schedule basal inference. The algorithm itself looks good.

One question: I was expecting `pumpRecordsBasalProfileStartEvents` to show up somewhere but I don't know how PMs that record basal profile start events work at all so I'm really not sure how they should be handled.

Trio/Sources/APS/DeviceDataManager.swift
dnzxy: Added serialized handling for this one.

Trio/Sources/APS/APSManager.swift
dnzxy: Now includes suspends and resumes.

Trio/Sources/APS/Storage/TDDStorage.swift
dnzxy: Could actually have been. Added handling for this.

dnzxy: We will need to probably revisit this at some point and remove my added handling. 
Now, rows without deliveredUnits fall into the inferred-events path and get the pre-PR calculation (rate x duration), so reportedInsulin only sums rows that carry the "actual" reported pump total. So post-upgrade history computes exactly as it did before, and new rows get the more accurate pump-reported totals.

This semi out-manoveurs the whole idea of rows that have deliveredUnits actually being finished dosages, so may be worth addressing again at a later point.

Trio/Sources/APS/Storage/ScheduledBasalInference.swift
dnzxy: Should be good now.

dnzxy:
Pushed updates addressing your comments.

It's my understanding that PMs that record basal-profile-start events (only MinimedKit does this afaik) emit .basal NewPumpEvents, which we store as temp-basal rows but tagged as isScheduledBasal. TDD counts those as recorded scheduled delivery, and they enter the inference timeline as coverage. Our in-memory inference should then only fill windows no event covers (as a fallback). Non-scheduled-basal-recording pumps like Omnipod, Medtrum or Dana, leave gaps, which is exactly what the inference is for. Hope we are covering both potential SBR paths now.

Trio/Sources/APS/Storage/TDDStorage.swift
```diff
@@ -65,10 +65,21 @@ final class BaseTDDStorage: TDDStorage, Injectable {
...
+        let now = Date()
+        // gaps no event covers ran the pump's schedule; inferred in memory, never persisted
+        let inferredSegments = ScheduledBasalInference.segments(
+            events: Self.timelineEvents(from: pumpHistory),
+            profile: basalProfile,
+            now: now
+        )
+
         // Create pairs of suspend + resume events
         let suspendResumePairs = zip(pumpSuspendEvents, pumpResumeEvents).filter { suspend, resume in
```
kingst: I know you didn't change this code, but I think there is an issue. The number of suspends and resumes aren't always balanced, which can cause the wrong pairs to get matched. Assume the pump is currently suspended and we have S1 R1 and S2. Since it's sorted descending, R1 will pair with S2, which the filter drops.

Trio/Sources/APS/Storage/TDDStorage.swift
kingst: The problem is at the boundary between the old events and the new ones. The most recent event with `deliveredUnits == nil` will not see the next event that has `deliveredUnits != nil` since it's in a separate list. So it's delivery will not get truncated. Admittedly this is small (1 TBR for 24 hours after upgrade) but since this point of this PR is to clean up our insulin accounting it might be worth getting all of these corner cases.

kingst (commented):
My previous comments are all addressed except for the mixed `deliveredUnits = nil` and `deliveredUnits != nil` mixed records. I also potentially found a new issue from the existing code that I commented on.

Trio/Sources/APS/Storage/TDDStorage.swift
dnzxy: After discussing this via DM, we'll leave it as-is.

kingst (approved):
Lgtm from a code review perspective just waiting on test results before merging

bastiaanv:
One suggestion while I was looking into the Dana pump:

HistoryRootView+Treatments.swift, line: 66
`Text(tempBasal.isScheduledBasal ? "Basal" : "Temp Basal")`

Otherwise every normal basal is shown as Temp Basal

bastiaanv:
Currently testing a fix for DanaKit (https://github.com/bastiaanv/DanaKit/pull/52). DanaKit is not emitting temp basal's correctly causing the following issue:
`reason: "Warning: currenttemp running but lastTemp from pumphistory ended 10m ago; canceling temp"`

(Note, the 1.25U/hr are scheduled basal's, not TBR)
IMG_9797

Screenshot by @ETolboom

trioneer-dev:
@bastiaanv ’s issue per his comment was addressed via #1412. 
Testing on MedtrumKit is successful and not showing any faults. 
I’d suggest we proceed to move this to dev. Any objections? @kingst @MikePlante1 @marionbarker ?
@jwoglom we potentially need to adjust TandemKit to adhere to this.

kingst (approved):
re-approving. Assuming testing is done and all of the pump drivers are updated as needed.

MikePlante1:
Copying these in-vitro tests from a PM to here:

version: [a5a464a](https://github.com/nightscout/Trio/pull/1300/commits/a5a464a8e1525e7fd300673873300b888c3ed562)
phone: SE (2nd gen)
iOS: 26.6
CGM: simulator
pump: MiniMed 723 + OrangeLink

<h3>Test:</h3>

- I added some rows to scheduled basal in Therapy Settings in Trio around 12:18.
- Trio loops at 12:25 and sets a TBR for 30 mins, after which I turn off my OrangeLink
- Around 14:26 I turn the OL back on and force a loop cycle where Trio sets a TBR
- History correctly shows all three Scheduled Basal rates after the TBR ended before the new TBR got set
- ‼️ **_But the logs only ever show 0.00 for Scheduled Basal in the TDD Summary_**

[log 20 (1).txt](https://github.com/user-attachments/files/31745098/log.20.1.txt)

2 screenshots before reconnecting the OL:

| . |   |
|---|---|
|  |  |

2 screenshots after reconnecting the OL and forcing a loop cycle:

| . |   |
|---|---|
|  |  |

I did another test because of the weird entry in History at 12:18... and found that if I change the current scheduled basal rate in Therapy settings in Trio... after the next loop cycle, it erroneously interjects a Basal row in the History screen. Erroneous because the actual pump was running a TBR the entire time (verified by looking at the pump's screen).

<details>
<summary>Claude analysis</summary>

<b>Cause</b>: `BasalProfileStartPumpEvent` marks a schedule change, not a delivery change. Saving the profile makes Trio push `setBasalProfileStandard`, the pump logs a `BasalProfileStart` for the newly programmed segment, and we record it as scheduled-basal delivery even though the TBR never stopped. LoopKit doesn't have this problem because `reconciled()` trims a `.basal` dose against the doses around it, so an overlapping temp basal wins; we store the dose raw.
Good news: TDD is unaffected now — the sweep ignores these rows entirely after your fix. The remaining impact is all in the consumers that still read them:

- History — a `Basal` row for a span the pump was running a TBR.
- BasalChart — worse than an extra bar. `barEnd = max(timestamp, nextStart ?? now)` gives the scheduled row a span, and the live TBR's own bar ends at `min(end, nextStart)` where `nextStart` is now this row. So the real TBR gets truncated and scheduled basal is painted over the remainder.
- `GarminManager.fetchTempBasals` — `pumpHistoryLast24h AND tempBasal != nil`` with `fetchLimit: 1` and no `isScheduledBasal` filter, so this row becomes "the current temp basal" on the watch. (`recentPumpHistory` got the filter; this parallel query didn't.)

</details>

Also some slightly weird looking screenshots from beginning of the tests:

| . |   |
|---|---|
|  |  |

mountrcg:
I had a simulation, where I had a suspended pump and later a pump with no more insulin in reservoir.
The later case gets not displayed properly - either it needs the same grey depiction as suspend or a new one (orange). At the moment it is negated and the last known TBR point is directly connected without stepping to a new TBR.

| graph | history |
| --- | -- |
| Screenshot 2026-09-04 at 08 55 43 | Screenshot 2026-09-04 at 08 56 02 |
| Screenshot 2026-09-04 at 11 31 53 | -- |

So I guess after suspension or zero insulin or whatever similar, the basal graph needs to start from zero TBR instead of last TBR.

I can spot the same behaviour in @MikePlante1 screens, just not as prominent.

trioneer-dev:
@mountrcg could you attach logs for that time frame of that simulation? Would make it easier going through the flow of events rather than going by the screenshot 😂

mountrcg:
... brain failure wonder why I did not do it.
[2026-09-04-PR1300-issue.log](https://github.com/user-attachments/files/31858248/2026-09-04-PR1300-issue.log)

trioneer-dev:
@mountrcg according to your logs you were running a state of code that this PR cannot produce.

MikePlante1:
Those updates seem to have addressed the issues I saw. 👍

[log 21.txt](https://github.com/user-attachments/files/31914777/log.21.txt)
[log_prev 18.txt](https://github.com/user-attachments/files/31914780/log_prev.18.txt)

marionbarker:
## Test with Omnipod

✅ Probably successful test
* open loop and phone power off issues can be addressed separately
* criteria for good agreement in the difference between Pod 24 hr ground truth and Trio TDD: 
   * average should be less than +/- 0.1 U
   * min/max should be less than or equal to absolute value of 0.5 U

### Configuration

This is very old data that I finally got around to analyzing. This was also using the *ble-heartbeat* method for pod communications which was backported from next dev before it was merged into OmnipodKit main (as it was 11 Sep 2026).

Set up to use finalized-pump-events, commit 1440ac973

Build at 2026-08-08 15:43:58-07:00

First pod: Activated at 2026-08-08 22:45:39 (UTC), faulted with 0x69 at 2026-08-10 10:35:59, 35 hours
Second pod: Activated at 2026-08-10 14:58:46 (UTC), faulted with 0x6a at 2026-08-12 13:02:04, 46 hours

### Analysis

Compare reported TDD from Trio with respect to "ground truth" from pod total insulin delivery
* For each reported TDD, compare reported insulin delivered by pod minus insulin delivered 24 hours earlier
* Only analyze the second pod because the first 24 hours are needed for establishing the 24-hour-ago record and the rest of the time is used for the analysis
* The analysis uses 
   * Values extracted from the TDD table in the Trio log which is updated at 5 minute intervals (when Trio is looping)
   * Output of the Omnipod Parser which reports every message between the phone and pod, filtered to provide unique Total Insulin Delivered values with the first time stamp at which that value was reported
   * A spreadsheet is used to index into the time stamps from the Trio TDD table into the columns for the unique insulin reported by the pod

#### Issues

The phone and pod were running unattended for the most part and there were interruptions for various reasons. (Phone turned off at one time, Trio stopped looping due a user error).

There were periods of very good agreement between Pod 24 hours delivery value and the TDD reported from Trio.

**Good Agreement**

There was a period of about 12 hours where the unattended phone was well behaved. (Remember that 12 hours of good data requires 36 hours of well-behaved test configuration).

Time range for TDD values (UTC): from 2026-08-11 20:03:10 to 2026-08-12 7:58:04.

All values are in units of insulin. 

✅ The Pod 24 - Trio TDD statistics meet criteria.

| Measurement | Pod 24hr TDD | Trio TDD | Pod 24 - Trio TDD |
|:--|:--|:--|:--|
| Average | 14.48 | 14.47 | 0.01 |
| Minimum | 13.70 | 13.35 | -0.25 |
| Maximum | 15.15 | 15.15 | 0.50 |

marionbarker (approved):
Approve from [analysis of a test](https://github.com/nightscout/Trio/pull/1300#issuecomment-5653786544). Trusting the devs to ensure code review is OK.

PR #1302, opened 2026-07-13
feat(smoothing): replace the exponential glucose smoother with Adaptive Smoothing

tim2000s:
Replaces Trio's double-exponential CGM glucose smoother with **Adaptive Smoothing**. The existing *Settings → CGM → Smooth Glucose Value* toggle now drives Adaptive Smoothing directly; the exponential algorithm and its settings are removed. **Off by default** — with the toggle off, nothing here runs and behaviour is byte-identical to today.

## Why replace exponential
The double-exponential smoother trades lag for smoothness with fixed coefficients: it lags fast moves and can overshoot sustained ones (on real data it dips *below* a held low while masking its onset). Adaptive Smoothing estimates both the glucose **level and its rate**, learns each sensor's noise online, and reacts to genuine trends while rejecting one-off spikes — less lag *and* less jitter, with no per-user tuning.

## What it actually does
Despite the core class name, this is **much more than a Kalman filter**. It estimates `x = [G, Ġ]` (level mg/dL and rate mg/dL·min⁻¹); the useful behaviour comes from the layers around a UKF core:

1. **Unscented state estimation** — constant-velocity + rate-decay model propagated through Van der Merwe sigma points; analytical 2×2 Cholesky, **no linear-algebra dependency**.
2. **Online-learned measurement noise** (the "adaptive" part) — estimates *this* sensor's noise from its own innovations (trimmed-mean) and eases toward it; a quiet sensor is trusted more, a noisy one smoothed harder, automatically.
3. **Huber-style outlier rejection** — a reading far outside the filter's own predicted uncertainty barely moves the estimate (an isolated 300 among 100s is pulled well below 200).
4. **Real-trend detection** — a 2-of-3 same-sign >2σ gate inflates process noise so the filter accelerates onto a genuine fast move instead of lagging.
5. **IOB-gated compression-low guard** — a steep low with little insulin on board is treated as probable sensor compression and down-weighted, capped so a real low isn't masked for long. Requires an IOB provider; **off by default** (fail-safe, matching the production default).
6. **Gap-aware segmentation** — splits at >60-min gaps / invalid spacing / error-code readings; bridges minor gaps by decaying the rate.
7. **Rauch–Tung–Striebel backward pass** — the forward filter serves the live/dosing point; the RTS pass refines past points for the chart.
8. **Fail-safe throughout** — every returned point is ≥39 and non-nil; a point it can't model is filled with the raw value, so removing the exponential fallback loses nothing (an unset point falls back to `glucose` in oref, as before).

The core (`GlucoseSmoothingCore`, a dependency-free Swift package) is a 1:1 port of AndroidAPS `UnscentedKalmanFilterPlugin.kt`, validated **bit-exact** against a reference Python implementation and 9 golden vectors. Full detail: `GlucoseSmoothing/README.md`.

## Integration
`FetchGlucoseManager.applyAdaptiveSmoothingAndStore` runs the engine and writes `smoothedGlucose` (which feeds oref when smoothing is on, and the chart's smoothed line). The primary glucose number and dots always remain the **raw** reading.

## Production faithfulness — matches the real AndroidAPS *pipeline*, not just per-call
AAPS's smoother is a **persistent singleton fed 5-min-bucketed data over a 34 h window**. Per-call bit-exactness isn't enough — the branch reproduces the whole production pipeline, because each layer was found to matter on real data:

- **Persistent instance** — one reused smoother, so learned state (`learnedR`, innovation windows, session counters) carries **across cycles** and **across app restarts** (persisted to `UserDefaults`; reset on sensor change). Rebuilding the filter each cycle diverged from AAPS by up to **~15 mg/dL** on real data.
- **34 h input window** — matches AAPS `LoadBgDataWorker` (24 h + 10 h max DIA). This fetch is **private to smoothing**, so display / dosing / watch fetches are untouched.
- **5-min bucketing** — the window is bucketed onto a 5-min grid before smoothing (a faithful port of AAPS `AutosensDataStoreObject.createBucketedData`), then the smoothed grid is projected back onto each stored reading by linear interpolation. Without it, densely/irregularly-spaced CGM data broke segmentation and spiked the line (see below). Bucketing is pure preprocessing — the UKF core is unchanged, so its per-call parity still holds.

## Bugs this surfaced (fixed here)
- **Ordering** — `fetchGlucose` returns readings oldest-first but the filter requires newest-first; fed backwards it copied raw and went inert. Reversed, guarded by a regression test.
- **No cross-cycle persistence** — the engine was rebuilt every fetch, so learned state never carried while AAPS persists it (~15 mg/dL divergence on real data; invisible to per-call tests). Fixed by reusing one persistent, restart-surviving instance.
- **Sub-2-minute "V-spike"** — a CGM backfill/catch-up landing <2 min after the previous reading tripped a segment break, re-initialising the filter from that raw value → a sharp downward spike below the trace, then a near-vertical recovery. Root-caused by exact replay of a reporting user's trace; fixed by the AAPS-faithful 5-min bucketing above (which regularises the spacing). Regression-tested in `GlucoseBucketingTests`.
- **Live display refresh** — the view context served cached objects, so smoothed values written in another context only surfaced after a relaunch. This is now handled by `dev`'s `NSFetchedResultsController` + persistent-history merge (the earlier manual refresh was superseded when this branch merged `dev`).

## Testing — three-way equivalence to the real AAPS pipeline
Validated that Trio's Swift engine reproduces the **actual shipped AndroidAPS code** and the Python reference, over real + synthetic data:

- **Per-call algorithm** — Trio Swift ≡ real AAPS Kotlin `UnscentedKalmanFilterPlugin` ≡ Python `V4UKF`, **bit-exact across ~665k points** (33 months of real Dexcom data + a synthetic edge-case suite).
- **5-min bucketing** — Swift `GlucoseBucketing` ≡ real AAPS `AutosensDataStoreObject.createBucketedData`, **bit-exact on values *and* timestamps** over the full **333,601-reading, 33-month** real history tiled into production-sized 34 h windows (698 traces / 292,771 bucket points, 0 mismatches).
- **`bucket → smooth` pipeline** — Swift ≡ Python `V4UKF` over the same corpus, worst |Δ| ≈ 1×10⁻¹¹.
- Unit / integration — `GlucoseSmoothingCore` golden vectors + Python parity + `GlucoseBucketingTests`; `GlucoseSmoothingTests` + `UkfOrderingRegressionTest` (TrioTests).
- End-to-end — full iOS-Simulator soak on live Nightscout: smooths live, driven by the toggle, learned state persists across a terminate/relaunch, and the smoothed line refreshes live on new readings.

Harnesses + method live in a separate back-test repo; the bucketing / pipeline parity there uses the real shipped AAPS `createBucketedData` and `smooth()`.

## Scope / safety
A **sensing** change only — no dosing, time-in-range, or glycaemic-outcome claim. Gated behind a default-off toggle; with smoothing off, none of this runs. Fail-safe by construction (any degenerate input copies the raw value, floored at 39 mg/dL).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tim2000s:
This PR is backed up with the following backtest repo that contains the scripts and backtest results run against a couple of years of nightscout data. It's provided as a replicable set-up for reproducibility purposes, and showed parity across the adaptive version in AAPS dev, a python reference build that was used for UKF investigations and the swift version here. 

It has not yet been run with on-device data from a CGM, only live and backtested Nightscout data. 

https://github.com/tim2000s/trio-ukf-backtest

This also ties in with the request raised for AAPS to consolidate smoothing: https://github.com/nightscout/AndroidAPS/issues/4965

mountrcg:
Definitely a progression from Tsunami UCK. Haven't followed what is happening there.

The IOB gated compression low protection is not really activated yet. There is also an issue as Trio always runs the smoothing on entire 24-hour window. With only the latest iob value you would be re-judging the history. Not critical (the 3-reading cap and the <75/>30-drop pattern bounds the distortion, and dosing reacts to the newest points) but also not very clean.
IOB per historic reading is easy to get in Trio, because every loop cycle already persists its IOB.

You might want to store Integer values.

mountrcg:
In the future BG chart will show last 72 hours. Dan has a change in the works.

mountrcg:
I think your golden vector lacks an iob trace as fixed input

tim2000s:
Note / follow-up: the golden-vector reference lacks an IOB input as a fixed vector.

It's a port of the Boost code (which is AAPS version of the adaptive filter with an additional compression protection sourced from Tsunami). As the compression logic isn't currently used in AAPS it was left disabled in this version.

mountrcg:
comparison of smooting in a frequent situation:
| Adaptive UKF | Tsunami UKF | Exponential |
| -- | -- | -- |
| IMG_0966 | IMG_0967 | IMG_0968 |

PR #1306, opened 2026-07-13
Add forecasts to the watchapp

EJM0:
this PR adds the ability to display the Forecast (model take by the forecasts for the live activity) on watch app

chart in chartview was adjusted to span the y-axis over the min and max value of currently selected time range(values of cone or forecast lines included)

Different view with either cone true and false
incoming-169E9546-3385-470F-B504-10B42D49BB23
incoming-498CD1E5-D3B6-415B-BD5A-04EA17201E4F
incoming-7DA7FB89-6C77-42A7-AB97-B6AD5C9E73C4
(bg color is not altered in this pr only on my branche)

settings:
IMG_3731


dnzxy:
Forecast is cool 👏 . Combined view does not work out – everything is clashing.
We'll have to be very careful with sending forecasts to the watch; it adds quite a bit of data. Any regressions on in performance or cutouts on a real device?

EJM0:
Sending forecast worked out pretty well. After a fresh install, the watch app loads a bit, but after that, the updates are coming in not any slower than before. Maybe before merging other Peps wanna try out the pr... to really exclude performance issues
That’s sad since a combined view would greatly improve the interaction hassle :( (everything at on view)

bc of performance issues i first tried to calc taylor polynomials but they where not worth the hustle. since when the graph is bottoming out they freakout. thats why I stuck to the live activity implementation, since this also has a tight data envelope.

dnzxy:
Agreed, good idea. Luckily, we have a few "hardcore" watch users that will be able to support us in testing this. 

Re: the view: We designed the watch app following Apple's design guidelines which has "promoted" the crown to be a primary interactor to move between views, and for view to focus on single core contents and the user to "swipe" between content views. 

These two screenshots below show that it clearly does not work out, and this is probably not the smallest watch screen available to Trio.

| &nbsp; | &nbsp; |
|--------|--------|
| image | image |

EJM0:
okay, I'll remove the combined view. looking forward to some test reports :)

EJM0:
i'd still change the normal chart view behaviour to scale the y-axis range to the min and max value of the glucose+forecast values to that the forecast is better visible

Trio/Sources/Services/WatchManager/AppleWatchManager.swift
```diff
@@ -475,7 +543,15 @@ final class BaseWatchManager: NSObject, WCSessionDelegate, Injectable, WatchMana
             WatchMessageKeys.maxProtein: state.maxProtein,
             WatchMessageKeys.bolusIncrement: state.bolusIncrement,
             WatchMessageKeys.confirmBolusFaster: state.confirmBolusFaster,
-            WatchMessageKeys.units: state.units.rawValue
+            WatchMessageKeys.units: state.units.rawValue,
+            WatchMessageKeys.showForecastWatch: state.showForecast,
+            WatchMessageKeys.isForecastCone: state.isForecastCone,
+            WatchMessageKeys.forecastData: [
+                WatchMessageKeys.forecastStartDate: state.forecastStartDate?.timeIntervalSince1970 as Any,
```
marv-out: the as Any cast could become problematic here when forecastStartDate gets set to nil in the else closure of setupWatchState() (e.g. fresh installation, stale determination etc). It does not unwrap the optional. When the dictionary is handed to the WCSession it gets bridged to NSDictionary and becomes NSNull which could throw an NSInvalidArgumentException. 

Simple fix would be to build it conditionally: 

```
if let start = state.forecastStartDate?.timeIntervalSince1970 {
    forecastData[WatchMessageKeys.forecastStartDate] = start
}
```

Trio.xcodeproj/project.pbxproj
```diff
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		041D1E995A6AE92E9289DC49 /* TreatmentsDataFlow.swift in Sources */ = {isa = PBXBuildFile; fileRef = C8D1A7CA8C10C4403D4BBFA7 /* TreatmentsDataFlow.swift */; };
+		09701ADF300546BE00D4F5C4 /* CombinedGlucoseChartview.swift in Sources */ = {isa = PBXBuildFile; fileRef = 09701ADE300546BE00D4F5C4 /* CombinedGlucoseChartview.swift */; };
```
marv-out: pls clean the project file as you've removed the CombinedGlucoseChartView

Trio Watch App Extension/Views/GlucoseChartView.swift
```diff
@@ -26,12 +26,89 @@ struct GlucoseChartView: View {
         }
     }
 
-    // TODO: should we only change the x axis here like we do in the main chart instead of filtering the values?
+    private var xAxisDomain: ClosedRange<Date> {
+        let now = Date()
+        let pastOffset = -Double(timeWindow.rawValue - 2) * 3600
```
marv-out: just from reading the code it seems a bit strange that we are always subtracting 2h from now but only adding them again if showForecast is true. This should lead to a graph only showing 2h less than we want to if no Forecasts are shown. Have you tested that? I haven't built this yet...

Trio/Sources/Models/WatchState.swift
```diff
@@ -27,6 +27,14 @@ struct WatchState: Hashable, Equatable, Sendable, Encodable, Decodable {
     var bolusIncrement: Decimal = 0.05
     var confirmBolusFaster: Bool = false
 
+    // Forecast options
```
marv-out, line 30: Watch State implements Equatable and Hashable manually and it seems you forgot to add your added variables. I guess it still works because the apple watch path dedupes by using date and sendDataToWatch sets state.date to Date() when sending...anyway, it should be added

Trio.xcodeproj/project.pbxproj
```diff
@@ -1027,6 +1028,7 @@
 /* End PBXCopyFilesBuildPhase section */
 
 /* Begin PBXFileReference section */
+		09701ADE300546BE00D4F5C4 /* CombinedGlucoseChartview.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = CombinedGlucoseChartview.swift; sourceTree = "<group>"; };
```
marv-out: see comment above

Trio.xcodeproj/project.pbxproj
```diff
@@ -5734,6 +5737,7 @@
 				BDA25F222D26D62800035F34 /* BolusInputView.swift in Sources */,
 				BDFF7A882D25F97D0016C40C /* TrioMainWatchView.swift in Sources */,
 				BDA25F202D26D5FE00035F34 /* CarbsInputView.swift in Sources */,
+				09701ADF300546BE00D4F5C4 /* CombinedGlucoseChartview.swift in Sources */,
```
marv-out: see above

Trio.xcodeproj/project.pbxproj
```diff
@@ -3690,6 +3692,7 @@
 				DD8262CA2D289297009F6F62 /* BolusConfirmationView.swift */,
 				BDA25F212D26D62200035F34 /* BolusInputView.swift */,
 				BDA25F1F2D26D5FB00035F34 /* CarbsInputView.swift */,
+				09701ADE300546BE00D4F5C4 /* CombinedGlucoseChartview.swift */,
```
marv-out: see above

marv-out (changes requested):
Generally looking good code-wise. Haven't tested it yet. A few comments though that should be addressed before merging. And ideally a few people would test that before that gets in

Trio.xcodeproj/project.pbxproj
```diff
@@ -470,6 +469,7 @@
 		BD11C001000000000000C001 /* CGMSensorDisplayState.swift in Sources */ = {isa = PBXBuildFile; fileRef = BD11C000000000000000C001 /* CGMSensorDisplayState.swift */; };
 		BD11C001000000000000C003 /* SensorLifecycleArcView.swift in Sources */ = {isa = PBXBuildFile; fileRef = BD11C000000000000000C003 /* SensorLifecycleArcView.swift */; };
 		BD11C001000000000000C004 /* SensorStatusTagView.swift in Sources */ = {isa = PBXBuildFile; fileRef = BD11C000000000000000C004 /* SensorStatusTagView.swift */; };
+		BD136FAE2F6AC55F002217F9 /* NSFetchedResultsControllerDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = BD136FAD2F6AC555002217F9 /* NSFetchedResultsControllerDelegate.swift */; };
```
EJM0, line 475: i don't know why Xcode for me randomly moves definitions in the project file... seems like it did not change anything

EJM0:
Corrected it, sould I ask in the discord?

dnzxy:
@EJM0 please update the PR description and its screenshots 

@marv-out are the changes sufficient? If so, please resolve the open conversations and approve.

Trio/Sources/Models/WatchState.swift
marv-out: you still need to add the missing variables to `static func == (lhs: WatchState, rhs: WatchState) -> Bool`

marv-out (changes requested):
See comment about WatchState

Trio/Sources/Models/WatchState.swift
EJM0: Sorry overlooked those

marv-out (approved):

PR #1309, opened 2026-07-14
Remove explicit allowsConstrainedNetworkAccess setting for Nightscout uploads

Marketka01:
## Summary

This PR removes the explicit `allowsConstrainedNetworkAccess` setting from all Nightscout upload requests.

This restores the default `URLRequest` behavior and lets iOS decide how constrained network access should be handled according to the user's system settings.

## Background

While investigating intermittent Nightscout upload failures, repeated `Constrained path prohibited` errors were observed in the logs. This PR follows the review discussion and removes the explicit flag instead of overriding the system default.

### DIFF
--- a/Trio/Sources/Services/Network/Nightscout/NightscoutAPI.swift
+++ b/Trio/Sources/Services/Network/Nightscout/NightscoutAPI.swift
@@ -85,7 +85,6 @@ extension NightscoutAPI {
         }
 
         var request = URLRequest(url: url)
-        request.allowsConstrainedNetworkAccess = true
         request.timeoutInterval = Config.timeout
 
         if let secret = secret {
@@ -138,7 +137,6 @@ extension NightscoutAPI {
         }
 
         var request = URLRequest(url: components.url!)
-        request.allowsConstrainedNetworkAccess = true
         request.timeoutInterval = Config.timeout
 
         if let secret = secret {
@@ -171,7 +169,6 @@ extension NightscoutAPI {
         ]
 
         var request = URLRequest(url: components.url!)
-        request.allowsConstrainedNetworkAccess = true
         request.timeoutInterval = Config.timeout
         request.httpMethod = "DELETE"
 
@@ -210,7 +207,6 @@ extension NightscoutAPI {
         }
 
         var request = URLRequest(url: url)
-        request.allowsConstrainedNetworkAccess = true
         request.timeoutInterval = Config.timeout
         request.httpMethod = "DELETE"
 
@@ -240,7 +236,6 @@ extension NightscoutAPI {
         }
 
         var request = URLRequest(url: url)
-        request.allowsConstrainedNetworkAccess = true
         request.timeoutInterval = Config.timeout
         request.httpMethod = "DELETE"
 
@@ -279,7 +274,6 @@ extension NightscoutAPI {
         }
 
         var request = URLRequest(url: components.url!)
-        request.allowsConstrainedNetworkAccess = true
         request.timeoutInterval = Config.timeout
 
         if let secret = secret {
@@ -313,7 +307,6 @@ extension NightscoutAPI {
         }
 
         var request = URLRequest(url: requestURL)
-        request.allowsConstrainedNetworkAccess = true
         request.timeoutInterval = Config.timeout
         request.addValue("application/json", forHTTPHeaderField: "Content-Type")
 
@@ -350,7 +343,6 @@ extension NightscoutAPI {
         components.path = Config.uploadEntriesPath
 
         var request = URLRequest(url: components.url!)
-        request.allowsConstrainedNetworkAccess = true
         request.timeoutInterval = Config.timeout
         request.addValue("application/json", forHTTPHeaderField: "Content-Type")
 
@@ -386,7 +378,6 @@ extension NightscoutAPI {
         components.path = Config.statusPath
 
         var request = URLRequest(url: components.url!)
-        request.allowsConstrainedNetworkAccess = true
         request.timeoutInterval = Config.timeout
         request.addValue("application/json", forHTTPHeaderField: "Content-Type")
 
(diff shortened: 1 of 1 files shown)


dnzxy (commented):
Hi @Marketka01 
thanks for this pull request, its clear description and the targeted changes. 

Per [Apple's documentation](https://developer.apple.com/documentation/foundation/urlrequest/allowsconstrainednetworkaccess), `allowsConstrainedNetworkAccess` only applies when users specifically and intentionally enable Low Data Mode in their device's cellular settings. Setting `allowsConstrainedNetworkAccess` to true overrides this, and will lead to quite a high upload volume, i.e. data traffic, when user's explicitely enabled Low Data Mode. 

Apple further recommends to 
> Limit your app’s of use of constrained network access to **user-initiated tasks**, and **put off discretionary tasks until** a **nonconstrained interface becomes available**.

I'd argue that our current setup is following this to a tee and am not fully persuaded this change will bring an improvement in the grand scheme of data usage and uploads from Trio on the user's device. 

I'd welcome and be curious what others that heavily rely on NS data think about this — @bjorkert @codebymini (maintainers of LoopFollow) @AndreasStokholm (maintainer of GlucoseBar) @bjornoleh (core member of our team, and caregiver). 

Thanks! Best, Dan

Marketka01:
Hi Dan,

Thank you for taking the time to review this PR and for the detailed explanation.

You're absolutely right that according to Apple's documentation, `allowsConstrainedNetworkAccess` is intended for specific scenarios and overrides the user's Low Data Mode preference.

That is exactly why this behaviour surprised us.

None of the devices used during testing had Low Data Mode intentionally enabled, yet we repeatedly observed the following errors in the logs:

- `NSURLErrorDomain Code=-1009`
- `Constrained path prohibited`

One additional detail that may be relevant is that the device we tested on is a child's iPhone managed using Apple Screen Time / Family with parental restrictions enabled (limited websites and other content restrictions). We don't know whether these restrictions could influence iOS network behaviour, but we thought it was worth mentioning as part of the test environment.

After enabling `allowsConstrainedNetworkAccess`, these errors disappeared during several days of normal use and Nightscout uploads became reliable again.

We are not claiming that this is necessarily the correct solution or that it explains every Nightscout-related issue. Our main goal was to understand why iOS considered the network to be constrained in this situation when we did not expect it to.

If it would be helpful, we can also provide anonymized logs collected before and after this change.

Thanks again for taking the time to review this PR.

Marketka01:
One additional observation that may be helpful.

The problem did not only occur while moving the phone or when network conditions changed. We also observed it several times while the phone had been lying untouched on a table for approximately one hour, with the child remaining nearby the entire time. Despite that, the logs still repeatedly reported `Constrained path prohibited`.

This made us wonder whether the issue might not be related only to Low Data Mode or changing network conditions, but possibly to another situation where iOS classifies the network path as constrained.

We're not claiming this is the cause—we simply wanted to share another observation from real-world use that might help with the investigation.

dnzxy:
@Marketka01 Trio logs don’t really contain PID markers other than potentially the teamID when running with telemetry enabled, so please do provide logs, we can then try to look into it. 

It may be worth for you to spin up #1290 and run that as a test for some time. I imagine that constrained paths may become temporarily prohibited by a high # of semi-concurrent or very rapid subsequent requests being fired by a single source, i.e., Trio. 

What was the typical # of occurrence of these upload errors over say an hour or a 24 hr window?

Marketka01:
I have attached an anonymized comparison bundle containing log samples from before and after the change, together with occurrence counts by hour.

Please note that the counts represent individual log entries, not necessarily separate user-visible incidents, because one constrained-path event may generate errors for several Nightscout endpoints.
[nightscout_PR1309_anonymized_logs.zip](https://github.com/user-attachments/files/30001351/nightscout_PR1309_anonymized_logs.zip)

bjorkert:
As a parent, I support uploading Nightscout data with high priority under any condition.

Apple's guidance says to "limit your app's use of constrained network access to user-initiated tasks, and put off discretionary tasks until a nonconstrained interface becomes available." A parent does not consider receiving potentially life-critical information, and being able to act on it, a discretionary task. Quite the opposite, nothing is more important.

allowsConstrainedNetworkAccess defaults to true, and as I read it the flag is meant for things like photo syncing, autoplaying video in feeds, and backup tasks, not small health telemetry.

That said, I've never hit this myself, we've been on Trio since the early dev versions and this flag predates that.

For reference, Loop doesn't set this flag at all, so its uploads aren't limited on constrained networks.

My preference would be to simply remove these lines rather than set the flag to its default value.

bjornoleh:
As a caregiver I agree with @bjorkert , the uploading of Trio data to NS is critical in this context. I did actually bump into this when trying to fit the data usage to the cheapest available data plan, but this broke the uploading to NS. It was also not at all obvious to me that this was the cause of my lack of data in NS, so it took me some time to troubleshoot that.

dnzxy:
While I still don’t think this is the actual cause of the issue, I suggest that instead of setting the flag to true, to remove it entirely for all upload requests, as it then will default to true (and respect whatever the phone settings are). 

As for the reasons of uploads potentially failing; I‘d suggest running with the latest fixes for core data and once PR #1290 goes in with that and re-assess :) 

Can you please make those changes to this PR @Marketka01 ?

Marketka01:
Done. I removed the explicit `allowsConstrainedNetworkAccess` setting from all Nightscout upload requests, so the default URLRequest behavior is now used.

Thank you for the suggestion.

dnzxy (dismissed):
Thank you for adjusting the PR.
LGTM. Merging… 🚢

dnzxy:
Just a side note for next time. Review our [contributing guidelines](https://github.com/nightscout/Trio?tab=contributing-ov-file#contributing-to-trio), please always issue PRs from a feature branch based on dev, and target dev with your PRs.

dnzxy (approved):

PR #1310, opened 2026-07-14
Add AccuChek SmartGuide CGM to dev

dnzxy:
## Summary

This PR adds the AccuChek SmartGuide CGM to Trio `dev`.

The Accu-Chek CGM is a 14-day CGM made by Roche.
It comes with its own app and a prediction app, no direct integrations with a pump possible.

This CGM needs to be calibrated twice at the beginning of its lifespan, at the 12h mark and 30 min after.
The first calibration will stay open until you enter the calibration (possible via Trio or the official app).
The second calibration needs to be done after 30min of the first calibration and before 2.5hr.
If you miss this calibration, this process needs to be re-done.

## Additional Notes

This CGM and its Trio driver has been used by multiple users over the span of at least one sensor up to multiple months. We are moving ahead in adding it to Trio `dev` to make it available for more users and potential testing.


MikePlante1 (approved):
Code review and simulator test looks good to me.

Only thing to note is that Accu-chek is missing from this list, but so is Eversense E3 / 365:
https://github.com/nightscout/Trio/blob/2bd5ef07af594c35dd932773c2af183d2f3825a2/Trio/Sources/Modules/CGMSettings/View/CGMRootView.swift#L207-L219

PR #1311, opened 2026-07-14
[Part 5 of 6] Remove JSON encoding for pump events

kingst:
This commit removes the old:
* CoreData -> DTOs -> JSON string -> [PumpHistoryEvent]

pipeline and replaces it with a direct pipeline:
* CoreData -> [PumpHistoryEvent]

The result of the two pipelines should be identical, except for the
handling of the subtype tempType of temp basal events, which we
nullify if we get an unrecognized string vs crashing as the old
pipeline would. This change is safe as Trio only ever records
"absolute" types and the value is copied but never actually used by
the algorithm.


scripts/run_unit_tests.sh
dnzxy, line 1: We can execute tests by just executing the `xcodebuild test` command in this file, instead of calling `./scripts/run_unit_test.sh`. Is this finally not rather obsolete?

Trio/Sources/APS/OpenAPS/OpenAPS.swift
```diff
@@ -133,87 +133,50 @@ final class OpenAPS {
...
-            .date(
-                byAdding: .second,
-                value: -1,
-                to: Date()
-            )! // adding -1s to the current Date ensures that oref actually uses the mock entry to calculate iob and not guard it away
-        let dateFormatted = PumpEventStored.dateFormatter.string(from: oneSecondAgo)
+    private func createSimulatedBolusEvent(simulatedBolusAmount: Decimal) -> PumpHistoryEvent {
+        // Adding -1s to the current Date ensures oref actually uses the mock entry to calculate iob
+        // and does not guard it away.
+        let oneSecondAgo = Date().addingTimeInterval(-1)
```
dnzxy: I'd suggest to remove this and just assign `Date().addingTimeInterval(-1)` to `timestamp`.

dnzxy (approved):
Two minor comments. Overall LGTM, and halleluja 👏  for removing this overhead and JS(ON) boilerplate code.

PR #1315, opened 2026-07-14
Fix Temp Target and Override delete confirmation auto-dismissing

### LINKED ISSUE #1314: Deleting Temp Target (TT) by swiping not possible
## Describe the bug
The UI hinders the user to delete a Temp Target by swiping left as the confirmation popup at the bottom is only displayed for a split second. 

## To Reproduce
1. Create a new Temp Target in 'Adjustments'
2. After creation of the new TT swipe left on it to display then edit- and delete-buttons
3. Hit delete
4. The info-text jumps over the new TT-entry and the confirmation popup on the bottom is displayed only for a second

## Expected behavior
The info-text below of the TT-entries stays at the same place and the confirmation popup waits for user interaction. 

## Screenshots
I attached a short screencast. 

### Smartphone:
* Hardware: *iPhone 13*
* OS Version: *iOS 26.5.2*

### Pump:
* Manufacturer: *Insulet*
* Model: *Omnipod Dash*

### CGM:
* Device: *Dexcom G7*
* Manager app: *Dexcom App*

### Trio Version:
* Version Number: *0.8.4.16* (dev) but issue occurs also on *0.8.4* (main)
* Repo: *[felix-t1d-apps/Trio](https://github.com/felix-t1d-apps/Trio)*
* Git commit: *47028d4*

https://github.com/user-attachments/assets/03f5292a-8802-4b3d-998c-1d11f0f5c1ae

bjorkert:
Fixes #1314

Deleting a Temp Target or Override preset by swiping was unreliable. The delete confirmation only appeared for a moment before auto-dismissing, so the preset often couldn't be deleted.

The confirmation dialog was attached to the ForEach inside the List and driven by two state variables set at the same time, which made it dismiss itself. Reworked it to match the History screen: a single optional drives the dialog via presenting:, attached once to the view. Applied to both Temp Targets and Overrides.

### DIFF
--- a/Trio/Sources/Modules/Adjustments/View/AdjustmentsRootView.swift
+++ b/Trio/Sources/Modules/Adjustments/View/AdjustmentsRootView.swift
@@ -16,7 +16,8 @@ extension Adjustments {
         @State var selectedTempTargetPresetID: String?
         @State var selectedOverride: OverrideStored?
         @State var selectedTempTarget: TempTargetStored?
-        @State var isConfirmDeletePresented = false
+        @State var overrideToDelete: OverrideStored?
+        @State var tempTargetToDelete: TempTargetStored?
         @State var isPromptPresented = false
         @State var isRemoveAlertPresented = false
         @State var removeAlert: Alert?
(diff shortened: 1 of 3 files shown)


dnzxy (approved):
Tested in simulator. Looks good. 
Thank you!

Issue #1308, opened 2026-07-14
Just updated to v0.8.4.15 and some manual boluses continue forever according to treatments which blocks future manual bolus until Trio is forced closed and restarted


kfahling:
## Describe the bug
When manually bolusing now sometimes the bolus is not cleared. Bolus progress at bottom of treatments is stuck at 0 and can't be cancelled thus requiring a restart of Trio before another bolus can be attempted.

## Attach a Log
Need to figure how to upload from iOS if needed or wait till I get wifi

## To Reproduce
*Steps to reproduce the behavior:*
1. Initiate manual bolus 
2. Wait for bolus to complete 
3. Note in treatments if bolus progress is still displayed. If not try again 

## Expected behavior
Treatments should show accurate progress of bolus and progress bar should disappear after bolus completed 

## Screenshots
This was taken first of main screen showing bolus progress 

![image](https://github.com/user-attachments/assets/5e31c7bd-7e33-4f7a-b5da-4c1f8217560d)

This was taken after bolus was completed. And clicking on x does not close it

![image](https://github.com/user-attachments/assets/5c1b4bc9-ae4b-4af0-b952-d6afaaeea753)

### Smartphone:
* Hardware: iPhone 14 Pro
* OS Version: iOS 26.5.2

### Pump:
* Manufacturer: Insulet
* Model: Omnipod Dash

### CGM:
* Device: Dexcom G7
* Manager app: Dexcom App

### Trio Version:
* Version Number: v0.8.4.15
* Repo: *nightscout/trio*
* Git Reference: a874fa0

## Technical Details
*If applicable, provide any technical details that might help in diagnosing the problem. This could include logs, error messages, or relevant configuration details.*

## Additional context
*Add any other context about the problem here.*

kfahling:
I just poked around a bit more and realized that the bogus "bolusing" indicator in treatments is only surfaced AFTER the second bolus is entered into the bolus field.  It did not appear when going into treatments the second time. I also noticed that it cleared up on its' own at least one time if I waited a bit.
I can do more tests if needed just lmk.

kfahling:
I just did a bit more testing and the "stuck stale" bolus indicator in the treatments screen can happen after Trio determined and manual boluses. Restarting Trio has cleared it up so far in all circumstances.

premnirmal:
I am seeing this issue as well

DavidRYL:
[...]

PR #1320, opened 2026-07-15
Make bolus-in-progress card in Treatments view more reliable

bjorkert:
Some manual boluses left a bolusing card stuck at 0% in the Treatments view that blocked further boluses and could not be dismissed until the app was restarted (#1308).

The card was gated on bolusStatus, which mirrors the pump's reported bolusState and can stay at inProgress after delivery has finished. Since bolusTrigger now retains that value, the card resurfaced once a new amount was entered, and the X could not clear it because there was no active bolus to cancel.

Gate the card on bolusProgress instead, the local dose reporter signal that clears reliably and that the Home view already uses, keeping the initiating state so the pre-delivery view still shows.

### DIFF
--- a/Trio/Sources/Modules/Treatments/View/TreatmentsRootView.swift
+++ b/Trio/Sources/Modules/Treatments/View/TreatmentsRootView.swift
@@ -693,9 +693,9 @@ extension Treatments {
         }
 
         private var bolusInProgressForEntry: Bool {
+            // .initiating covers pumps that take a few seconds before reporting progress
             (state.bolusProgress != nil || state.bolusStatus == .initiating) &&
-                state.amount > 0 && !state.externalInsulin &&
-                (state.carbs == 0 || state.fat == 0 || state.protein == 0)
+                state.amount > 0 && !state.externalInsulin
         }
 
         private var disableTaskButton: Bool {
--- a/Trio/Sources/Modules/Treatments/View/TreatmentsRootView.swift
+++ b/Trio/Sources/Modules/Treatments/View/TreatmentsRootView.swift
@@ -492,8 +492,7 @@ extension Treatments {
         }
 
         var treatmentButton: some View {
-            let shouldDisplayBolusProgress = state.bolusStatus != .noBolus && state.amount > 0 &&
-                !state.externalInsulin && (state.carbs == 0 || state.fat == 0 || state.protein == 0)
+            let shouldDisplayBolusProgress = bolusInProgressForEntry
 
             var treatmentButtonBackground = Color(.systemBlue)
             if limitExceeded {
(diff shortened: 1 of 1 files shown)


Trio/Sources/Modules/Treatments/View/TreatmentsRootView.swift
```diff
@@ -693,12 +692,14 @@ extension Treatments {
             pumpBolusLimitExceeded || externalBolusLimitExceeded || carbLimitExceeded || fatLimitExceeded || proteinLimitExceeded
         }
 
+        private var bolusInProgressForEntry: Bool {
+            (state.bolusProgress != nil || state.bolusStatus == .initiating) &&
```
kingst, line 697: Question: Do we want the `== .initiating` clause here? We don't have it in the HomeRootView, which only checks `bolusProgress != nil`.

Trio/Sources/Modules/Treatments/View/TreatmentsRootView.swift
```diff
@@ -693,12 +692,14 @@ extension Treatments {
...
+                state.amount > 0 && !state.externalInsulin &&
+                (state.carbs == 0 || state.fat == 0 || state.protein == 0)
```
kingst: I know that this logic was copied, but what is it trying to do? I can't figure it out and it seems like it might be a bug. I'm talking about the `(state.carbs == 0 || state.fat == 0 || state.protein == 0)` part

kingst (commented):
The PR looks great and fixes a real bug. I had two questions about the change, but I tested it in the simulator and confirmed that it's working as expected.

Trio/Sources/Modules/Treatments/View/TreatmentsRootView.swift
bjorkert: It’s copied and I did not question why it’s there, just wanted to get rid of the bolusStatus.

I can dig a bit deeper into this tomorrow.

Trio/Sources/Modules/Treatments/View/TreatmentsRootView.swift
dnzxy: Intentionally added via #1206 (will be required for Tandem Mobi, as bolusing takes ~5 sec to actually begin).

Trio/Sources/Modules/Treatments/View/TreatmentsRootView.swift
dnzxy: It's supposed to not show when you just enter carbs/fat/protein, but insulin. 
So if you ever enter just a meal, it won't ever show. If you enter a meal and insulin other than external, it'll show.

bjorkert: I removed (state.carbs == 0 || state.fat == 0 || state.protein == 0) now.

bjorkert: In vivo-test successful.

kingst (approved):
LGTM

PR #1321, opened 2026-07-15
Fix Tidepool uploads with serialization

### LINKED ISSUE #1235: Tidepool stops getting my data after a few hours, even though Nightscout keeps working fine
## Describe the bug
I use Trio with Tidepool turned on. My data shows up in Tidepool fine for a few hours, and then it just stops. Nothing in Tidepool updates after that.

The weird part is that everything else keeps working the whole time:

- My CGM readings keep coming in on the Trio screen
- Nightscout keeps getting all my data with no gaps
- Trio keeps looping and dosing normally

It's only Tidepool that stops. Both my glucose readings AND my carbs stop showing up in Tidepool, but they're still in Nightscout.

The only way I've found to fix it is to log out of the Tidepool service in Trio settings and log back in. After I do that, it works again for a few hours, and then it stops again. I have to keep doing this over and over.

## Attach a Log

[log.txt](https://github.com/user-attachments/files/29309537/log.txt)
[log_prev.txt](https://github.com/user-attachments/files/29309576/log_prev.txt)

## To Reproduce
1. Log into Tidepool in Settings - Services - Tidepool
2. Use Trio
3. Wait between 5 minutes and 5 hours and observe missing data in Tidepool ([app.tidepool.org](https://app.tidepool.org/))

## Expected behavior
Trio should continually send data to Tidepool as it does for Nightscout.

## Screenshots

Image

Image

### Smartphone:
* Hardware: *iPhone 17 Pro*
* OS Version: *iOS 26.5.1*

### Pump:
* Manufacturer: *Insulet*
* Model: *Omnipod DASH*

### CGM:
* Device: *Dexcom G7*
* Manager app: *Dexcom App*

### Trio Version:
* Version Number: *0.8.3*
* Repo: *loudestnoise/trio*

 ## What I've already checked

I looked at my Trio logs to try to understand it, and here's what I found:

- The last time Tidepool got any data was at 1:50 AM. After that, nothing.
- But the logs show Trio was wide awake and busy all night — it wasn't asleep or
  closed. It kept doing its normal stuff every 5 minutes the whole time.
- Nightscout kept getting glucose every 5 minutes all night with no problem.
- There were NO error messages about Tidepool. It didn't say anything failed. It
  just quietly stopped trying.
- The next morning, the only time it tried Tidepool again was the exact moment I
  opened the app — and even then it didn't seem to finish.

So from what I can tell, Trio isn't even *trying* to send to Tidepool after it
stops. It's not that it tries and fails — it just stops trying, silently, while
everything else keeps going.

# Some extra detail (in case it helps)

I'm not a developer, but I poked around the code a little with some help from Claude. It looks
like Trio sends glucose to both Nightscout and Tidepool off the same internal
"new glucose" signal. Nightscout keeps responding to that signal, but Tidepool
stops responding to it partway through — so something on the Tidepool side seems
to be quietly dropping off or shutting down while the app is still running.

I also noticed Trio doesn't log anything when this happens, so if there were a way
to add some logging around the Tidepool upload trigger, it might make this much
easier to track down.

# Other things I tried (so you don't have to ask)

- It's **not** the app being closed or asleep — the logs prove it was running.
- I tried a couple of code tweaks myself (with help from Claude) around a token/login timing
  thing and a separate glucose-marking thing. The login-timing change actually
  helped — it went from stopping after about 1 hour to lasting about 5 hours — but
  it still eventually stops the same way. So I think there's a second, separate
  problem that those changes didn't fix.

# What I'm hoping for

Could someone who knows the Tidepool upload part of Trio take a look and see why it
stops getting triggered after a while? If the issue is with `TidepoolKit` I can see if I can get some help on the Tidepool side, but `Loop` uses the same Tidepool Service and does not seem to suffer from this issue.

MikePlante1:
Fixes #1235

## Problem

Trio fired Tidepool uploads concurrently and unserialized: each pipeline's trigger (glucose, carbs, dose, settings observers) independently spawned a `Task` into the shared TidepoolKit `TAPI` actor, so several uploads were routinely in flight at once.

When the OAuth access token expired, multiple in-flight requests each passed `session.shouldRefresh()` and called `refreshSession()` in parallel, reusing the single-use rotating refresh token. That race corrupted the session; subsequent upload completions never fired, so uploads stopped silently with nothing logged.

Confirmed from the issue logs in #1235: last Tidepool success at 01:50:06, zero Tidepool errors after, app running continuously to 07:33; at 07:31 a dose reached the upload path but its completion never returned.

Nightscout uploads were unaffected (separate manager), and Loop avoids the bug by funneling uploads through one serialized `RemoteDataServicesManager` cycle.

## Fix

A new `TidepoolUploadSerializer` actor routes every upload (glucose, carbs, dose/pump events, settings, deletes) through a single chain so only one network request is in flight at a time. This guarantees at most one session refresh, eliminating the concurrent-refresh race, and a wedged request now resolves to a logged failure instead of silently stalling every later upload forever.

Design points (several adopted from the Nightscout upload serializer in #1290):

- **Run-time fetching + coalescing.** Each pipeline's operation fetches its pending rows (or rebuilds its payload) inside the serialized run rather than at trigger time. Two triggers firing close together previously could both snapshot the same unmarked rows and upload them twice; now the second coalesces into a follow-up that re-fetches after the first run marked its rows. A trigger burst costs at most one waiting run per pipeline. Delete pipelines carry unique payloads and never coalesce.
- **Per-chunk marking.** Chunked uploads mark only the chunk that succeeded as uploaded, instead of marking the full batch after the first chunk.
- **Watchdog.** A wall-clock watchdog abandons the chain when its head operation exceeds 10 minutes (e.g. the app was suspended mid-request and the completion was lost). It's checked at each enqueue and on app foreground — the reliable recovery points while the pipeline itself is stalled. Abandoning bumps a generation counter; operations re-check it between network calls so an orphaned chain can't fire requests alongside its replacement.
- **Timeout backstop.** `awaitUpload` bridges TidepoolKit's completion-based calls into async/await with a 120s timeout, so a call that never invokes its completion resolves to a logged `.timedOut` failure instead of suspending the chain forever.
- **QoS.** Chain tasks run at a pinned `.utility` rather than inheriting the caller's priority — `.utility` stays scheduled during background execution windows (`.background` would not), without running network uploads at UI priority.

## Also in this PR

- **Log fixes.** Tidepool upload results were logged as `[Nightscout]`, making Tidepool failures look like Nightscout activity; they're now tagged `[Service]`, and carbs-delete results are worded distinctly from carbs uploads.

## Tests

`TidepoolUploadSerializerTests` covers:

- Operations run strictly one at a time, in enqueue order (peak concurrency stays at 1).
- A request burst for a coalescing pipeline folds into exactly one follow-up run; delete pipelines never coalesce; coalescing is keyed per pipeline.
- The watchdog abandons a wedged chain so later uploads run again, clears coalescing state so the pipeline can run on the fresh chain, and leaves an idle chain alone; an abandoned operation sees a stale generation and bails out, and its late completion can't corrupt the fresh chain's watchdog state.
- `awaitUpload` returns the completion's result, resolves to `.timedOut` when the completion never fires, and ignores a late completion after the timeout.


Trio/Sources/Services/Network/TidepoolManager.swift
```diff
@@ -364,102 +402,102 @@ extension BaseTidepoolManager {
...
+            guard await uploadSerializer.isCurrent(generation) else { return }
+
+            let pumpResult = await TidepoolUploadSerializer.awaitUpload("pumpEvents") { completion in
+                tidepoolService.uploadPumpEventData(pumpEvents, completion: completion)
+            }
+            switch pumpResult {
+            case .success:
+                debug(.service, "Success synchronizing pump events data. Upload to Tidepool complete.")
+                let pumpEventType = events.map { $0.type.mapEventTypeToPumpEventType() }
+                let pumpEventsToMark = events.filter { _ in pumpEventType.contains(pumpEventType) }
```
dnzxy: `pumpEventType.contains(pumpEventType)` compares the array against itself, so this filter is always true.

`pumpEventsToMark` always holds all fetched events, not just the ones that mapped to pump events. 

So, if `uploadDoseData` fails but `uploadPumpEventData` succeeds, this marks the bolus/temp-basal rows `isUploadedToTidepool = true` even though their dose upload failed, and those doses are permanently dropped from Tidepool, as subsequent fetches will exclude them via the fetch predicate. 

This is a pre-existing issue I only noticed through review + backtesting this PR against TP's integration environment, but since these lines moved anyway, let's fix it:
`let pumpEventsToMark = events.filter { $0.type.mapEventTypeToPumpEventType() != nil }` pairs each marking with the upload that actually carried it.

Trio/Sources/Services/Network/TidepoolUploadSerializer.swift
```diff
@@ -0,0 +1,173 @@
...
+        )
+        generation += 1
+        tail = nil
+        waiting.removeAll()
+        headStart = nil
+        headPipeline = nil
+    }
+
+    private func begin(_ pipeline: TidepoolUploadPipeline, generation: Int) -> Bool {
+        guard isCurrent(generation) else { return false }
```
dnzxy: When the iOS watchdog abandons a chain, an operation queued behind (intentional) "throttle" the serializer adds, things will bail out here silently. For coalescing pipelines that's lossless, because every next pipeline trigger triggers re-fetches, but for the deletion pipelines the payload will be gone for good and the remote treatment on TP's end silently survives. 

Accepting that trade-off seems fine from Trio's side (replaying deletes across an abandoned element has its own risks), but a log entry `warning(.service, ...)` here, at least when `!pipeline.coalesces`, would mayybe make the loss diagnosable in logs.

I think this is a super edge case.

Trio/Sources/Services/Network/TidepoolManager.swift
```diff
@@ -303,31 +329,34 @@ extension BaseTidepoolManager {
...
+            syncVersion: nil,
+            userCreatedDate: nil,
+            userUpdatedDate: nil,
+            userDeletedDate: nil,
+            operation: LoopKit.Operation.delete,
+            addedDate: nil,
+            supercededDate: nil
+        )]
+
+        Task { [weak self] in
```
dnzxy, line 350: Wrapping enqueue in an unstructured Task means two rapid delete calls can race to the actor and enqueue out of call order (our old TP uploader serial processQueue was FIFO). Practically irrelevant for user-initiated deletes, I'll just note the behavior change (as certain operations in Trio may also lead to implicit deletions).

Trio/Sources/Services/Network/TidepoolManager.swift
```diff
@@ -364,102 +402,102 @@ extension BaseTidepoolManager {
...
-                        case let .failure(error):
-                            debug(.nightscout, "Error synchronizing dose data with Tidepool: \(String(describing: error))")
-                        case .success:
-                            debug(.nightscout, "Success synchronizing dose data. Upload to Tidepool complete.")
-                            Task {
-                                let insulinEvents = events.filter {
-                                    $0.type == .tempBasal || $0.type == .tempBasalDuration || $0.type == .bolus
-                                }
-                                await self.updateInsulinAsUploaded(insulinEvents)
+                    debug(.service, "TIDEPOOL DOSE ENTRIES: \(insulinDoseEvents)")
```
dnzxy: Existing behavior, but maybe worth removing: this dumps _every single dose entry_ on _every run_ of the uploader. Given this PR's log-tagging cleanup (thanks Mike!) consider trimming to a `count ("TIDEPOOL DOSE ENTRIES: \(insulinDoseEvents.count)")` maybe?

dnzxy (commented):
Overall great enhancement / refactor based on Jonas' work for the NS serialzer.

I ran the serializer test suite on the PR head and all tests pass, no flakiness across runs.
I also backtested against Tidepool's integration env with a test account firing glucose/carbs/bolus/pump-event pipelines concurrently, in same shape as our triggers firing together. With a fresh token all uploads go through serialized, no spurious refresh.
With the session force-expired, the first op refreshes against Keycloak, everything still uploads, session stays intact. I verified via the TP's data API and on the dashboard that all records actually landed.

One caveat: TP's integration env seems to tolerate refresh-token reuse (two refreshes with the same token both return http status 200), so the #1235 corruption itself can't be reproduced there. I'm guessing their prod env is configured stricter. However, the  client-side race this PR fixes is real either way: TAPI.refreshSession() suspends at three awaits, so you will receive concurrent 401s for all POST requests using the same refresh token. In other words, serializing our upload is the right fix.

dnzxy:
How‘s this one looking?

MikePlante1:
> How‘s this one looking?

I just pushed the changes.

trioneer-dev (approved):

PR #1322, opened 2026-07-15
Enable Telemetry and Crash Reports by Default

dnzxy:
* Remove the onboarding diagnostics step and the one-time migration sheet.
* Both data streams (telemetry and crash reporting) now default to on (nil flag = enabled)
* Enables telemetry and crashlytics crash report sharing for every user when this is installed – post-install opt-out still open to every user
* Opt-out remains in Settings > Features > App Diagnostics
* Update privacy policy; drop obsolete shadow-mode `PRIVACY.md` (was a left over from our shadow mode testing of oref-swift port)


t1dude (approved):
Approved as per discussion. I have _not_ done a thorough code review.

MikePlante1 (approved):

MikePlante1 (approved):

PR #1325, opened 2026-07-16
Fix temp target presets showing up in the main chart

### LINKED ISSUE #1324: TempTarget create bug from #1315
#1315 introduced a bug

Create a new Temp Target
Image

and save
Image

even though never activated, it will show up in Chart, even though another TT is running
Image

bjorkert:
Fixes #1324

Saving a new Temp Target as a preset made it show up in the main chart even though it was never activated. The chart's fetch predicate freezes its date anchors when it is built and only re-anchors when the app comes to the foreground, so a fresh preset (date now, enabled false) matched the clause meant for scheduled temp targets. The same stale anchor also kept a cancelled temp target's full bar on the chart until the app was backgrounded.

Presets are now excluded from the scheduled clause, and the fetched results are filtered again with a fresh date before they reach the chart. Added tests covering the predicate and the visibility rules.

### DIFF
--- a/Model/Helper/TempTargetStored+Helper.swift
+++ b/Model/Helper/TempTargetStored+Helper.swift
@@ -18,16 +18,23 @@ extension NSPredicate {
     static var tempTargetsForMainChart: NSPredicate {
         let date = Date.oneDayAgo
         return NSPredicate(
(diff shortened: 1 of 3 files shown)


mountrcg:
tested successfuly

dnzxy (approved):
Looks good! Thanks for also adding another test case.

PR #1326, opened 2026-07-16
[Part 1 of 7] Home Layout v2 – Set Home zoning with fixed slots and flexible chart

dnzxy:
Part 1 of 7 of the home layout v2 series.

Introduces the zone architecture for the home dashboard: fixed-height slots for header, meal row, and bottom controls (`HomeLayout` constants), with the chart as the only flexible zone taking the remaining height. Bottom controls move into a `safeAreaInset` so the tab bar can never cover them; dashboard type size is capped at XXL so the fixed zones cannot bust.

State changes (banner, badge, bolus bar) no longer reflow the layout. Verified on SE3, 17 Pro, and 17 Pro Max simulators.

No visual changes of the home view.


marv-out (approved):
Code wise LGTM

PR #1327, opened 2026-07-16
[Part 2 of 7] Home Layout v2 – Rework Home layout: panels, TIR banner, meal row

dnzxy:
Part 2 of 7 of the home layout v2 series, stacked on `home-refactor-zoning`.

Applies the layout design worked out with @marv-out out onto the zone structure: restructured header with pump/loop panels, glass-style adjustments card with remaining-time bar, stats banner slot, meal info row with alarms pill, floating chart-info button, and the top-trailing temp basal rate label. Time-interval buttons are removed; the chart y-axis gains a domain floor so threshold/target lines never clamp to the plot edge.

No Liquid Glass yet (compatibility mode stays on until part 7). Verified light/dark on SE3, 17 Pro, and Pro Max.

| 17 Pro iOS 26.4 | 17 Pro Max iOS 26.5 | 16 Pro iOS 18.5 | SE 3 iOS 26.5 |

### DIFF
--- a/Trio/Sources/Modules/Home/View/HomeRootView+BottomControls.swift
+++ b/Trio/Sources/Modules/Home/View/HomeRootView+BottomControls.swift
@@ -207,9 +207,6 @@ extension Home.RootView {
                     .foregroundStyle(.secondary)
             }
         }
-        .onTapGesture {
-            selectedTab = 2
-        }
     }
 
     @ViewBuilder func adjustmentsTempTargetView(_ tempTargetString: String) -> some View {
@@ -222,9 +219,6 @@ extension Home.RootView {
                     .font(.caption)
                     .foregroundStyle(.secondary)
             }
-        }
-        .onTapGesture {
-            selectedTab = 2
         }
     }
 
@@ -305,8 +299,6 @@ extension Home.RootView {
             Image(systemName: "xmark.app")
                 .font(.title)
                 .foregroundStyle(Color.clear)
-        }.onTapGesture {
-            selectedTab = 2
         }
     }
 
@@ -359,16 +351,19 @@ extension Home.RootView {
                 )
                 .frame(height: HomeLayout.bottomPanelHeight)
                 .overlay(alignment: .bottom) {
-                    if isConcurrent {
-                        HStack(spacing: 6) {
-                            remainingBar(overrideRemainingFraction, tint: .purple)
-                            remainingBar(tempTargetRemainingFraction, tint: .loopGreen)
+                    // inset clears the corner curve so the bar stays inside the shape
+                    Group {
+                        if isConcurrent {
+                            HStack(spacing: 6) {
+                                remainingBar(overrideRemainingFraction, tint: .purple)
+                                remainingBar(tempTargetRemainingFraction, tint: .loopGreen)
+                            }
+                        } else if let tint = tint {
+                            remainingBar(overrideRemainingFraction ?? tempTargetRemainingFraction, tint: tint)
                         }
-                        .padding(.horizontal, 4)
-                    } else if let tint = tint {
-                        remainingBar(overrideRemainingFraction ?? tempTargetRemainingFraction, tint: tint)
-                            .padding(.horizontal, 4)
                     }
+                    .padding(.horizontal, 14)
+                    .padding(.bottom, 3)
                 }
                 .clipShape(RoundedRectangle(cornerRadius: 17, style: .continuous))
                 .shadow(color: Color.black.opacity(colorScheme == .dark ? 0.25 : 0.10), radius: 3, y: 1)
@@ -440,7 +435,13 @@ extension Home.RootView {
                 } message: {
                     Text("Select Adjustment")
                 }
-        }.padding(.horizontal, 10)
+        }
+        // whole panel navigates; the cancel buttons' own gestures take precedence
+        .contentShape(Rectangle())
+        .onTapGesture {
+            selectedTab = 2
+        }
+        .padding(.horizontal, 10)
     }
 
     @ViewBuilder func bolusView(_ progress: Decimal) -> some View {
@@ -531,7 +532,37 @@ extension Home.RootView {
         }
     }
 
+    /// Mean glucose (mg/dL) of today's readings, nil without data.
+    private var todayMeanGlucose: Double? {
+        let startOfDay = Calendar.current.startOfDay(for: Date())
+        let values = state.glucoseFromPersistence
+            .filter { ($0.date ?? .distantPast) >= startOfDay }
+            .map { Double($0.glucose) }
+        guard !values.isEmpty else { return nil }
+        return values.reduce(0, +) / Double(values.count)
+    }
+
+    private var todayAverageString: String {
+        guard let mean = todayMeanGlucose else { return "--" }
+        if state.units == .mmolL {
+            return Decimal(mean).asMmolL.formatted(.number.precision(.fractionLength(1))) + " " + GlucoseUnits.mmolL.rawValue
+        }
+        return "\(Int(mean.rounded())) " + GlucoseUnits.mgdL.rawValue
+    }
+
+    private var todayGMIString: String {
+        guard let mean = todayMeanGlucose else { return "--" }
+        let gmiPercentage = 3.31 + 0.02392 * mean
+        // settingsManager is injected after first render; default until then
+        if state.settingsManager?.settings.eA1cDisplayUnit == .mmolMol {
+            let gmiMmolMol = (gmiPercentage - 2.152) * 10.929
+            return "\(Int(gmiMmolMol.rounded())) mmol/mol"
+        }
+        return gmiPercentage.formatted(.number.precision(.fractionLength(1))) + " %"
+    }
+
     @ViewBuilder func statsBanner() -> some View {
+        let face = state.settingsManager?.settings.homeStatsPanelFace ?? .timeInRange
         let distribution = state.todayGlucoseDistribution
         let coveragePct = distribution.veryLowPct + distribution.lowPct + distribution.inRangePct + distribution
             .highPct + distribution.veryHighPct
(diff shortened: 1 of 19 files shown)


EJM0:
How about moving the alert button and info next to the adjustments.
With that the chart could be cleaner no overlapping basal and info button

Mockup:
image

dnzxy:
It‘s intentionally placed in an overlapping elements way :) both those elements belong to the chart semantically.

polyfeux:
Can the TIR-Banner be disabled optionally?

I don’t want/need to see it at all times (the statistics are sufficient for me), it takes also a lot of space & is positioned very dominantly in the foreground compared to little info it provides.

dnzxy:
@polyfeux it cannot I‘m afraid, as it (will) serve(s) multiple purposes, see #1328 :) and we‘ll add a few more in upcoming PRs, such as information around releases, different system states that prompt a call-to-action, etc. :)

t1dude:
> @polyfeux it cannot I‘m afraid, as it (will) serve(s) multiple purposes, see #1328 :) and we‘ll add a few more in upcoming PRs, such as information around releases, different system states that prompt a call-to-action, etc. :)

If we wanted to comply with a «don’t throw TIR in my face please» approach, we could leave a «No notifications» or something like that in the banner for those who don’t want to show TIR. Just thinking out loud.

polyfeux:
Then let me re-phrase: it would be great if the „TIR xyz% graph“ could be hidden.

Using the space for notifications is understandable, while something like this could be also put into a notification-panel, which could be typical opened by a 🔔-button. 

But while I am already sometimes a bit overwhelmed by all the numbers on the main screen, a TIR-graph which is displayed _always_, would finally break my camels back. 

Especially for those who simply cannot reach 100% TIR and would then reminded of this failure every time they open the app.

polyfeux:
> If we wanted to comply with a «don’t throw TIR in my face please» approach, we could leave a «No notifications» or something like that in the banner for those who don’t want to show TIR. Just thinking out loud.

Or, to only use the limited phone screen (especially in that area) for stuff with actual information, just don’t show anything if there is currently no notification to display.

dnzxy:
Thanks for the feedback, appreciate it. Given the timing of this PR landing, and the first few comments, have you yet had the chance to _use_ the proposed changes, or just looked at screenshots? If not, I'd ask you to use it, make notes, then come back to giving feedback, not the other way around. 

> Then let me re-phrase: it would be great if the „TIR xyz% graph“ could be hidden.
> Using the space for notifications is understandable, while something like this could be also put into a notification-panel, which could be typical opened by a 🔔-button.

I can see if we can potentially make something adjustable, but it's not about _just_ showing "No notifications", but the panel also being an action button, i.e. for the TIR panel it takes you to Statistics, for another panel it may open a different view or prompt. 

This goes very much into #1328, but you will run into a situation, where there's another active panel, and we're not looking to block access to functionalities (just realizing it's missing in the committed state, but on the various testing builds here), so there's a "fallback" access path via History > Toolbar Item.

> Especially for those who simply cannot reach 100% TIR and would then reminded of this failure every time they open the app.

I'd argue that the TIR consensus goal are set at >70%, not 100%; not even a normoglycemic person will get to 100% TIR. It's never been about achieving 100. 

>  for stuff with actual information,

We have quite a few diabetes apps that have set precedent around displaying this information on the main screen, in the commercial and the OSS space. TIR is information, it's not "fluff".

t1dude:
I _have_ been using it. And I must say the banner does not bother me, it does tell me in a good way when readings are missing etc and the TIR graphic is discrete. I see that Dexcom has moved statistics away from the main page again after having it there for a while. Not sure why. I think this comes down to personal preference.

dnzxy:
> I can see if we can potentially make something adjustable […]

I have added some alternative options to this PR now to lessen the "numbers in my face" concerns.
I have also added the missing – by mistake – second path to access Statistics via History > leading Toolbar item.

| Settings | TIR w/ number | Distribution only | Today's Averages |
|  image | image | image | image | 

The panel itself will not be made hideable, and it will stay there, as it's an integral part of the new home layout v2.

polyfeux:
Thank you for taking notice of my concern and adding other options to display. 👍 

May I suggest another option: just make another entry to display the word ‚Statistics‘ in the banner, so the banner-item stays there and provides a direct jump to statistics?

As I am still away in Rehab till next week, I have only my phone with me and will surely test the new features as soon as they are merged into dev.

dnzxy:
> May I suggest another option: just make another entry to display the word ‚Statistics‘ in the banner, so the banner-item stays there and provides a direct jump to statistics?

At that point it truly becomes a "useless", i.e. no information-carrying, panel and just an oversized button, so I'm not inclined to adopt this. 

If you have any suggestions to provide Statistics that you do not deem overwhelming or stress-inducing, I'm all for it. 
I'd have guessed an average glucose value and GMI or just a progress bar does that already. I could potentially see something like average glucose, total daily insulin and total daily carb consumption (for today, so since midnight, like the other displayed metrics), e.g. 
`ø 135 mg/dL  ·  TDI 34 U   ·  103 g`
`Today's averages and consumption`

As mentioned, just a label and an on-tap-action is something I'm opposed to adopting, apologies.

Trio/Sources/Modules/Home/View/HomeRootView+BottomControls.swift
```diff
@@ -0,0 +1,667 @@
...
+        let hasData = coveragePct > 0
+        let tirString = hasData
+            ? distribution.inRangePct.formatted(.number.precision(.fractionLength(0 ... 1))) + " %"
+            : "-- %"
+        let segments: [(color: Color, fraction: CGFloat)] = hasData ? [
+            (.red, CGFloat(distribution.veryLowPct / 100)),
+            (.orange, CGFloat(distribution.lowPct / 100)),
+            (.loopGreen, CGFloat(distribution.inRangePct / 100)),
+            (.purple, CGFloat((distribution.highPct + distribution.veryHighPct) / 100))
+        ] : [(Color.secondary.opacity(0.3), 1)]
```
MikePlante1, line 579: Why just split very low from low, and not also very high from high and in tight range from in range?

The stats banner's color scheme seems to mostly match the dynamic color schedule (even when the static color scheme is selected), except with the system purple used by Override UI elements rather than the more muted purple. Since it's taking you to the stats view, though, I'd recommend just using the color scheme from the daily distribution chart (non-daily splits highs into one more range)

| . | TIR stats banner | static color | dynamic color | daily distribution stats chart |
|---|---|---|---|---|
| very low | `#FF3B30` system red | `#FF3B30` system red | `#E65C5C` muted red | `#AF52DE` system purple |
| low | `#FF9500` system orange | `#FF3B30` system red | `#E65C5C` muted red | `#FF3B30` system red |
| in tight range | `#6FCF97` loopGreen | `#34C759` system green | gradient, `#5CE65C` muted green at target | `#34C759` system green |
| in range | `#6FCF97` loopGreen | `#34C759` system green | gradient green → teal/blue | `#2A9F47` darkGreen |
| high | `#AF52DE` system purple | `#FF9500` system orange | gradient blue → purple | `#FFC145` loopYellow |
| very high | `#AF52DE` system purple | `#FF9500` system orange | `#A15CE6` muted purple | `#FF9500` system orange |

dnzxy: @MikePlante1 I looked into this and ran it changed locally. My conclusion is that for the majority of users (those running dynamic) this will create a mismatch in what they see on the chart to what they see in the progress bar; so a mismatch directly on their home screen. 

I suggest we adjust the daily distribution charts (which I haven't seen many people use tbf), not the other way around.

We should use the same color codes throughout, I agree!

MikePlante1: I had the same thought this morning. I’ll PR an update for both distribution charts.

dnzxy: Thank you! 🙏

dnzxy:
Slightly adjusted the alarm button and some alignment after some feedback: 
image

Ready for review now.

marv-out (approved):

g5WS:
> If you have any suggestions to provide Statistics that you do not deem overwhelming or stress-inducing, I'm all for it. I'd have guessed an average glucose value and GMI or just a progress bar does that already. I could potentially see something like average glucose, total daily insulin and total daily carb consumption (for today, so since midnight, like the other displayed metrics), e.g. `ø 135 mg/dL · TDI 34 U · 103 g` `Today's averages and consumption`

I agree that the TIR data is likely to distress some users and will create conditions ripe for self-judgement.  For some users, it would be the equivalent of asking someone to step on a bathroom scale 20+ times a day — it would require a whole lot of self-acceptance not to find that data confrontational. I think it’s reasonable to predict that this implementation would worsen mental health in some users. 

I might suggest displaying an emoji based on a user-configured set of thresholds. Users could determine these with their healthcare team to define the thresholds that worked best for them. For example, the default could be:

95<=TIR<=100: 🏆
90<=TIR<95: 😁
70<=TIR<90: 🙂
50<=TIR<70: 😐
TIR<50: 😬

Personally, I would find autosense ratio and basal ratio to be the more helpful data here. Nothing predicts a site fail faster for me, and this information is not available on the home screen. 

I think average glucose, median glucose, CV and eA1c would be equally confrontational.  I additionally note that xDrip’s main screen allows users to hide all statistics.

kristinkand:
> > If you have any suggestions to provide Statistics that you do not deem overwhelming or stress-inducing, I'm all for it. I'd have guessed an average glucose value and GMI or just a progress bar does that already. I could potentially see something like average glucose, total daily insulin and total daily carb consumption (for today, so since midnight, like the other displayed metrics), e.g. `ø 135 mg/dL · TDI 34 U · 103 g` `Today's averages and consumption`
> 
> 
> 
> I agree that the TIR data is likely to distress some users and will create conditions ripe for self-judgement.  For some users, it would be the equivalent of asking someone to step on a bathroom scale 20+ times a day — it would require a whole lot of self-acceptance not to find that data confrontational. I think it’s reasonable to predict that this implementation would worsen mental health in some users. 
> 
> 
> 
> I might suggest displaying an emoji based on a user-configured set of thresholds. Users could determine these with their healthcare team to define the thresholds that worked best for them. For example, the default could be:
> 
> 
> 
> 95<=TIR<=100: 🏆
> 
> 90<=TIR<95: 😁
> 
> 70<=TIR<90: 🙂
> 
> 50<=TIR<70: 😐
> 
> TIR<50: 😬
> 
> 
> 
> Personally, I would find autosense ratio and basal ratio to be the more helpful data here. Nothing predicts a site fail faster for me, and this information is not available on the home screen. 
> 
> 
> 
> I think average glucose, median glucose, CV and eA1c would be equally confrontational.  I additionally note that xDrip’s main screen allows users to hide all statistics. 
> 
> 

I'm not a T1D (just a parent of a T1D), but I totally agree that even the progress bar without the % can be felt like a slap in the face, every time you look at the home screen.

At the same time I can also see that other users might use the progress bar as an encouragement to try harder. 

I think the option to toggle the progress bar off, would be a good option, so everyone chooses what fits best. 

Btw for deeming if a day is going better/worse compared to the previous one, I added a line on the home screen which shows the previous day bg. So, that we can look at it and ask ourselves: was today's lunch actually that bad or do I just have this impression? So much depends on the setting/mood etc

Anyway only my 🪙🪙

You guys always do such a great job! Thank you! 😊

dnzxy:
> I additionally note that xDrip’s main screen allows users to hide all statistics.

It's a multi-purpose panel still functioning as the accessor to Statistics (with a second path way now integrated through History view's toolbar items), so hiding it is not an option here, sorry. 

We've been discussing this a lot on Discord today and yesterday, too. I have suggested a "loop performance" display, e.g., a computed value between received glucose values, ran algorithm cycles and enacted algorithm cycles; which could then depict something like a "Time in Automation" or "AID Performance" (labels completely tbd). 

I'd personally still find today's insulin, i.e. TDI since midnight, and today's carbs, i.e. total carbs since midnight, a usefull metric.

g5WS:
This ticket is closed but the statistics issue remains unresolved. Is a fix coming?  

Multiple people have identified the emotional harm caused by being confronted with statistics on a daily basis. Let us reflect on the Hippocratic Oath: *first, do no harm*.

dnzxy:
This ticket is not a ticket but a pull request and got closed when it got merged. 

Something will be added, yes, but there are already alternative ls present today. 

No one here took the hippocratic oath.

PR #1328, opened 2026-07-16
[Part 3 of 7] Home Layout v2 – Add multi-use panel with prioritized states

dnzxy:
Part 3 of 7 of the home layout v2 series, stacked on `home-refactor-layout`.

Turns the stats banner slot into a multi-use panel: one slot, highest-priority state wins — notifications disabled (critical), pump time mismatch, stale/missing CGM (manual glucose entry sheet), Max IOB = 0 (deep-links to the setting with the search-result scroll + wiggle), and real time-in-range stats as the default face. TIR is computed from today's readings via the distribution helper extracted from the Stat module, honoring the user's time-in-range type.

Priority resolution is a pure function with unit tests (`MultiUsePanelStateTests`). Verified by forcing each state on the sims.

| No Notification Permissions | CGM Error / Signal Loss / Calibration | Pump Time Zone Shift | Max IOB 0 |


t1dude:
This is a good and future-proof approach. It will let us add more to the same panel later.

marv-out (approved):

PR #1329, opened 2026-07-16
[Part 4 of 7] Home Layout v2 – Add draggable chart with pinch zoom and pull-to-loop

dnzxy:
Part 4 of 7 of the home layout v2 series, stacked on `home-refactor-multiuse-panel`.

Makes the main chart interactive: offset-based panning with momentum, pinch zoom (2–24 h), double-tap zoom presets, hold-to-scrub selection popover, and pull-down-to-force-loop with an inline refresh indicator. The y-axis is pinned during pans, the basal pane draws inverted so bars hang from the top edge, and scheduled-basal recalculation follows profile changes.

Verified gesture-by-gesture on SE3, 17 Pro, and Pro Max with CGM/pump sim data.


EJM0:
New chart introduces a problem when a user tries to scroll upwards, outgoing from the header or blankspace around the bottom controls. Howl header and chart is scrollable. I think the intended behaviour is to stay fixed.

https://github.com/user-attachments/assets/362bbabd-fc3f-44cc-9c8d-4a81ebeb1dbe

t1dude:
> New chart introduces a problem when a user tries to scroll upwards, outgoing from the header or blankspace around the bottom controls. Howl header and chart is scrollable. I think the intended behaviour is to stay fixed.
> 

I don’t understand why this is an issue?

EJM0:
Before it was fixed, I thought it would be unintended behaviour. It feels off for the howl main view to be scrollable.

dnzxy:
I think we may have different assumptions for what a scrollable view is. It‘s kinda normal for it to be "moveable" 👀

EJM0:
Oh okay tough normal iOS behavior was to be fixed ... still learning :)

EJM0:
Ahhh sorry did not see the forceable loop feature

dnzxy:
Yes you can drag it around because of the looping behavior :)

EJM0:
Jup reading the pr name should be learned🤓

MikePlante1:
Should the glossary sheet be updated to include all the gestures of the home chart as well?

dnzxy:
I’d rather not change this, as the headline will no longer make sense but is fully translated and gestures are not really "shapes and elements".

We‘re open to adding that in there though, of course.

marv-out (approved):

PR #1330, opened 2026-07-16
[Part 5 of 7] Home Layout v2 – Extend chart history to 72h with windowed rendering

dnzxy:
Part 5 of 7 of the home layout v2 series, stacked on `home-refactor-chart-gestures`.

Extends chart history to 72 hours: all chart-feeding fetch controllers take span-aware predicates anchored at `chartHistoryStartDate`, re-anchored on foregrounding, with windowed rendering keeping only the visible span plus padding in the plot. The initially prototyped lazy backfill was dropped after real-device testing (visible load hitch) in favor of always loading 72 h. The TIR banner is labeled "today" to distinguish its daily scope from the 72 h chart.

Verified on device with 90 days of data and on the sims.

### DIFF
--- a/Trio/Sources/Helpers/MainChartHelper.swift
+++ b/Trio/Sources/Helpers/MainChartHelper.swift
@@ -42,7 +42,8 @@ enum MainChartHelper {
         /// How far back the chart's `startMarker` is anchored — the fixed 24 h
         /// history window loaded on every open. Independent of the currently
         /// visible viewport, which the user can pinch-zoom within this range.
-        static let chartHistorySeconds: TimeInterval = 24 * 3600
+        static let chartHistorySeconds: TimeInterval = 72 * 3600
+        /// Backfill triggers when the visible leading edge gets this close
         /// Visible x-axis window seeded on first launch of the chart (matches the old 6 h default).
         static let defaultVisibleSeconds: TimeInterval = 6 * 3600
         /// Tightest pinch-in zoom.
@@ -51,10 +52,19 @@ enum MainChartHelper {
         static let maxVisibleSeconds: TimeInterval = 24 * 3600
         /// Double-tap cycles the visible window through these presets.
         static let zoomPresets: [TimeInterval] = [6 * 3600, 12 * 3600, 24 * 3600]
+        /// Render window extends this many visible-windows beyond each visible edge.
+        static let renderWindowPadFactor = 1.5
+        /// Re-anchor when the visible edge gets within this fraction of a
+        /// visible-window of the render window's edge.
+        static let renderWindowMarginFactor = 0.5
         /// Geometric grid for pinch commits (~4 % per step). Every committed zoom step
         /// re-lays the full-width canvas, so this bounds a halving of the visible window
         /// to roughly 18 re-layouts instead of hundreds.
         static let zoomStepRatio: Double = 1.04
+        /// Live pinch previews as a transform; once the stretch drifts past
+        /// this ratio a crisp re-layout is committed mid-gesture, so the
+        /// distortion stays bounded.
+        static let pinchCommitScaleDrift: Double = 1.25
         /// How far (pt) a one-finger touch may travel and still count as a stationary
         /// press-to-inspect; beyond this the touch becomes a pan.
         static let inspectMovementTolerance: CGFloat = 10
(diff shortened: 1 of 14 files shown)


MikePlante1:
https://github.com/user-attachments/assets/3f2131d8-d039-4b8e-9aa8-f17ef6f769c0

Not a blocker, but scrolling with 72hr can get kinda choppy, even on an iPhone 17 Pro Max. I just zoomed in and flicked in left, waited for it to settle, then flicked it left again.

dnzxy:
I’m afraid this won’t be something we can address properly. That’s a render issue not even something we can change in a data driven way. I already removed a previous approach of lazy loading, as it didn’t do much :(

Trio/Sources/Helpers/MainChartHelper.swift
```diff
@@ -39,6 +39,42 @@ enum MainChartHelper {
...
     enum Config {
+        /// How far back the chart's `startMarker` is anchored — the fixed 24 h
+        /// history window loaded on every open. Independent of the currently
+        /// visible viewport, which the user can pinch-zoom within this range.
+        static let chartHistorySeconds: TimeInterval = 72 * 3600
+        /// Backfill triggers when the visible leading edge gets this close
+        /// Visible x-axis window seeded on first launch of the chart (matches the old 6 h default).
+        static let defaultVisibleSeconds: TimeInterval = 6 * 3600
+        /// Tightest pinch-in zoom.
+        static let minVisibleSeconds: TimeInterval = 3 * 3600
```
MikePlante1: I'd like this to be reduced to `1` hour to better visualize 1-min readings. For reference, LoopFollow zooms in even further to 10 minutes.

MikePlante1:
When a new CGM reading is populated in Trio, the graph automatically shifts to the far right. When zoomed in to 3 hours (the current minVisibleSeconds), the current glucose doesn't even get displayed. My suggestion would be to just automatically zoom out to 6 hours when a new reading shifts the graph all the way to the right, unless the current scale is already 6 hours or higher.

example:

Trio/Sources/Helpers/MainChartHelper.swift
dnzxy: Of course lowering this to 1 hour is easy, but I just want to note:
- We currently don't support 1-min readings
- I cannot see the use case for allowing you to zoom in to 10min on a 4320 min spanning chart. Do you truly need 72 1-hour segments?

dnzxy:
Will need to look into that. It may make more sense for UX to no longer "snap" to current time.

Trio/Sources/Helpers/MainChartHelper.swift
MikePlante1: > We currently don't support 1-min readings

Just an example of where this can be extra helpful. I don't think it needs to wait for 1-min readings to be added before this change is made. It doesn't seem to behave any worse for me when zoomed in to 1 hr versus 3 hr, so I don't see any harm in adding it now.

> I cannot see the use case for allowing you to zoom in to 10min on a 4320 min spanning chart. Do you truly need 72 1-hour segments?

Seems to be two things mixed into one here... 

1) I'm not advocating for a 10-min window in Trio, just pointing out that people have used a graph that zooms in that far for a long time now with LoopFollow. I've seen many comments from Trio users about them switching to LoopFollow to view their chart at times because they need to zoom in more to better read SMBs and such.

2) I won't need to look at the entire 72 hour chart one hour at a time... but I would like to be able to zoom into any specific hour, usually the most recent hour

MikePlante1: To expand on 2, though... I personally think for the main chart, being able to zoom in to 1 hour is more useful than being able to scroll back 72 hours, and the 72 hour extension seems to have made scrolling choppier anyway.

dnzxy: >  I've seen many comments from Trio users about them switching to LoopFollow to view their chart at times because they need to zoom in more to better read SMBs and such.

I haven't seen a single such report.

We can scale down 72 -> 48. It's not the amount of data, it's the amount of element and points in chart to render that's the issue here, because we need to recompute 3 charts all the time 🙃 

But yeah, generally speaking, there's 0 harm in moving this to 1 hr (or less). I was more so interested in the wider use case of it 🤓

dnzxy:
All addressed. Zoom now possible up to 1 hr, and the "jumping" now address the potentially very big forecast by moving to the now-marker with some space-buffer.

marv-out (approved):

PR #1334, opened 2026-07-18
Preserve non-manager CGM lifecycle state on Home

paulplant:
## Summary

Preserve CGM lifecycle and status state published by non-manager `GlucoseSource` implementations on the Home screen.

This supports non-direct CGM connections such as xDrip4iOS AppGroup, where Trio receives lifecycle information through the glucose source rather than through a native `CGMManager`.

## Why

Home has a periodic refresh that re-reads CGM lifecycle and status information. Before this change, that refresh only used a native `CGMManager`, with a simulator fallback. Non-manager sources can publish lifecycle and status through `GlucoseSource`, but the refresh path cleared those values because no native manager exists.

The Home view also used `cgmSensorExpiresAt` as part of the sensor ring display gate. Non-manager sources can provide valid lifecycle progress without a manager-derived expiration date, so Home now tracks when progress came from a source publisher and allows that source-origin progress to render.

## Changes

- Preserve source-published lifecycle and status during the Home refresh for non-manager glucose sources.
- Track whether lifecycle progress came from a non-manager source.
- Allow source-origin lifecycle progress to show the sensor ring without requiring a native manager expiration date.
- Keep native `CGMManager` and simulator paths separate.

## Compatibility

This remains fully backward compatible with xDrip4iOS versions before v7. Legacy AppGroup payloads that only write the top-level readings array continue to publish nil lifecycle and status state, so Trio does not fabricate a sensor ring.

Native CGM managers still refresh directly from `cgmLifecycleProgress` and `cgmStatusHighlight`. Native sources with `cgmSensorExpiresAt` still use the existing 48 hour display gate. Simulator behavior remains special-cased.

## Testing

Validated on-device with xDrip4iOS AppGroup data from the forthcoming xdripswift v7 update:
- [JohanDegraeve/xdripswift#730](https://github.com/JohanDegraeve/xdripswift/pull/730)

Tested cases:
- lifecycle ring displays from source-published state
- ring remains visible across Home refresh ticks
- legacy and no-lifecycle payloads do not fabricate ring state
- soon-to-expire `maxSensorDays` cases display correctly
- expired `maxSensorDays` cases display correctly

### DIFF
--- a/Trio/Sources/Modules/Home/HomeStateModel.swift
+++ b/Trio/Sources/Modules/Home/HomeStateModel.swift
@@ -498,31 +498,36 @@ extension Home {
                     // warmup / stabilizing / expiry. Simulator has no
                     // CGMManager, so fall back to reading its synthetic
                     // lifecycle / highlight so the bobble sees the same
-                    // data shape a real CGM would deliver.
+                    // data shape a real CGM would deliver. Other non-manager
+                    // sources publish their own state through `GlucoseSource`.
                     let manager = self.fetchGlucoseManager.cgmManager
                     let source = self.fetchGlucoseManager.glucoseSource
                     let progress: DeviceLifecycleProgress?
-                    let highlight: DeviceStatusHighlight?
+                    let displayState: CgmDisplayState?
                     if let manager {
                         progress = manager.cgmLifecycleProgress
-                        highlight = manager.cgmStatusHighlight
+                        displayState = manager.cgmStatusHighlight.map {
+                            CgmDisplayState(
+                                localizedMessage: $0.localizedMessage,
+                                imageName: $0.imageName,
+                                status: CgmDisplayStatus.from($0.state)
+                            )
+                        }
                     } else if let sim = source as? GlucoseSimulatorSource {
                         progress = sim.cgmLifecycleProgress
-                        highlight = sim.cgmStatusHighlight
+                        displayState = sim.cgmStatusHighlight.map {
+                            CgmDisplayState(
+                                localizedMessage: $0.localizedMessage,
+                                imageName: $0.imageName,
+                                status: CgmDisplayStatus.from($0.state)
+                            )
+                        }
                     } else {
-                        progress = nil
-                        highlight = nil
+                        progress = source?.cgmProgressHighlight.value
+                        displayState = source?.cgmDisplayState.value
                     }
                     self.cgmProgressHighlight = progress
-                    if let highlight {
-                        self.cgmDisplayState = CgmDisplayState(
-                            localizedMessage: highlight.localizedMessage,
-                            imageName: highlight.imageName,
-                            status: CgmDisplayStatus.from(highlight.state)
-                        )
-                    } else {
-                        self.cgmDisplayState = nil
-                    }
+                    self.cgmDisplayState = displayState
                     self.cgmSensorExpiresAt = Self.resolveSensorExpiresAt(
                         manager: manager,
                         glucoseSource: source,


dnzxy:
Thanks for this! The `HomeStateModel` fix is correct and welcome — the periodic refresh wiping source-published state on every tick is the actual bug here 👍

I'm not sold on the `cgmProgressComesFromSource` + `CurrentGlucoseView` gate change though. With the refresh fix alone, the existing gate (`cgmProgress?.progressState != .normalCGM`) already covers all your tested cases — warmup, soon-to-expire, expired. Your xdripswift side even marks the last 48 hrs as `warning` for exactly that reason. The only thing the view change adds is a permanently visible ring during normal mid-life for xDrip users, which diverges from how the other CGMs behave.

For reference: other CGMs display the lifetime ring once the sensor is within its last 48 hrs, but keep the state `normal` until 24 hrs remain, and only then switch to `warning` (`critical` for grace/expired). Compare G7SensorKit: https://github.com/loopandlearn/G7SensorKit/blob/0c87905297b70a7f0fa41e87fa0fbe1efff6c29f/G7SensorKitUI/G7CGMManager/G7CGMManager+UI.swift#L102-L127

I'd suggest dropping the view gate change here and instead having xDrip4iOS follow that same convention — only publish lifecycle within the last 48 hrs (plus warmup/expired), `normal` until 24 hrs, then `warning`. That gives full parity with the native managers, ring color included.

(Edit to add: comment written via LLM on my own notes, for ease of workflow.)

paulplant:
Thanks Dan, that makes sense. Looking back at PR #1205 , I do not think the exact display rules were completely clear. The later G7 review note about having `1d 12h` left and expecting an arc does suggest the ring was intended to appear before expiry, not only once the sensor is already expired or in warmup.

> I'd suggest dropping the view gate change here and instead having xDrip4iOS follow that same convention — only publish lifecycle within the last 48 hrs (plus warmup/expired), `normal` until 24 hrs, then `warning`. That gives full parity with the native managers, ring color included.

I agree the HomeStateModel refresh issue was the real bug. The 30-second timer should not wipe valid lifecycle/status state from a non-manager glucose source just because there is no native CGMManager so I've removed it.

A later discussion could really be about how rich non-manager sources should be allowed to be. xDrip4iOS can provide more lifecycle detail through the AppGroup feed than most CGM providers expose through their managers, so should we preserve that richer feed, or deliberately dumb it down to match the smaller _surface area_ available from those sources? (Although from your reply, I already got the answer 😂)

To keep this PR focused and following your comments, I have removed the view-gate change and left only the Home refresh fix. That keeps Trio’s existing display rules unchanged while still preventing source-published lifecycle/status state from being cleared between AppGroup updates.

dnzxy:
> Although from your reply, I already got the answer 😂

Always happy to help without helping 😂 

For now, we basically need 3 things from xDrip4iOS, which ultimately come down to a replication of what the LoopKit-style CGM managers expose:
- CGM Status Highlight
- CGM Lifecycle
- Specifically for Dexcom G6 / One, whether the paired CGM is an Anubis TX, so we can show correct warm-up period and lifecycle — this goes hand in hand with 1) and 2).

Subsequently, or rather, because we're seeking the above replication, an arc will always show, when a CGM lifecycle is exposed (and not `nil`), and the "tag" (the small label below the glucose bobble) will show, when a CGM Status Highlight is exposed. The Status Highlight can also provide an SF Symbol image-name which will then hide the glucose bobble in Trio and display that icon. 

In a later step, I'd love to collaborate on providing much more rich information so that we can build out the xDrip plugin "settings view" to display information similar to how other CGM manager settings views expose them.

dnzxy (approved):

PR #1336, opened 2026-07-19
Quick-Pick Treatments

t1dude:
**Iteration on Quick-Pick Boluses**

Adding the ability to log carbs based on frequently logged carb amounts, following the same logic as bolus amounts.

Notes:

1. Replaced the custom slider with the default slider that is also used by Omnipod and TAI
2. If there are no carbs in history, no carb buttons will be shown - respecting users who do not log carbs
3. Added "Updating IOB" "Updating COB" and "Updating COB and IOB" messages
4. Raising error messages if one or both actions fail
5. Adding a line of text to the (?) file pointing to the normal Treatments View if the user wants a recommendation or a calculation
6. Renamed all files, names, etc from QuickPickBolus to QuickPickTreatments (sorry for all the changes, but it's the only clean way to do this...)
7. Fixed the alignment issue with German translation of "U"

Needs testing and code review.

https://github.com/user-attachments/assets/aa658c35-95b1-4ce9-863f-56dfd7ee8965


t1dude:
Unit tests are failing, and gh says it's an issue unrelated to my PR:

> You're absolutely right. Looking at the PR changes, **you did NOT modify anything related to glucose level settings or the `FileStorageTests.swift` test**. 
> 
> Your PR only touches:
> - **QuickPickBoluses → QuickPickTreatments** renaming/refactoring
> - **Settings model** (`enableQuickBolus` → `enableQuickPickTreatments`)
> - **HomeStateModel** for bolus and carb suggestion logic
> - **UI components** for the quick-pick treatments sheet
> - **Localization strings**
> 
> The failing test `"Can parse mmol/L settings to mg/dL"` is **unrelated to your changes**. This appears to be a **pre-existing test failure** in the codebase.
> 
> The test failure indicates an issue with:
> - Settings parsing that converts mmol/L glucose values to mg/dL
> - The threshold_setting field not being parsed correctly
> - Expected value 100 mg/dL but got 60
> 
> **This is a baseline issue in the repository, not caused by your PR.** You can safely proceed—your changes don't touch any glucose settings parsing logic. The test failure exists independent of your Quick-Pick Treatments refactoring.
> 
> If you need to merge this PR, you may want to note this in the PR comments so reviewers understand the test failure is unrelated.

Trio/Sources/Modules/Home/HomeStateModel.swift
dnzxy, line 1: Let's split the quick pick logic into a file of its own via `Extension`

dnzxy (commented):
Brief review, looks good. Will test this a bit, then give an in-depth review.

We may need to revisit the copy for ease of understanding, but that's disconnected from the functionality.

t1dude:
Moving Quick-Pick Treatments into a separate file to avoid too much code in HomeStateModel.

Trio/Sources/Modules/Home/HomeStateModel.swift
t1dude: See commit 06c1851

t1dude:
> Brief review, looks good. Will test this a bit, then give an in-depth review.

Cool, thanks! I'm running it as well. Looking forward to in-depth review - feel free to DM if you want to chat about it.

> We may need to revisit the copy for ease of understanding, but that's disconnected from the functionality.

I'm not sure what "copy for ease of understanding" means.

Trio/Sources/Modules/Home/HomeStateModel+QuickPickTreatments.swift
```diff
@@ -0,0 +1,208 @@
...
+
+/// Scores samples by recency, time-of-day, and weekday/weekend similarity to "now", then returns the
+/// rounded amounts with the highest scores. Shared by the bolus and carb Quick-Pick Treatment suggestion
+/// loaders so the two amount types are ranked with identical logic.
+private func topQuickPickSuggestions(
+    from samples: [QuickPickSample],
+    roundingScale: Int,
+    limit: Int = 5
+) -> [Decimal] {
+    let now = Date()
```
dnzxy: If this becomes a parameter in the function, it becomes replay-able and deterministic.

Trio/Sources/Modules/Home/HomeStateModel+QuickPickTreatments.swift
```diff
@@ -0,0 +1,208 @@
...
+        .filter { $0.key > 0 && $0.value >= 0.1 }
+        .sorted { $0.value > $1.value }
+        .prefix(limit)
+        .map(\.key)
+}
+
+/// Fetches entities of `type` from the last `lookbackDays` days matching a cutoff-built predicate, and
+/// scores them into Quick-Pick suggestions. Shared by the bolus and carb suggestion loaders, which differ
+/// only in the entity type, predicate shape, and how a `QuickPickSample` is extracted from each entity.
+private func fetchQuickPickSuggestions<T: NSManagedObject>(
```
dnzxy, line 61: Just noting: I'm not sure this wrapper function here is worth it given that all we save is 2 fetches and some post-processing. The same could be achieved by just relying on our util function `CoreDataStack.shared.fetchEntitiesAsync()` and a subsequent map insight a perform closure. 

This does not break anything, and it works well, just noting.

Trio/Sources/Modules/Home/HomeStateModel+QuickPickTreatments.swift
```diff
@@ -0,0 +1,208 @@
...
+            predicate: { cutoff in
+                NSPredicate(
+                    format: "isSMB == false AND isExternal == false AND pumpEvent.timestamp >= %@",
+                    cutoff as NSDate
+                )
+            },
+            sortKey: "pumpEvent.timestamp",
+            roundingScale: 2
+        ) { bolus in
+            guard let nsAmount = bolus.amount, nsAmount.doubleValue > 0, nsAmount.doubleValue <= maxBolusUnits,
```
dnzxy: What does `ns` stand for in this context?

Trio/Sources/Modules/Home/HomeStateModel+QuickPickTreatments.swift
```diff
@@ -0,0 +1,208 @@
...
+        async let carbs = loadQuickPickCarbSuggestions()
+        let (bolusSuggestions, carbSuggestions) = await (boluses, carbs)
+
+        await MainActor.run {
+            quickPickBolusSuggestions = bolusSuggestions
+            quickPickCarbSuggestions = carbSuggestions
+        }
+    }
+
+    private func loadQuickPickBolusSuggestions() async -> [Decimal] {
```
dnzxy, line 106: This is a wrapper of a wrapper that's only used once. 
Not sure we need this level of encapsulation.

Trio/Sources/Modules/Home/HomeStateModel+QuickPickTreatments.swift
```diff
@@ -0,0 +1,208 @@
...
+            sortKey: "pumpEvent.timestamp",
+            roundingScale: 2
+        ) { bolus in
+            guard let nsAmount = bolus.amount, nsAmount.doubleValue > 0, nsAmount.doubleValue <= maxBolusUnits,
+                  let timestamp = bolus.pumpEvent?.timestamp else { return nil }
+            return QuickPickSample(amount: nsAmount as Decimal, timestamp: timestamp)
+        }
+    }
+
+    private func loadQuickPickCarbSuggestions() async -> [Decimal] {
```
dnzxy, line 126: This is a wrapper of a wrapper that's only used once. 
Not sure we need this level of encapsulation.

Trio/Sources/Modules/Home/HomeStateModel+QuickPickTreatments.swift
```diff
@@ -0,0 +1,208 @@
...
+            debug(.bolusState, "Quick-pick treatment bolus authentication error: \(error)")
+            return .failed
+        }
+    }
+
+    /// Enacts a Quick-Pick Treatment. Carbs are stored and the bolus is authenticated/enacted
+    /// concurrently, since neither depends on the other's outcome, so a carb-only or bolus-only pick
+    /// works the same as selecting both. Each half reports its own success/failure so the caller can
+    /// tell the user about a partial failure instead of silently losing one half of a combined pick.
+    func enactQuickPickTreatment(bolusAmount: Decimal?, carbAmount: Decimal?) async -> Home.QuickPickTreatmentOutcome {
```
dnzxy, line 192: Storing/enacting carbs and bolus run concurrently via async let.
The bolus half opens with unlockmanager.unlock(). So carbs are written to Core Data before the user authenticates, and cancelling Face ID still logs the carbs. 

This is not per se a bug, as treatment view state follows a similar workflow, but it's potetially worth making deliberate decision here to follow this flow for quick pick carbs, too. Some users may expect cancelling to abort the whole quick pick selection. If that's the desired behavior (abort all if any single action is cancelled or aborted),  we need to guard carb storage behind auth or authenticate first, then fan out, or handle errors / abort differently.

Trio/Sources/Modules/Home/HomeStateModel+QuickPickTreatments.swift
```diff
@@ -0,0 +1,208 @@
...
+        async let bolusResult = enactQuickPickBolus(bolusAmount)
+        var outcome = Home.QuickPickTreatmentOutcome()
+        (outcome.carbsResult, outcome.bolusResult) = await (carbsResult, bolusResult)
+
+        // A successful bolus already triggers its own determine-basal sync; sync here whenever carbs
+        // were saved and the bolus (if any) wasn't successfully enacted, matching
+        // Treatments.StateModel.saveMeal().
+        if outcome.carbsResult == .succeeded, outcome.bolusResult != .succeeded {
+            do {
+                try await apsManager.determineBasalSync()
```
dnzxy, line 203: If this here succeeds, we'll see a second determineBasalSync from APSManager (via bolus dose enactment) which could potentially get into a race with this call here.

Trio/Sources/Modules/Home/HomeStateModel+QuickPickTreatments.swift
```diff
@@ -0,0 +1,208 @@
...
+/// A single historical amount (bolus units or carb grams) used to build Quick-Pick Treatment suggestions.
+private struct QuickPickSample {
+    let amount: Decimal
+    let timestamp: Date
+}
+
+/// Scores samples by recency, time-of-day, and weekday/weekend similarity to "now", then returns the
+/// rounded amounts with the highest scores. Shared by the bolus and carb Quick-Pick Treatment suggestion
+/// loaders so the two amount types are ranked with identical logic.
+private func topQuickPickSuggestions(
```
dnzxy: This function would be worth to be covered by Max-Bolus/Max-Carb filtering, weekday/weekend weighting, recency decay, rounding, and empty history via unit tests. It's safety-relevant and very testable. At a minimum the cap-filtering deserves a / some testing.

To add testing, this would need to become `public`.

Trio/Sources/Modules/Home/View/QuickPickTreatments/QuickPickTreatmentsView.swift
```diff
@@ -0,0 +1,328 @@
...
+        .padding(.horizontal)
+    }
+
+    private var displayedBolusSuggestions: [Decimal] { Self.displayedSuggestions(from: bolusSuggestions, limit: 3) }
+    private var displayedCarbSuggestions: [Decimal] { Self.displayedSuggestions(from: carbSuggestions, limit: 3) }
+
+    /// Grows the sheet to fit however many pill rows are actually shown, rather than assuming both rows render.
+    private var sheetHeight: CGFloat {
+        let rowCount = (displayedBolusSuggestions.isEmpty ? 0 : 1) + (displayedCarbSuggestions.isEmpty ? 0 : 1)
+        return rowCount >= 2 ? 380 : 320
```
dnzxy: This hardcodes 320/380 by row count. With Dynamic Type or long localized labels (e.g., German would be something like "Schieben um zu protokollieren und abzugeben"), content could clip, likely will. Should probably be handled differently.

dnzxy (commented):
Overall good. I had some comments.

Trio/Sources/Modules/Home/HomeStateModel+QuickPickTreatments.swift
t1dude: Fixed in 29d3d80ca

Trio/Sources/Modules/Home/HomeStateModel+QuickPickTreatments.swift
t1dude: Yup, that's a good point. I think I'll keep it as is for easy error handling, but feel free to change it if OCD kicks in ;)

Trio/Sources/Modules/Home/HomeStateModel+QuickPickTreatments.swift
t1dude: Random prefix. Changed it for clarity in db4597f91.

Trio/Sources/Modules/Home/HomeStateModel+QuickPickTreatments.swift
t1dude: It's a preference I guess, and coming from SQL in the 2000's I prefer readability when browsing code. Still new to Swift. I could in-line this if you prefer?

Trio/Sources/Modules/Home/HomeStateModel+QuickPickTreatments.swift
t1dude: Same as above, really. Want me to in-line this?

Trio/Sources/Modules/Home/HomeStateModel+QuickPickTreatments.swift
t1dude: If we change this here, we should also change it in the treatment view for consistency. Leave it for a separate PR changing behavior in both use cases?

Trio/Sources/Modules/Home/HomeStateModel+QuickPickTreatments.swift
t1dude: I don’t see a race here, the outer sync is sequenced after both async lets join and is already mutually exclusive with the internal one as far as I can see. But I found something else: enactQuickPickBolus always returns .succeeded once the pump call was made. Fixed in 476c58a13.

Trio/Sources/Modules/Home/HomeStateModel+QuickPickTreatments.swift
t1dude: Changed from private to internal (doesn't seem to need to be public?). Didn't touch the cap-filtering since that's actually outside this function. Added tests with help from CC, please review them as well. de691c981

Trio/Sources/Modules/Home/View/QuickPickTreatments/QuickPickTreatmentsView.swift
t1dude: I have changed it to make it more dynamic, but man, that German translation example is hard to support :D And the shared slider package has some hight constraints as well. I'll do the best I can. Screenshot coming soon...

t1dude: German translation of QPT

t1dude: Improved (made more dynamic) in e2067886e

Trio/Sources/Modules/Home/HomeStateModel+QuickPickTreatments.swift
trioneer-dev: Please track this in an issue of its own, @t1dude . We’ll ship this as-is. 
Thanks for raising that it also concerns the “normal” treatment view path.

trioneer-dev (approved):
Reviewed the changes. LGTM.
Tested on a simulator. Let unit tests run. Compiles clean, runs successfully.

Let’s 🚢  this!

PR #1343, opened 2026-07-24
Make Watch-app & -complication follow ConfigOverride.xcconfig

LiroyvH:
The main app name can be overriden by either changing it in XCode or, easier for all future use, put an entry in ConfigOverride.xcconfig to overrule it.

However, this currently only affects Trio's main target as the rest is hardcoded to "Trio" within this file instead of referencing $(APP_DISPLAY_NAME). So the Watch app and Complication get set to "Trio" and ignores the ConfigOverride value.

This minor change makes XCode also look at ConfigOverride.xcconfig for the Watch and Watch Complication targets. When the file isn't present or does not contain a line to override APP_DISPLAY_NAME, it defaults back to "Trio" as defined in Config.xcconfig (it follows what's in that file when ConfigOverride.xcconfig isn't present or lacks the key). 

See comments for an additional remark.

### DIFF
--- a/Trio.xcodeproj/project.pbxproj
+++ b/Trio.xcodeproj/project.pbxproj
@@ -6147,7 +6147,7 @@
 				GCC_C_LANGUAGE_STANDARD = gnu17;
 				GENERATE_INFOPLIST_FILE = YES;
 				INFOPLIST_FILE = "Trio Watch Complication/Info.plist";
-				INFOPLIST_KEY_CFBundleDisplayName = Trio;
+				INFOPLIST_KEY_CFBundleDisplayName = "$(APP_DISPLAY_NAME)";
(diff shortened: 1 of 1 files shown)


LiroyvH:
Additional comment:
When testing this in an already existing clone, the XCode UI **_may_** sometimes not update the value displayed in "Display Name" in the target's "General" settings tab, but shows a cached version. It will show it properly in the Build Settings-tab and in my test it did still build it properly with the name set in ConfigOverride.xcconfig. 

This appears to be (a bit of an annoying) XCode display bug and may take 1 or 2 restarts of XCode for the UI to update the General tab. On a fresh clone (assuming you created the ConfigOverride.xcconfig before starting XCode) it will always immediately show it properly in the General tab as well.

dnzxy (approved):

PR #1344, opened 2026-07-25
Refactor to enable oref from CLI tools

kingst:
This change does a small refactor to enable us to run oref functions
from the CLI. By running oref functions from a CLI we can do more
extensive testing on the algorithm by replaying recorded inputs.

Any CLI programs we write will need to import the Trio package as
@testable for now. If we want to change that later, we'll need to add
`public` annotations to key functions and data structures to be able
to access them.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -47,7 +47,7 @@ let algorithmHelpers = [
 let package = Package(
     name: "TrioAlgorithm",
     defaultLocalization: "en",
-    platforms: [.iOS(.v17), .macOS(.v14)],
+    platforms: [.macOS(.v14)],
     products: [
         .library(name: "Trio", targets: ["Trio"])
     ],
--- a/Package.swift
+++ b/Package.swift
@@ -40,6 +40,7 @@ let algorithmHelpers = [
     "JSON",
     "Rounding",
     "String+Extensions",
+    "TherapySettingsUtil",
     "TimeInterval+Convenience"
 ].map { "Helpers/\($0).swift" }
 
--- a/Package.swift
+++ b/Package.swift
@@ -1,0 +1,71 @@
+// swift-tools-version:5.9
+import PackageDescription
+
+// Builds the oref algorithm as a standalone, macOS-capable module
+// so the algorithm test suite can run with `swift test`
+//
+// This is a "shadow" package: it compiles the *existing* files in
+// place rather than owning its own copy, so there is exactly one
+// copy of every source file and the Xcode app target keeps
+// compiling the same ones.
+//
+// Usage:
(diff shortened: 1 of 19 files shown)


Trio/Sources/APS/Extensions/DecimalExtensions.swift
```diff
@@ -1,10 +1,6 @@
 import Foundation
 
 extension Decimal {
-    func clamp(to pickerSetting: PickerSetting) -> Decimal {
```
dnzxy, line 4: I'm not sure why this is being removed here and added to [DecimalPickerSettings.swift](https://github.com/nightscout/Trio/pull/1344/changes#diff-b31b04b0859031c435d0286deb21539b2089d941b97fcfc7d0d2c46e9c06ea89).
It's a clamping extension for Decimal, in a file called DecimalExtensions.

Trio/Sources/Models/DecimalPickerSettings.swift
```diff
@@ -143,6 +143,16 @@ struct DecimalPickerSettings {
     )
 }
 
+extension Decimal {
```
dnzxy: See other comment. Not sure I understand this port.
I see, the docstring points to it, still not sure this is a good move, with all the other cleanup this PR does.

Package.swift
```diff
@@ -0,0 +1,71 @@
...
+    "JSON",
+    "Rounding",
+    "String+Extensions",
+    "TimeInterval+Convenience"
+].map { "Helpers/\($0).swift" }
+
+let package = Package(
+    name: "TrioAlgorithm",
+    defaultLocalization: "en",
+    platforms: [.iOS(.v17), .macOS(.v14)],
```
dnzxy: Minimum build target for Trio is iOS 18.0

dnzxy (approved):
Overall good. Few small questions.

Trio/Sources/APS/Extensions/DecimalExtensions.swift
kingst: I moved it to a file called DecimalExtensions+SwiftUI.swift to keep with the convention I used for the other changes.

Package.swift
kingst: It should only run on MacOS so I removed the iOS statement.

kingst:
@dnzxy I addressed your comments, and I added a run of `swift test` to CI to catch errors that might happen since this is a separate pipeline. I tested the CI changes on my own repo: https://github.com/kingst/Trio/actions/runs/30288721064

dnzxy (approved):

PR #1346, opened 2026-07-25
Fix duplicate remote meal being re-stored on replay

### LINKED ISSUE #1348: Remote meal duplicate guard logs a rejection but stores the carbs anyway
## Describe the bug

`handleMealCommand` has a guard that rejects a remote meal command when newer carb entries already exist. The guard logs "Command rejected: newer carb entries have been logged since the command was sent" but never aborts the handler, because its `return` is inside a nested `Task {}` in `taskContext.perform {}`. The carbs are stored regardless, so a replayed or resent command doubles COB.

## To Reproduce

1. Set up LoopFollow remote control against a Trio phone
2. Send a remote meal command, then send another a few seconds later (or log a carb entry on the Trio phone between the two)
3. The log shows the rejection message for the second command, but both carb entries appear in Trio and Nightscout

## Expected behavior

A rejected command stores nothing.

## Additional context

Ran into this while testing remote meal entry for our daughter and firing test commands close together. The guard's intent is clearly right, the early exit just doesn't reach the handler. Fix in #1346.

JustMaier:
## Summary

The guard in `handleMealCommand` that rejects a remote meal when newer carb entries already exist never actually aborts the handler. Its `return` sits inside a nested `Task {}` inside `taskContext.perform {}`, so it only exits that closure. The handler logs "Command rejected" and then stores the carbs anyway.

This computes the fetch result via `await taskContext.perform { ... }` returning a `Bool` and guards from the handler itself, so the rejection stops storage.

## Background

Noticed while running Trio with LoopFollow remote control for our young daughter. If a meal command is delivered twice (APNs redelivery, or a caregiver resend after an unclear failure), the duplicate carbs are stored and dosed for, while the log claims the command was rejected.

I'm @justmaier on the Trio Discord if it's easier to discuss there.

Fixes #1348

### DIFF
--- a/Trio/Sources/Services/RemoteControl/TrioRemoteControl+Meal.swift
+++ b/Trio/Sources/Services/RemoteControl/TrioRemoteControl+Meal.swift
@@ -44,17 +44,15 @@ extension TrioRemoteControl {
             ), key: "date", ascending: false
         )
 
-        await taskContext.perform {
-            guard let recentCarbEntries = results as? [CarbEntryStored] else { return }
-            if !recentCarbEntries.isEmpty {
-                Task {
-                    await self.logError(
-                        "Command rejected: newer carb entries have been logged since the command was sent.",
-                        payload: payload
-                    )
-                    return
-                }
-            }
+        let hasNewerCarbEntries = await taskContext.perform {
+            (results as? [CarbEntryStored])?.isEmpty == false
+        }
+        if hasNewerCarbEntries {
+            await logError(
+                "Command rejected: newer carb entries have been logged since the command was sent.",
+                payload: payload
+            )
+            return
         }
 
         let actualDate = payload.scheduledTime.map { Date(timeIntervalSince1970: $0) }
--- a/Trio/Sources/Services/RemoteControl/TrioRemoteControl+Meal.swift
+++ b/Trio/Sources/Services/RemoteControl/TrioRemoteControl+Meal.swift
@@ -37,10 +37,13 @@ extension TrioRemoteControl {
 
         let payloadDate = Date(timeIntervalSince1970: payload.timestamp)
         let taskContext = CoreDataStack.shared.newTaskContext()
+        // Only entries already in the past can indicate a replay; equivalents and scheduled
+        // meals are dated ahead and would otherwise reject every command until their date passed.
         let results = try await CoreDataStack.shared.fetchEntitiesAsync(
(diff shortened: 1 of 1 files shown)


bjorkert:
Thanks for finding and solving this!

Now that this guard actually runs, it will also match future-dated entries, like FPU carb equivalents or a meal scheduled ahead of time. Any of those would reject remote meal commands until their dates pass.
I'd bound the predicate to entries that already happened, e.g. date > %@ AND date <= %@ with the payload date and now. Replays are still caught, since the duplicate's entry lands between send time and delivery.

What do you think?

JustMaier:
Good catch. That would bite harder than it looks. With the default 60 minute delay and 30 minute spread the last FPU equivalent sits two hours out, so any meal with fat or protein would block remote meals for that whole window. Four hours at the max settings.

Your bound fixes it. I'll change the predicate to `date > %@ AND date <= %@` with the payload date and now. Replays still get caught for the reason you gave, the duplicate's entry lands between send and delivery.

Scheduled-ahead meals are the same shape so that covers those too.

JustMaier:
Pushed. Predicate is now `date > %@ AND date <= %@` with the payload date and now.

Worth flagging that this does give one case up. A replay of a meal scheduled for later stores its entry at the scheduled date, so the upper bound excludes it and the duplicate is no longer caught. It needs the redelivery to land inside the 600 second command window, so it's narrow, and I think it's the right trade against blocking every remote meal for two to four hours after any meal with fat or protein. There isn't a clean way to have both, since `CarbEntryStored` only has the entry date and no creation timestamp.

Happy to look at an idempotency key on the payload separately if you think that's worth doing.

dnzxy:
We‘ve discussed adding more date properties to carb entries in the past. The swift struct representation carries two dates.

JustMaier:
Yeah, that would be the clean fix. The struct has both, but `CarbEntryStored` only persists `date`, so there's nothing to ask whether an entry was stored after the command was sent. With a `createdAt` on the entity the guard stops looking at entry dates at all, and the scheduled meal case I mentioned goes away with it.

I'd rather not grow this one into a model migration though. The unbounded guard is live today and blocks remote meals for hours after any meal with fat or protein. Happy to do the `createdAt` change as a follow-up if you want it.

Was there a conclusion from that earlier discussion I should read first?

trioneer-dev (approved):

PR #1347, opened 2026-07-25
Do not enact a meal's explicit bolus when the meal is rejected

### LINKED ISSUE #1349: Remote meal's explicit bolus is enacted even when the meal is rejected
## Describe the bug

For a remote meal command that includes `bolusAmount`, the `.meal` dispatch stores the meal and then enacts the bolus whenever `bolusAmount != nil`. `handleMealCommand` reports validation failures by logging and returning rather than throwing, so a rejected meal (for example carbs over the configured max) still delivers its bolus. The caregiver gets a rejection notification and has no reason to think insulin went through.

## To Reproduce

1. In Trio's remote settings, set max carbs to a low value (e.g. 30g)
2. From LoopFollow, send a meal above the cap with a bolus amount attached
3. The meal is rejected with a notification; pump history shows the bolus delivered anyway

## Expected behavior

If the meal half of the command is rejected, the bolus half should not be enacted. The existing validation was clearly meant to gate the whole command.

## Additional context

Found while reviewing the remote meal path (same testing as #1348). Fix in #1347.

JustMaier:
## Summary

For a remote meal command with `bolusAmount`, the `.meal` dispatch calls `handleMealCommand` and then enacts the bolus whenever `bolusAmount != nil`. Since `handleMealCommand` reports validation failures by logging and returning (it doesn't throw), a rejected meal still delivers its bolus. Carbs over the configured max is the easiest way to hit it.

This moves the bolus decision inside `handleMealCommand`, after the meal has validated and stored, so an early return also suppresses the bolus. The standalone `bolus` command is unchanged.

## Background

The max-carbs check exists to catch fat-fingered remote entries, but today it rejects the carbs and delivers the insulin anyway. The caregiver sees a rejection notification and has no reason to think a bolus went through. Found while reviewing the remote meal path for our daughter's caregiver setup.

I'm @justmaier on the Trio Discord if it's easier to discuss there.

Fixes #1349

### DIFF
--- a/Trio/Sources/Services/RemoteControl/TrioRemoteControl+Meal.swift
+++ b/Trio/Sources/Services/RemoteControl/TrioRemoteControl+Meal.swift
@@ -68,12 +68,15 @@ extension TrioRemoteControl {
 
         try await carbsStorage.storeCarbs([mealEntry], areFetchedFromRemote: false)
 
-        if payload.bolusAmount == nil {
-            await logSuccess(
-                "Remote command processed successfully. \(payload.humanReadableDescription())",
-                payload: payload,
-                customNotificationMessage: "Meal logged"
-            )
+        if payload.bolusAmount != nil {
+            try await handleBolusCommand(payload)
+            return
         }
+
+        await logSuccess(
+            "Remote command processed successfully. \(payload.humanReadableDescription())",
+            payload: payload,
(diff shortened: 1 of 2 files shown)


dnzxy (approved):
Approving based on code review and after having discussed this with the LoopFollow team :) 

Thank you for finding this, and promptly providing a fix to it.

PR #1353, opened 2026-07-28
Fix the mac build script

kingst:
Previously there was a Package.swift file at the root directory of the
repo, so calling `xed .` opened that package instead of
Trio.xcworkspace, causing issues for the build script. This change
moves the package to a directory.

This commit was done collaboratively with @dnxzy


LiroyvH (commented):
Appears to work with 'xed .' again and builds OK!

dnzxy (approved):

PR #1362, opened 2026-07-31
Escape literal percent in Swedish exercise-mode translation

sjolundjohn:
Xcode fails to compile the string catalog with:

```
Unable to create a Swift type from the format specifier "%-i"
```

It's in the Swedish translation of the exercise-mode help text ("Set to a number, e.g. 160, …"), where "50%-ig" ("fifty-percent") happens to read as a `%-i` format specifier. Escaping the percent as `%%` fixes the parse, and the string renders exactly the same to Swedish users.

The string came in with the "Merge translations into a String Catalog" migration, and it's the only occurrence of the pattern in the catalog — I checked all 2,557 keys across every localization and variation.

### DIFF
--- a/Trio/Sources/Localizations/Main/Localizable.xcstrings
+++ b/Trio/Sources/Localizations/Main/Localizable.xcstrings
@@ -242019,7 +242019,7 @@
         "sv" : {
           "stringUnit" : {
             "state" : "translated",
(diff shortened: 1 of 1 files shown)


trioneer-dev:
@t1dude can you double check that this fix makes it back to Crowdin please?

trioneer-dev (approved):

PR #1365, opened 2026-07-31
Update TreatmentsRootView.swift to fix auto-scrolling

LiroyvH:
Fixes issue #1115.
The treatments menu would, when tapping the bolus screen, not scroll up properly on some devices. It would fall behind the keyboard.

PR #1238 by @Sjoerd-Bo3 fixed this in an acceptable manner, but that version is outdated now and cannot be merged with the current version of Trio due to changes to this very menu, causing build errors.

This PR re-used @Sjoerd-Bo3 code and was rewritten to make it compatible again. Like Sjoerd's code, this was generated with Claude. So far, it seems to work fine and as expected on an iPhone 16 Pro and an iPhone 11. Both running iOS 26.6 and synced to the latest dev branch.

### DIFF
--- a/Config.xcconfig
+++ b/Config.xcconfig
@@ -19,7 +19,7 @@ TRIO_APP_GROUP_ID = group.org.nightscout.$(DEVELOPMENT_TEAM).trio.trio-app-group
 
 // The developers set the version numbers, please leave them alone
 APP_VERSION = 0.8.4
-APP_DEV_VERSION = 0.8.4.67
+APP_DEV_VERSION = 0.8.4.68
 APP_BUILD_NUMBER = 1
 COPYRIGHT_NOTICE =
 


Trio/Sources/Modules/Treatments/View/TreatmentsRootView.swift
```diff
@@ -202,190 +202,225 @@ extension Treatments {
...
+        private func scrollFocusedFieldIntoView(_ field: FocusedField?, proxy: ScrollViewProxy) {
+            guard let field else { return }
+
+            // Fat and Protein share a single row, which is tagged with `.fat`.
+            let targetID: FocusedField = (field == .protein) ? .fat : field
+
+            // Delay slightly so the keyboard's presentation animation has begun and the List has
+            // resized its scrollable area before we scroll, otherwise the target can land back
+            // underneath the keyboard.
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.3) {
```
trioneer-dev: Two notes:
- Hard-coded timeouts like this one are a bad practice. 
- If this is “supposed to” be happening in the background, unnoticed to the user, this needs to be <= 200ms, as anything >200ms stops being recognized as instantaneous to humans.

LiroyvH: @trioneer-dev 
Thanks for the check! Ok, so the hard-coded wait was added - according to the AI - to address a problem where not waiting for the iOS keyboard animation to complete could make the bolus field end up behind the keyboard again if it triggers too soon. It chose 0.3 because that's roughly how long the keyboard takes to come up and finish its "slide up" animation. However, it agreed it was a sub-optimal solution and made changes:

Track keyboard visibility via UIResponder.keyboardWillShowNotification / keyboardWillHideNotification. Then:
1.) If the keyboard is already up (switching between fields) → scroll immediately, no delay.
2.) If the keyboard is currently animating in (first tap into the screen) → trigger the scroll off that notification (which also hands you the real animation duration/curve iOS is using) instead of a guessed constant.

The modified code passed an AI four-eyes check and I'm now testing the results myself, will submit an update to the PR once testing is complete. Looks good so far. :) 


Last but not least, an extra comment: I dug through everything and made the AI's do the same and I can't find any reason why a modification to TreatmentsRootView.swift could trigger the unit test error that was thrown in my PR.

trioneer-dev: Sounds like a good addition, @LiroyvH .
There are also a few adjustments landing in Loop around keyboard behavior.
Potentially worth a look: https://github.com/tidepool-org/LoopKit/pull/749

LiroyvH: @trioneer-dev Thanks! :) I've made some updates because of recent changes to this file in dev and it appears to be working fine in production so far on latest dev - happy camping. Scrolls exactly where it should be, no weirdness, no bugs. I'll have a look at those LoopKit files to see if the approach ought to be altered (and whether that's for later care or immediate). If not: I'll try committing the changed file tomorrow, actually need to figure out *how* to add it to the existing PR. :|

LiroyvH: @trioneer-dev Well, the way I just tried to get to it apparently resulted in a force-push that killed this PR, so err... I'll open a new one. Sorry :X

Issue #1351, opened 2026-07-26
Improve Accessibility: make the core screens usable with VoiceOver, low vision, and reduced transparency


dnzxy:
## Is your feature request related to a problem? Please describe.

Trio has effectively **no accessibility support today**. Across the entire app there are only a handful of `accessibility*` modifiers (a few tab-bar labels, one "Cancel bolus" label, one reduce-motion check). Nothing uses Reduce Transparency, Increase Contrast, or chart audio descriptors.

Because Trio is used to manage insulin dosing, this is safety-relevant for a portion of its user base:

- **VoiceOver users** cannot reliably read current glucose/trend, cannot interpret the recommendation math, and in at least one place **cannot perform the action at all** — the Quick-Pick "slide to confirm" deliver control is a custom drag with no accessible equivalent.
- **Low-vision users** hit translucent Liquid Glass panels that can fall below readable contrast, and state that is conveyed by color only (glucose range, dose-component sign, limit-exceeded warnings).

This issue tracks a prioritized, phased effort to close those gaps on the core screens: **Home, the treatment/bolus flow, and the core Settings screens.**

## Describe the solution you'd like

Fix in priority order, safety of the insulin path first, and prefer shared components so we fix many screens at once. Nothing here changes dosing behavior — we only **add accessible representations** of existing information and **supplement** (never replace) the current warning/limit cues.

### Phase 0 — Adopt basic iOS accessibility capabilities for Liquid Glass adoption 
- [x] Respect `accessibilityReduceTransparency` and `colorSchemeContrast` in `GlassChrome.swift` (and the glass tab bar): fall back to an opaque, higher-contrast fill instead of glass/blur. All glass panels inherit it.
- [x] Establish the color-independence rule for the codebase: any state conveyed by color must also appear in text or an accessibility trait. Honor `Differentiate Without Color` where a glyph is the fallback.
- [x] (No change) Dynamic Type stays capped at `.xxLarge` for now; may extend to `.xxxLarge` later.

### Phase 1 — Improve shared components / controls
- [x] `Trio/Sources/Views/SettingInputSection.swift` — `accessibilityElement(children: .combine)` per row; label the icon-only "?" hint button from the existing `hintLabel`; expose value+unit as `accessibilityValue`. Covers Units & Limits, Bolus Calculator, SMB, Algorithm Advanced, most of Meal Settings, and ~20 modules.
[...]

PR #1332, opened 2026-07-16
[Part 7 of 7] Home Layout v2 – Enable Liquid Glass and adapt home chrome

dnzxy:
Part 7 of 7 of the home layout v2 series, stacked on `home-refactor-sensor-arc`.

Turns off `UIDesignRequiresCompatibility`: real Liquid Glass on iOS 26 via a shared `GlassChrome` (glass panels with rim strokes, capsule and circle variants) with ultraThinMaterial fallbacks below 26. The adjustments card, multi-use/TIR banners, and bolus progress panel move onto the shared chrome; the adjustments card keeps split tints and a bicolor rim for concurrent override + temp target. The tab bar becomes icon-only with an invisible width-holding middle slot and the overlaid treatment button (tap = treatment sheet, long-press = quick bolus), centered via a small UIKit probe; the pre-26 tab bar is unchanged. The chart's minimum height drops to fit the taller iOS 26 bar geometry on SE-class devices.

Verified light/dark on SE3 + 17 Pro (iOS 26.5), Pro Max, and iPhone 16 Pro on iOS 18.5 for the legacy path.

| 17 Pro iOS 26.4 | SE 3 iOS 26.5 | 16 Pro iOS 18.5 |
|--------|--------|--------|
| image | image | image |

PR #1366, opened 2026-08-01
Introduce Accessibility Improvements 

trioneer-dev:
## Summary

This PR introduces broad accessibility improvements  across the entire app, focused on VoiceOver support, Dynamic Type limits, and Reduce Transparency / Increase Contrast fallbacks. The goal throughout is to make status, controls, and charts that currently rely on color, icons, or gestures readable and operable for assistive-technology users.

The main areas covered:

- **Status readouts spoken in words** — loop pill, pump reservoir/battery/pod, IOB/COB, alarms, eventual glucose, and Nightscout/Tidepool connection state now announce their meaning instead of color and icon names. The glucose bobble reads value, range, trend, delta, age, and sensor warmup/expiry as one element.
- **Controls made operable via VoiceOver** — slide-to-confirm quick bolus (previously drag-only, which blocked delivery), stop-adjustment / stop-temp-target / stop-override cancels, tappable header panels, and picker rows across settings, onboarding, statistics, alarms, and therapy editors.
- **Grouped, labeled content** — recommendation-math cards, bolus-progress readouts, stat summaries, and settings value rows read as single coherent phrases; icon-only buttons (help, hint, add/remove, dismiss, time nudge, etc.) get spoken labels; decorative icons and row dots are hidden.
- **Charts named** — glucose, forecast, basal, carb-ratio, ISF, target, calibration, and the twelve statistics charts get VoiceOver labels.
- **Dynamic Type bounded app-wide** — capped at accessibility1 globally (tighter at xxLarge on the dense Statistics screen) so large text scales without breaking layouts.
- **Glass chrome fallbacks** — glass panels and small material affordances fall back to an opaque, higher-contrast fill under Reduce Transparency or Increase Contrast.

## Additional notes

This PR depends on #1332 and should not be merged until that lands.
Addresses #1351. 

## Reviewing

This is a rather large PR (**16 commits, 85 files**), but each commit is self-contained and scoped to a specific area or screen. **Please review commit-by-commit** rather than viewing the full diff. That keeps each change in context and makes the scope far easier to follow.


kingst (approved):
For a large mechanical PR like this it's very difficult to check the entire thing by hand. I checked a few patterns and they look good, and I do think that this is a meaningful improvement. So I'm going to approve it.

I did run it through an LLM code review that was trained on my code reviews :) The results are below and that flagged some issues that are optional. There does look like at least one real issue, the rest, I'm not sure.

====

Required

1. ForecastChart.swift:212 speaks the oldest glucose as "latest".

if let latest = state.glucoseFromPersistence.last {

TreatmentsStateModel.swift:140 sorts that controller ascending: false, so .last is the oldest reading in the fetch window — the state model itself uses .first for the latest (TreatmentsStateModel.swift:507). This is the bolus calculator screen, so a VoiceOver user sizing a dose hears a stale number. Should be .first.

Worth noting the trap: Home.StateModel.glucoseFromPersistence sorts ascending: true (HomeStateModel.swift:163) and Treatments.StateModel.glucoseFromPersistence sorts ascending: false. Same property name, opposite order. MainChartView.swift:213 gets it right by luck of being on the Home model. Can we make one of them a named accessor (latestGlucose) so callers stop having to know?

2. Units aren't spelled out in the two most-spoken strings.

- CurrentGlucoseView.swift:220 — \(units.rawValue), so the home-screen bobble announces "mg/dL". The doc comment three lines up literally says "Glucose 79 mg/dL, in range, falling…". Should be units.spokenValue.
- ForecastChart.swift:207 — String(localized: "mg/dL") instead of state.units.spokenValue.

Both are in the same PR that added GlucoseUnits.spokenValue (BloodGlucose.swift:309) for exactly this.

3. OnboardingRootView.swift:436 — 0.6s timer to move VoiceOver focus.

private func announceScreenChange() {
    DispatchQueue.main.asyncAfter(deadline: .now() + 0.6) {
        UIAccessibility.post(notification: .screenChanged, argument: nil)
    }
}

Any time I see a magic delay it makes me nervous unless we understand the failure mode exactly. On a slow device 0.6s isn't enough and focus lands on the old content; if the user taps Next twice quickly, two posts land and fight each other; and it fires for every user, VoiceOver running or not. What was the actual race — the scroll animation, or the step content not being in the hierarchy yet? If it's the latter, posting from the new step's .onAppear, or binding @AccessibilityFocusState to the step id, gets you the same result deterministically. At minimum gate on UIAccessibility.isVoiceOverRunning.

Separately, this is now called from eight identical onChange handlers (:409-424) with the two statements crammed onto continuation lines. One handleStepChange() that does both would read better.

4. UnitSpelling is keyed on English but fed localized strings.

BloodGlucose.swift:281 maps "mg", "dL", "U", "hr"… but TherapySettingEditorView.swift:314 passes unit.displayName, which is String(localized: "mmol/L/U") / String(localized: "U/hr") / String(localized: "g/U") (:358-373). In any locale that translates those abbreviations, every token misses the table and passes through unchanged — VoiceOver reads the abbreviation, which is the thing this PR is trying to fix. Feed it the untranslated form, or move a spokenName onto TherapySettingUnit next to displayName.

Also, English grammar: "g/U" → "grams per units" and "mmol/L/U" → "millimoles per liter per units". The "U" token needs a singular form when it's the denominator.

5. Plurals: "1 minutes ago".

TimeAgoFormatter.swift:33 formats "%d minutes ago" unconditionally, and this is spoken on the home screen every minute. Same shape in PumpView.swift ("%1$d days %2$d hours", "%d hours", "%d minutes"), CalibrationsChart ("%d calibration points"), and HomeRootView+MealPanel ("snoozed, %d minutes remaining"). There are no plural entries in Localizable.xcstrings today, so this needs either "Vary by Plural" in the catalog or ^[\(n) minute](inflect: true).

6. AdjustmentsRootView+TempTargets.swift:232-233 — non-tappable rows announce as buttons.

The hint at :227 is correctly conditional on onTap != nil, but the trait and the action right below aren't. A read-only row gets .isButton and an activation that does nothing. Make all three conditional. (And Text("") for the nil hint — better to not attach a hint at all than to attach an empty one.)

---

Design questions

7. Why cap Dynamic Type in an accessibility PR?

TrioApp.swift:347 caps the whole app at accessibility1, and StatRootView.swift:55 caps Statistics at xxLarge — which is below every accessibility size, so that screen now ignores accessibility text scaling entirely. That's a deliberate reduction aimed squarely at the users this PR is for. I understand dense charts break, but was ViewThatFits / switching to a stacked layout at AX sizes considered instead? If the cap stays, I think it needs a rationale in the PR description, because it's the one change here that takes something away.

8. Hints that describe the gesture.

Five hints say "Double tap to …" (AdjustmentsRootView.swift:199, AdjustmentsRootView+Overrides.swift, +TempTargets, QuickPickTreatmentsView). VoiceOver already appends "Double tap to activate" to anything with .isButton, so those read as "Double tap to edit. Double tap to activate." — and the gesture is simply wrong for Switch Control, Voice Control, and braille users. The rest of this PR gets it right ("Opens a picker to change this value", "Opens pump settings"). Worth making all of them result-phrased.

9. Two switches that have to stay in sync, in three files.

displayString / accessibilityValueString now exist as near-identical twins in SettingInputSection.swift:229/250, DeliveryLimitsStepView.swift:116/136, and AlgorithmSettingsSubstepView.swift:306/331. They already differ in shape between files. One function is enough:

private func valueString(for setting: PickerSetting, decimalValue: Decimal, spoken: Bool) -> String

with the unit picked as spoken ? UnitSpelling.spoken("U") : String(localized: "U"). Then the display and spoken forms can't drift.

10. Every .accessibilityAction duplicates its .onTapGesture body verbatim.

HomeRootView+BottomControls.swift:229/256/287/455, HomeRootView+Header.swift:45/73/104, MealSettingsRootView, UserInterfaceSettingsRootView, AlarmBGSection, AlarmGramsSection… roughly ten copies of the same closure written twice. Somebody will fix a bug in one copy and not the other. A small modifier would collapse the whole class:

func tappable(label: LocalizedStringKey, hint: LocalizedStringKey? = nil, action: @escaping () -> Void) -> some View

wiring .contentShape + .onTapGesture + .isButton + .accessibilityAction from a single closure.

11. The header's two panels are treated differently.

pumpView (HomeRootView+Header.swift:73) gets .accessibilityElement(children: .combine) + .isButton + hint + action. cgmView right above it (:45) gets only the two actions — no .combine, no .isButton, no hint. Was that deliberate (because CurrentGlucoseView already collapses itself), or an oversight? Related: CurrentGlucoseView.swift:92 adds .isButton to the "Add CGM" empty state while the actual gesture lives on the parent.

12. MainChartView.chartAccessibilitySummary runs on every body pass.

:213-227 does a filter + map + min + max over the full chart history on the hottest view in the app — the body re-evaluates continuously while the user scrubs the chart, VoiceOver running or not. Consider UIAccessibility.isVoiceOverRunning as an early out, or computing it in the state model when the render window changes.

Two accuracy nits in the same property: the label says "Visible range" but renderWindowStart is the render window, which is wider than the viewport; and the visible.isEmpty fallback reports the entire history under that same "Visible range" wording. Also $0.date ?? .distantPast — I'd rather see compactMap on the date than a sentinel that silently sorts to the far past.

---

Minor

- [minor] ToggleStyles.swift:24,58 — you add .isButton + an On/Off value. Is there a reason not to use .isToggle (iOS 17+)? That's the trait that makes VoiceOver announce it as a switch and handle the state properly. Also Text("On")/Text("Off") are bare literals here while everything else in the PR uses String(localized:).
- [minor] Spoken units still abbreviated in several new strings, even though the helper now exists: PumpView.swift:201,203 ("U" → VoiceOver says "you"), HomeRootView+MealPanel.swift:31,55 (" g", " U"), AdjustmentsRootView+TempTargets.swift:224 ("min"), and formattedGlucose (AdjustmentsRootView.swift:48) which appends units.rawValue. And PopupView / TreatmentsRootView hardcode String(localized: "units") rather than going through UnitSpelling.spoken("U") — works, but that's now three ways to say the same thing. Also: is "\(percent) %" in batteryAccessibilityValue read correctly, or should it go through UnitSpelling.spoken("%") the way SettingInputSection does?
- [minor] CurrentGlucoseView.swift:217 — glucoseValue == 400 as the "high" sentinel, duplicating the same literal at :241. Pull it into a named constant.
- [minor] IconSelection.swift:1545 — iconDisplayName splits raw asset ids, so trioColorBG → "Trio Color B G", trio3D → "Trio3 D", and the wilford case announces "Diabeetus". These also never get translated. A displayName on Icon_ (Models/Icons.swift:5) with String(localized:) is cleaner and localizable.
- [minor] GlassChrome.swift:83-90 — GlassMaterialFill re-implements reduceTransparency || colorSchemeContrast == .increased instead of sharing prefersOpaque. And only HomeRootView.swift:99 was converted; TrioModalAlertScheduler.swift:275,404 and SettingsRootView.swift:356 still use raw .ultraThinMaterial. The alert scheduler is the glucose-alarm UI — leaving that translucent under Reduce Transparency seems like the one you'd most want to fix. Deliberate?
- [minor] AlarmGramsSection.swift:31 — Int(NSDecimalNumber(decimal: valueGrams).intValue); intValue is already Int.
- [minor] OnboardingRootView.swift:713 — let chapter = …; return chapter, just return the expression. Also it announces the chapter but not the step, while the bar itself shows step progress.
- [minor] PumpView.swift:76 and AdjustmentsRootView+TempTargets.swift:187 have pre-existing force unwraps (.string(from:)!) directly adjacent to the accessibility code you added, which correctly uses ?? "0". Worth killing while you're in these files.

---

Localization

~131 new String(localized:) calls and ~106 new Text("…") literals, and nothing under Trio/Sources/Localizations/ changed. I see the repo lands these in separate l18n commits (e1fc93ed9, ae3816152, 802cd305c), so this may just be the bot's job — can you confirm, so translators actually get the ~200 new strings rather than every non-English user hearing English?

Also worth a native-speaker's eye: several of the new spoken strings are assembled by concatenating localized fragments — AdjustmentsRootView+TempTargets.swift:221-226 (value + "for" + n + "min" + %) and TherapySettingEditorView.swift:314 (value + unit + "starts at" + time). Word order differs across languages; a single format string with positional arguments would translate correctly.

trioneer-dev:
@kingst Addressed the must-fixed. Please re-approve, then lets get this in :)

t1dude:
@trioneer-dev please throw me a Discord pm when this moves to DEV, as I need to coordinate a Crowdin update and put our translators to work :)

t1dude (approved):
LGTM

PR #1367, opened 2026-08-02
Refactor: Dependency Injection Hygiene

trioneer-dev:
## Summary

This PR tidies up dependency injection and adds tests to guard the container wiring.

- Dependency injection cleaned up by dropping unused `FetchGlucoseManager` injections and the orphaned `Router` resolve in `DeviceDataManager`, and dedupes eager resolves in `loadServices`. `TrioRemoteControl`, `TelemetryClient`, and `TelemetryAttestor` are now registered container-scoped instead of self-resolving static singletons, which removes the lazy-inject lock ceremony they required.
- Adds container resolution specific tests to check resolve of every eagerly loaded service and assert that the converted singletons stay identical across resolves. A missing or mis-scoped registration now fails in CI rather than at launch.

Blocked by #1300.


t1dude:
I guess this PR is just waiting for #1300. I'll mark it as a merge candidate to get the PR list filtered for release.

jwoglom (approved):
looks OK - I’m not super familiar with Swift DI patterns, but getting rid of the injectIfNeeded hack seems like an improvement.

dnzxy:
Merging w/ James' approval :)

PR #1374, opened 2026-08-03
Stop Crashlytics Uploads Immediately After Diagnostics Opt-Out

### LINKED ISSUE #1372: App connects to firebase when sharing is disabled
Trio v0.8.4.40 dev c171bd5

I disabled app diagnostics sharing immediately after installing v0.8.4.40.  Approximately 10 minutes later, Apple’s App Privacy Report logged communication from Trio to firebaselogging-pa.googleapis.com. 

Is this expected behavior?  It is concerning. 

(Also, users who previously deliberately set their analytics to “Disable Sharing” may feel violated and disrespected to have this setting changed surreptitiously. Please: make informed consent the standard.)

trioneer-dev:
`setCrashlyticsCollectionEnabled(false)` only gates future events. The FirebaseSessions event GoogleDataTransport already queued stays on disk and GDT retries it for seven days, so firebaselogging-pa.googleapis.com still shows up in App Privacy Report minutes after opting out.

Route both call sites through `CrashReportingGate: purge` GDT's store when a Trio user opts out, and skip `FirebaseApp.configure()` on any later launches so nothing can re-queue or upload queued sessions. 

Trio’s opt-out-by-default stays unchanged.

Fixes #1372.


MikePlante1 (approved):
Approved by code review.

PR #1375, opened 2026-08-03
Fix Alarm  Sounds and Low-Soon Alerts; add AlarmKit for >= iOS 26

### LINKED ISSUE #1371: [Notifications] No sounds are triggered
## Describe the bug
The alarms trigger a notification fine, but the sound never plays. (Despite being configured to play, silence on/off makes no difference, volume triple checked to not be on 0, and no focus mode active) 

## Attach a Log
Available upon request for one event. (Can collect more later, was first testing the waters to see if it was something I did (eg: misconfig) or if it always fails)

## To Reproduce
Alarms are configured:
:: Critical: Enabled, Active during Day & Night, Sound On, Override Focus Enabled, Play Sound: Enabled, Tone: Synth. 
:: Time Sensitive: Enabled, active during Day & Night, Sound On, Override Silence & Focus Mode disabled, Play Sound: Enabled, Tone: Synth

In any case, no alarm sound is played. The alarm sound plays OK when pressing the play button within the app. The app has all required permissions. It didn't play sounds for time-sensitive such as reservoir low, nor does it play a sound for the critical "patch empty" alarm. I do not have Critical Alarm entitlements from Apple. Note that a notification *is* triggered and with sound enabled you hear the default iOS sound. Also in-app the banner shows up properly. It just doesn't play the alarm sound.

## Expected behavior
Play the selected alarm sound when a problem occurs and the correct conditions are met. (In any case the Critical one should have played as it ignores pretty much everything.) 

## Screenshots
Not really relevant, though I did screenshot the in-app banner just in case somehow this is useful beyond a description in text.

### Smartphone:
iPhone 16 Pro sporting iOS 26.6.

### Pump:
Medtrum Nano 200U

### CGM:
Dexcom G6 (Anubis) running through xDrip4iOS 

### Trio Version:
0.8.4.35, dev f9199ed, Nightscout/Trio (Dev) 

## Technical Details
No further information in addition to what has already been written down.

## Additional context
None.

### LINKED ISSUE #1428: Low Glucose Soon notification not being triggered
## Describe the bug
Not getting expected low alert even though option is enabled.

## Attach a Log
Will attach in a bit from laptop 

## To Reproduce
1. Go to Notification and enable 'low glucose soon' alert 

## Expected behavior
I expect to be alerted when glucose is expected to be lower than my setting in next 20 minutes.

## Screenshots
Will attach 
image

image

### Smartphone:
* Hardware: iPhone 14 Pro
* OS Version: iOS 26.6.1

### Pump:
* Manufacturer: Insulet
* Model: *[e.g. Omnipo 5

### CGM:
* Device: Dexcom G7
* Manager app: Dexcom App

### Trio Version:
* Version Number: v0.8.4.63
* Repo: nightscout/trio
* Git Reference: Trio v0.8.4.63 (95) dev 38f3b36

## Technical Details
nothing I can think of

## Additional context
n/a

trioneer-dev:
Alarm sounds never reached the user. Four independent causes, one per commit.

Fixes #1371
Fixes #1428

### Causes

1. Device Alarms config (tone, Play Sound, Override Silence & Focus, day/night window) stopped being applied to catalog-known alerts in `00bdcb84b`; only the interruption level survived the switch to catalog routing. Plugins issue alerts with `sound: nil`, so device alarms had no tone at all.
2. Tones shipped as an Xcode folder reference, so they built to `Trio.app/Sounds/`. `UNNotificationSound(named:)` resolves only against the bundle root and `Library/Sounds`, so iOS silently substituted the default sound. The picker preview and `CriticalAlertAudioPlayer` worked because they were the only paths passing `subdirectory: "Sounds"`.
3. The critical-audio fallback, which is what pierces the silent switch and Focus without the Critical Alerts entitlement, rejected anything not `.immediate`. The not-looping watchdog is the only `.delayed` alert Trio creates, so the one alarm that fires unattended never got audio.
4. `evaluateForecast` bailed on the same CGM-ownership guard as the reading path, so Low Glucose Soon was never evaluated on Dexcom, xDrip4iOS or Libre, where "Use CGM App Alerts" defaults on.

### Commits

| Commit | Change |
| --- | --- |
| `86061e0` | Register the 18 `.caf` files individually so they flatten to the bundle root; drop the now-wrong `subdirectory:` from both lookups. Vendor sounds move to a flat `Library/Sounds/<manager>-<file>`. |
| `1420ca6` | Start playback from the modal scheduler's timer, whose `.insert` decision already means live, not stale, not snoozed. `.immediate` still plays at issue time. |
| `22dbce0` | Drop the CGM-ownership guard from `evaluateForecast`. A CGM app cannot compute Trio's oref forecast, so it can never cover this alarm. `evaluateCarbsRequired` already omits it, and the settings screen only moves reading-driven types into "Handled by CGM App". |
| `cdc46d2` | AlarmKit for critical alarms on iOS 26+, ported from Loop's `CriticalAlertAlarmScheduler`. Needs no entitlement and still fires when iOS has suspended Trio. Plays the user's selected tone. Stop acknowledges the alert rather than only silencing it. |

Earlier commit `690f24f0` (opt-in Critical Alerts entitlement) was dropped from this branch.

### Coverage

| Case | Before | After |
| --- | --- | --- |
| Critical, immediate, app running | default iOS sound | selected tone, pierces silent + Focus |
| Critical, delayed (not-looping) | silent | selected tone at fire time |
| Critical, app suspended, iOS 26+ | not possible | AlarmKit, selected tone |
| Critical, app suspended, iOS 17-25 | not possible | still not possible without the entitlement |

### Testing

Full suite green, 565 tests / 82 suites. New tests assert every `AlarmSoundCatalog` filename resolves via `Bundle.main.url(forResource:withExtension:)`, the call shape iOS uses, so a folder reference cannot silently regress this again.

Not covered by tests, needs a device: tone actually playing, piercing Silent and Focus, AlarmKit ringing on a suspended app, Stop acknowledging the alert. The AlarmKit custom tone in particular is unverified here.


LiroyvH:
@trioneer-dev I think it still doesn't work. I was waiting for a pump alarm to happen, but due to very heavy sensor noise I've got multiple "Trio is not looping" alarms. According to the app, this is supposed to trigger a critical alert. Whilst the push notification is triggered and the banner is shown in-app: no sound has been playing. The phone was on silent, but override is enabled. 

What I've done now is change the sound to something else and change it back to the intended one, just in case some setting didn't make it in initial configuration (stemming from before this PR). Also took the phone out of silent but I can only do that for some time as I want to suppress all other sounds when at work. Logs have been saved.

LiroyvH:
@trioneer-dev 
Alright, so, two alarms triggered today:
- Critical: Trio not looping for 20 mins
- Time-sensitive: Low reservoir

In different configs. Sound on, sound off (both have override enabled), checked volume. It just doesn't play the sound. The default iOS notification plays and that's it. (Note that with other apps like xDrip4iOS, the default iOS sound does not play (including when an alarm without sound is set))

As such I'm afraid to report that this doesn't appear to be fixed. I have generated new logs today if needed.

trioneer-dev:
Thanks! We’ll investigate and – hopefully 🤞 – come back with a good fix.
Appreciate the rigid testing!

trioneer-dev:
Moving this back to draft until the reported gaps/issues could be evaluated and addressed.

t1dude (approved):
Tested in-vivo for 24h on iOS 27. Works as expected.

PR #1382, opened 2026-08-07
Fix IOB area stacking + plot projected IOB decay

trioneer-dev:
### Fix: IOB shading rendered above the COB area

**Reported on Discord:** IOB plot fill was drawn above the COB chart.
**Reason:** The COB/IOB pane draws its marks with fixed styles and explicit series: identities (introduced with the home-refactor chart rework). Swift Charts stacks AreaMarks that share an x value by default, and series: does not opt out of stacking — so the IOB area was offset upward by the COB value whenever COB > 0, floating above the COB area instead of hugging the IOB line. Line marks never stack, which is why only the shading was wrong. Fixed by declaring stacking: .unstacked on both area marks. 

### Feature: projected active insulin in the COB/IOB pane

oref already computes a full IOB projection every loop cycle, up to 48 points at 5 minute cadence, 4 hours ahead (the same array whose activity values feed the glucose forecasting in determine basal). Trio already persists this to `monitor/iob.json` each loop cycle. This PR surfaces it:

  - `IOBService` (which already watches the iob file) exposes the array as `iobProjection`
  - `HomeStateModel` mirrors it alongside `currentIOB`
  - the COB/IOB pane draws it as a dashed line with a faint area from "now" into the future, anchored at the newest determination's `deliverAt`
  - Any stale projection (older than newest determination) renders nothing, and points beyond the chart window are dropped

Tested in the simulator with CGM + pump sims: carbs on board plus SMB-driven IOB confirm the shading now stays inside the IOB line with COB present, and the dashed decay curve renders from the current-time marker to the chart edge.

### Screenshots
trio-iob-chart-home 

trio-iob-chart-pane-closeup |


bjornoleh:
Thanks for adding IOB decay projection. 

Looks like the (i) overlaps a bit too much here to properly see the future IOB. Also, when checking this with sim pump/CGM, I got this rather short glimpse into the future. Possibly this is related to using the cone forecasts. More visible time ahead would be needed to make the future IOB projection readable, and this would help withvthe overlapping (i) button too. 

Perhaps this varies from loop to loop?

image

bjornoleh:
On a later loop with COB, there is more visible time into the future:

image

The auto scale determined by COB here does however makes it hard to interpret future IOB. Maybe the example with 1 U and 50 g COB is a bad one if normal carb ratios are considered. Here is a screenshot with more IOB:

image


And the IOB area stacking seems to be fixed 😊

trioneer-dev:
Thanks for testing this @bjornoleh

The forecasted IOB decay is a projected value that (will) adapt(s) with every subsequent loop cycle, as the AID doing “its job” will continue dosing &/or withholding insulin, which (intentionally) adapts the decay projection. the certainty of anything you see beyond the shortest forecast curve (which is what the cone is capped with) is fairly low, which is why I’d argue that any potentially overlap with the glossary button is highly negligible.

Speaking of the glossary button, I will push a follow-up PR that changes its style so that it more clearly distinguishes from the chart background in both light mode and dark mode views.

bjornoleh:
The first screenshot only displays one hour into the future. That's a bit short, and while the forecast curves does change a lot from loop to loop, the IOB projection will be a lot more consistent across loops, unless significant bolusing takes place of course. 

I would think the only actual use case for the IOB projection is to visually judge what IOB might look like a couple of hours ahead when planning for exercise. In this scenario one would probably set a high TT and not take another Bolus in a while. 

Does the projected IOB take an active (potentially 2 hours) zero temp into consideration?

trioneer-dev:
> The first screenshot only displays one hour into the future. That's a bit short, and while the forecast curves does change a lot from loop to loop, the IOB projection will be a lot more consistent across loops, unless significant bolusing takes place of course.

> I would think the only actual use case for the IOB projection is to visually judge what IOB might look like a couple of hours ahead when planning for exercise. In this scenario one would probably set a high TT and not take another Bolus in a while.

I’d argue that anything beyond 12-14 cycles is not worth putting a lot of confidence on, as it will change another 12-14x :) If you have high IOB now and are planning exercise, you will take countermeasures, just like you have been prior to this PR.

> Does the projected IOB take an active (potentially 2 hours) zero temp into consideration?

The projected IOB is the exact same that oref uses. So IOB module runs, and then its projected values are used to driev the forecast generator, and, subsequently oref’s dosing engine. We just plot this set of projected IOB values now. And FWIW / FYI, PR #1394 adds a plot of the projected COB decay as it happens in oref’s forecast generator.

bjornoleh:
Devils advocate: if your confidence in these projections is this low, why add any of this to clutter up the Home Screen?

Sorry for being dense, but I am still unsure about what I asked before, is currently active zero temps reflected in the IOB decay?

trioneer-dev:
> Devils advocate: if your confidence in these projections is this low, why add any of this to clutter up the Home Screen?

You are misunderstanding something replied here about the confidence of the output, with how the algorithm and more importantly the “self correcting” aspects of oref work. Forecasting has never been the true strength of any system; it’s how it recovers from dosing decision and potential “mistakes". 

We are adding it, because users ask for it. 
 
> Sorry for being dense, but I am still unsure about what I asked before, is currently active zero temps reflected in the IOB decay?

As said before, the IOB module calculates an active insulin projection, which then drives the forecasting generator, which then drives the dosing engine.

MikePlante1 (approved):
After the latest changes following private discussion, light code review and testing look good.

bjornoleh:
My question was: does a zero temp (or high/low temp basal for matter) affect the projected IOB for the duration of the temp basal? The future IOB trajectory will be very different if programmed basal is delivered, versus an already active two hour zero temp basal, or any other temp basal that significantly deviates from the programmed basal.

I see this is already merged, which is fine by me, but still curious about the above.

PR #1391, opened 2026-08-09
Fix Override Edit ISF/CR Picker Label

### LINKED ISSUE #1385: Override label changes from “Also Inversely Change” to “Also Change” when editing
## Describe the bug
when editing an override, the label “Also Inversely Change”

### Trio Version:
 Version Number: Trio v0.8.4

## Additional context
Hi, when editing an override, the label “Also Inversely Change” changes to “Also Change”. This makes it unclear whether the selected value will now change in the same direction as the basal adjustment, or whether it remains an inverse change.

Image

Image

trioneer-dev:
Align ISF/CR picker label in override edit form

Fixes #1385


MikePlante1 (approved):
Very simple fix. ✅

PR #1392, opened 2026-08-09
Fix Disappearing Target Line on Home Layout v2

### LINKED ISSUE #1383: Glucose target line disappears after midnight
Glucose target line disappears after midnight. Discussion on Discord, “I suspect that the issue is that we only "fetch" for 24 hr for the target thus it‘s only there for that one day and we now plot 3 days. Given that the profiles are just a 24 hr set of rates (or targets) this is probably just a simple computing issue... It‘s probably connected to https://github.com/nightscout/Trio/pull/1357 and the home refactor. “

On Trio Dev, 0.8.4.46, iOS 26.6, iPhone 16 Max Pro. 

Image

[log 3.txt](https://github.com/user-attachments/files/30858162/log.3.txt)

trioneer-dev:
Span glucose target profiles across full chart window.

Fixes #1383


t1dude (approved):
Code looks good, tested on 13 Mini test device.

PR #1393, opened 2026-08-09
Introduce Custom Liquid Glass-Styled Action Sheet as Confirmation Dialog Replacement

### LINKED ISSUE #1386: UI: Active button for overrides appears to be greyed out
Trio v0.8.4.48 dev bd9f20a

Perhaps consider a different colour scheme on this button? A button with a gray background gives an “inactive” appearance.  Also, the light purple on gray isn’t high contrast enough for clear readability. 

Image

trioneer-dev:
Add glass action sheet, replace confirmation dialogs:

- classic bottom action sheet with Liquid Glass panels
- opaque fallback when Reduce Transparency is on
- dim shows instantly, only the sheet slides
- converts all 15 confirmationDialog call sites

| Header | Header | Header |
|--------|--------|--------|
| image | image | image |

Fixes #1386

### DIFF
--- a/Trio/Sources/Views/GlassActionSheet.swift
+++ b/Trio/Sources/Views/GlassActionSheet.swift
@@ -117,7 +117,7 @@ private struct GlassActionSheetView: View {
 
     var body: some View {
         ZStack(alignment: .bottom) {
-            Color.black.opacity(dimmed ? 0.25 : 0)
+            Color.black.opacity(dimmed ? (reduceTransparency ? 0.4 : 0.25) : 0)
                 .ignoresSafeArea()
                 .onTapGesture { dismiss(then: onCancel) }
 
@@ -190,7 +190,7 @@ private struct GlassActionSheetView: View {
 
     @ViewBuilder private var panelBackground: some View {
         if reduceTransparency {
-            panelShape.fill(Color(.secondarySystemGroupedBackground))
+            panelShape.fill(GlassChrome.opaqueFill)
         } else if #available(iOS 26.0, *) {
             panelShape.fill(.clear).glassEffect(.regular, in: panelShape)
         } else {
--- a/Trio/Sources/Views/GlassActionSheet.swift
+++ b/Trio/Sources/Views/GlassActionSheet.swift
@@ -1,0 +1,212 @@
+import SwiftUI
+
+/// One tappable row of a `glassActionSheet`.
+struct GlassSheetAction: Identifiable {
+    let id = UUID()
+    let label: Text
+    let role: ButtonRole?
+    let handler: () -> Void
+
+    init(_ title: LocalizedStringKey, role: ButtonRole? = nil, handler: @escaping () -> Void = {}) {
+        label = Text(title)
+        self.role = role
+        self.handler = handler
+    }
+
+    /// Verbatim title, e.g. device or preset names that are not localization keys.
+    init(_ title: String, role: ButtonRole? = nil, handler: @escaping () -> Void = {}) {
+        label = Text(title)
+        self.role = role
+        self.handler = handler
+    }
+}
+
+extension View {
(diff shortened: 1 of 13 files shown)


munchyk:
**Edit: Skipped the part where you mentioned the issue you're trying to fix 😅**

<strike>Why not use the native swiftUI confirmation dialogs/popups/popovers? 

Note: the "popup - long list" image is a redacted screenshot. the "cancel" button is out of scroll view due to the list length.<strike>

| popover | popup | popup - long list |
| --- | --- | --- |
| VZzz5Uth | images | IMG_9132 |

trioneer-dev:
@munchyk iOS 26/27 and Liquid Glass have changed the old “Action Sheet” layout. Previously, it used to open very similar to the current PR proposal: a fullscreen cover, and a half-sheet-like dialog with a cancel button on the bottom and other action button(s) as the sheet’s content. The new layout is essentially a popup-style dialog that opens wherever the user prompted the action triggering the dialog, so a) placement is off b) readability is often off, and lastly c) context is often off. 

This PR introduces a (custom) alternative, that closely aligns two things:
1) the old, pre-Liquid Glass style of the Action Sheet
2) the new design that comes with Liquid Glass, i.e., half-transparent glass background, rounded shapes, etc.

munchyk:
Understood!

Missed the part about you reverting back to the old style as a fix (with the addition of a custom Liquid Glass implementation). 

Thought this was intended as an update away from the old style to the newer Liquid Glass styling.

trioneer-dev:
> Thought this was intended as an update away from the old style to the newer Liquid Glass styling.

@munchyk apologies for the confusion. Current `dev` holds the Liquid Glass default confirmation dialog; the one looking like a popover-like dialog, not the previous sheet-like dialog. 

It looks like this
image
image

This PR attempts to move back to the old UX, while keeping the new Liquid Glass style :)

munchyk:
@trioneer-dev yes! 

I’m on latest dev. Didn’t consider that it could be a reversion so thought it was a proposed update for an older UI element I missed.

trioneer-dev:
@munchyk It's not reverting Trio’s adoption of Liquid Glass by any means. The compiled app conforms to / displays all its elements with Liquid Glass. A reversion would be to set `UIDesignRequiresCompatibility` to `YES` again; however Apple is deprecating this build flag with the release of iOS 27. 

So, this PR is replacing the Swift default `.confirmationDialog()` with a custom view, that brings the old – pre iOS 26 – user experience and layout of `.confirmationDialog()`, but makes it be displayed in a Liquid Glass like style. It does this by introducing a custom view modifier `.glassActionSheet()` that holds a custom view `struct`, `GlassSheetAction()`.

Trio/Sources/Views/GlassActionSheet.swift
```diff
@@ -0,0 +1,212 @@
...
+                Button {
+                    dismiss(then: action.handler)
+                } label: {
+                    action.label
+                        .font(.title3)
+                        .foregroundStyle(action.role == .destructive ? Color.red : Color.accentColor)
+                        .frame(maxWidth: .infinity, minHeight: 57)
+                        .contentShape(Rectangle())
+                }
+                .buttonStyle(.plain)
```
EJM0, line 168: For further growth of cgm options or other bigger options, shouldn’t this be in a scroll view to prevent clipping of elements that don't fit the screen?

trioneer-dev: Good catch. CGM and pump selection will likely have to change into something that makes more sense, scrolling in a half-sheet is usually very buggy. Ideally something to address in a follow-up PR. This one here only proposes to change from default to custom dialog.

trioneer-dev: Will be addressed via #1439 .

AndreasStokholm (approved):
Code review looks good

PR #1394, opened 2026-08-09
Plot Projected COB Decay in Home View Chart 

trioneer-dev:
## Summary

> [!IMPORTANT]
> Depends on #1382 — this branch is based on it, so the diff shows its commits until it merges. Will be rebased onto dev afterwards; only the last commit is new here.

Twin of #1382's IOB projection, for carbs: the COB pane now shows a dashed orange decay curve of forecasted carb absorption from "now" into the future.

### Decay Calculation 
The series is derived inside forecastCOB from the exact carb-impact terms that already makes up the COB-specific forecast.
The linearly decaying observed carb impact plus the remaining-carbs triangle, divided by CSF, gives grams absorbed per 5-minute step, 48 values make up a 4 hours forecast.
Subtracting from mealCOB (floored at 0) yields remaining COB per step.
TODO: In a follow-up PR, refactor this to core data so per-cycle outputs become replayable.

### Additional Notes
This is oref's own (crude) absorption model (linear carb impact decay  and “triangle”-like instant decay, with a 30 g/hr cap), recomputed each cycle from observed deviations. The curve subsequently reshapes between algorithm runs. It shows what the algorithm currently “believes", not necessarily physiological prediction.

image


MikePlante1 (approved):
Merging this after private discussions, code-review, and testing.

Two things to consider for a follow-up PR:

1) Clip the COB projection line when it hits 0 g. This would match how no projection line is shown when current COB is 0 g, and also make it easier to see when the COB is projected to be depleted.

2) Max Meal Absorption Time is currently not considered in the COB projection line

PR #1396, opened 2026-08-09
Match Chart Glossary Button to Alarm Button Style in Meal Panel

trioneer-dev:
## Summary

The info button in the bottom-right of the main chart was styled independently of the alarm bell in the meal row directly below it, building on liquid-glass-esque material style. Especially on light mode this made it quite indistinguishable from the chart background. This PR aligns the glossary button to the bell button and moves it up slightly, to stay fully clear of char x-axis.

No behavior change. The button still toggles `state.isLegendPresented`.


t1dude (approved):
Code looks good. Tested on 13 Mini in light mode, looks good.

PR #1397, opened 2026-08-09
Show Release Notes for Installed Trio Version

trioneer-dev:
## Summary

Surfaces Trio's release notes in two places:

- **Home** — a "New Release" entry in the multi-use panel, opening a sheet the user acknowledges to dismiss it
- **Settings → Release Notes** — the current release, plus a **Previous Versions** screen for earlier ones

## Where the notes come from

Three sources, in order of preference:

1. A live fetch of the GitHub releases API, since notes are sometimes edited after publishing
2. The last successful fetch, cached
3. A copy bundled at build time by `scripts/capture-release-notes.sh`

The bundle covers **every stable release back to the last major or minor bump**, building `0.8.4` embeds `0.8.0` through `0.8.4`, so the list still works on a phone that has never been online.

That script can never fail a build. A missing release, an unreachable GitHub, an unparseable version or an empty environment all simply leave the fallback out; every path exits 0, and a stale bundle from a previous version is removed rather than left behind.

Drafts, prereleases, anything newer than the running build, and anything before `0.8.0` are all excluded. Sorting is by version.

## What the sheet renders

The release's opening prose, its GitHub alert callouts, and the bullets under **"What's Changed At A Glance"**. The full notes, including the pull request lists, stay on GitHub behind a link.

Two things worth knowing, both found by testing the parser against real releases rather than assuming a format:

- **The "At A Glance" heading is not universal.** Five of the six most recent releases have it; the `0.8.3` hotfix used "What happened" / "What's fixed" instead. Relying on that heading alone would have produced an empty sheet on exactly the kind of urgent hotfix people most need to read, so the parser falls back to the release's opening prose.
- **Releases carry `[!IMPORTANT]` callouts with safety-relevant content** — *"If you are using Medtrum Nano please update at your earliest convenience"*, *"your active Pod will be automatically transitioned"*. A bullets-only sheet would have hidden these behind a browser link, so they are parsed and shown.

## Acknowledgement

Stored per version, so installing a newer release surfaces the panel again with nothing needing to reset it. Only the sheet's button acknowledges — swiping it away, or browsing notes from Settings, leaves the panel in place.

The panel ranks below every warning state and above the stats. `MultiUsePanelState.resolve` gained a parameter, so its existing tests were updated and three added for the new ordering; all 10 pass.

### Screenshots

| ☾ Release Panel | ☀ Release Panel | Settings Row | Release Notes |

### DIFF
--- a/Trio/Sources/Modules/Settings/View/SettingsRootView.swift
+++ b/Trio/Sources/Modules/Settings/View/SettingsRootView.swift
@@ -36,6 +36,8 @@ extension Settings {
         )
         @State private var closedLoopDisabled = true
         @State private var showCopiedToast = false
+        @State private var showReleaseNotes = false
+        @ObservedObject private var releaseNotesService = ReleaseNotesService.shared
 
         @Environment(\.colorScheme) var colorScheme
         @EnvironmentObject var appIcons: Icons
@@ -289,6 +291,30 @@ extension Settings {
                         }
                     ).listRowBackground(Color.chart)
 
+                    if let notes = releaseNotesService.notes {
+                        Section(
+                            header: Text("Latest Release"),
+                            content: {
+                                Button {
+                                    showReleaseNotes = true
+                                } label: {
+                                    HStack {
+                                        Text(
+                                            "What's new? v\(notes.version)",
+                                            comment: "Settings row opening the release notes; the placeholder is a version number"
+                                        )
+                                        .foregroundColor(.primary)
+                                        Spacer()
+                                        Image(systemName: "chevron.right")
+                                            .foregroundColor(.secondary)
+                                            .font(.footnote)
+                                    }
+                                }
+                                .frame(maxWidth: .infinity, alignment: .leading)
+                            }
+                        ).listRowBackground(Color.chart)
+                    }
+
                     Section(
                         header: Text("Trio Backup"),
                         content: {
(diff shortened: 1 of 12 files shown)


Trio/Sources/Modules/Settings/View/SettingsRootView.swift
```diff
@@ -289,6 +291,30 @@ extension Settings {
                         }
                     ).listRowBackground(Color.chart)
 
+                    if let notes = releaseNotesService.notes {
+                        Section(
+                            header: Text("Latest Release"),
```
MikePlante1: I suggest retitling this to "Release Notes" since while this refers to the "latest release" installed on the device, it is not necessarily the "latest release" available of Trio.

Trio/Sources/Services/ReleaseNotes/ReleaseNotes.swift
```diff
@@ -0,0 +1,263 @@
...
+    let name: String
+
+    /// Full release body, in GitHub flavoured Markdown.
+    let body: String
+
+    /// Page to open for the complete notes.
+    let htmlURL: String
+
+    /// ISO-8601 publication timestamp. Empty when GitHub did not supply one.
+    let publishedAt: String
```
MikePlante1, line 24: It doesn't look like this is used anywhere. Should it be dropped from getting read, or should it get added to the Release Notes sheet?

Trio/Sources/Services/ReleaseNotes/ReleaseNotes.swift
```diff
@@ -0,0 +1,263 @@
...
+
+        func flush() {
+            defer {
+                currentKind = nil
+                currentLines = []
+            }
+            guard let kind = currentKind else {
+                return
+            }
+            let text = currentLines.joined(separator: " ").trimmingCharacters(in: .whitespaces)
```
MikePlante1: should these be joined with a "\n" instead of a " "? I'm just picturing a future release notes callout line not ending with a punctuation and then it flowing into the next line muddying its clarity.

Like, if v0.8.4 release notes were missing proper punctuation:

```
If you are using Medtrum Nano please update at your earliest convenience
If you are using any Omnipod model, we recommend you update soon
```

would display in the app like

<kbd>
If you are using Medtrum Nano please update at your earliest convenience If you are using any Omnipod model, we recommend you update soon
</kbd>

trioneer-dev:
Moving this back to draft to address @MikePlante1’s comments and also rework the feature a little bit to include old(er) releases as notes, too.

trioneer-dev:
Addressed your review comments @MikePlante1, and updated the PR to include more than just the latest notes.

AndreasStokholm (approved):
Tested and reviewed. Works well 👏

PR #1398, opened 2026-08-09
Show Dev Branch Warning Popup

trioneer-dev:
## Summary

Adds a startup warning, modelled on Loop's dev-branch alert, shown once per launch when the build's branch is anything other than main.

> [!IMPORTANT]
> Blocked by an update to TrioDocs.
> See `trio-docs` PR https://github.com/nightscout/trio-docs/pull/257

The branch is already recorded in `BuildDetails.plist` by `capture-build-details.sh` and read back through `BuildDetails.trioBranch`, so no new build-time plumbing is needed. Wording adapts: the dev branch is named directly, anything else names the branch that was actually built.

Suppressed while Core Data is loading and during onboarding. The scene phase hook alone cannot carry the warning on a cold launch, because the scene is already active by the time Core Data finishes, so it is also triggered as the loading screen gives way to the app.

Compile with `DEV_BRANCH_WARNING_DISABLED` to suppress it. The project-level Debug configuration hardcoded `SWIFT_ACTIVE_COMPILATION_CONDITIONS` without `$(inherited)`, which silently discarded the flag for Debug builds; adding `$(inherited)` makes `ConfigOverride.xcconfig` behave the same in Debug and Release. Browser builds can set the repository variable `DEV_BRANCH_WARNING_DISABLED` instead.

`AppVersionChecker` presents from the same root view controller on the same scene phase change, and UIKit drops a second presentation silently. showAlert now skips when another alert is already showing and reports whether it appeared, so its callers no longer burn their 24 hour and 2 week throttles on an alert that was never displayed.

### Screenshots

| ☀ feature branch | ☀ dev | ☾ feature branch | ☾ dev |
| feature-branch-light | dev-branch-light | feature-branch-dark | dev-branch-dark |

### DIFF
--- a/.github/workflows/build_trio.yml
+++ b/.github/workflows/build_trio.yml
@@ -238,6 +238,20 @@ jobs:
           # Several patches may be added per submodule.
           # Adding comments (#) is strongly recommended to easily tell the individual patches apart.
 
+      # Optional: silence the startup warning shown on builds made from a branch other
+      # than main. Off unless you deliberately set the repository variable
+      # DEV_BRANCH_WARNING_DISABLED to true (Settings > Secrets and variables > Actions
+      # > Variables). Intended for forks and redistributions that always build from
+      # their own branch - if you are testing dev, leave this alone and keep the warning.
+      - name: Configure Dev Branch Warning
+        if: vars.DEV_BRANCH_WARNING_DISABLED == 'true'
+        run: |
+          # ConfigOverride.xcconfig is gitignored and is included by Config.xcconfig,
+          # so writing it here feeds the flag into the Release build without touching
+          # any tracked file.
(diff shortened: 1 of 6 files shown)


MikePlante1:
For fresh builds, the warning doesn't show up at all until the next time their app cold-opens. Something like this would make the warning appear as soon as the onboarding completes:


Screenshot 2026-08-12 at 16 13 38

MikePlante1:
Suggestion: Check if the warning was acknowledged rather than whether it was requested to be shown:

MikePlante1 (approved):
I posed two suggestions for improvements, but neither are blockers so I've approved it as-is. Just right before merging, be sure to merge the [docs PR](https://github.com/nightscout/trio-docs/pull/257)

PR #1406, opened 2026-08-10
Clarify CGM smoothing requirement text

### LINKED ISSUE #1404: Badly worded description for glucose smoothing
The information dialog box for glucose smoothing states that 4 readings within 12 minutes are required. Can the wording be changed to correctly describe that it requires 4 readings with a gap of no longer than 12 minutes? Something like 'It needs 4 readings in a row, each arriving within 12 minutes of the one before it, to work properly.'

Image

MikePlante1:
Closes #1404

### DIFF
--- a/Trio/Sources/Localizations/Main/Localizable.xcstrings
+++ b/Trio/Sources/Localizations/Main/Localizable.xcstrings
@@ -11405,6 +11405,9 @@
         }
       }
     },
+    "%lld m" : {
+
+    },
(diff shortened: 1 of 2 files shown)


trioneer-dev (approved):

trioneer-dev (changes requested):
Please build this change so the String Catalogue gets re-synced and updated.

trioneer-dev (approved):

PR #1380, opened 2026-08-06
Bump OmnipodKit to e31a8d1

trioneer-dev:
Moves the `OmnipodKit` submodule to the current tip of [loopandlearn/OmnipodKit `dev`](https://github.com/loopandlearn/OmnipodKit/tree/dev).

`cc47471` → `e31a8d1`

### What's included

Two commits, [loopandlearn/OmnipodKit#124](https://github.com/loopandlearn/OmnipodKit/pull/124):

- `82ac60b` Fix self-inflicted connect → cancel → reconnect loop in connect-on-demand
- `e31a8d1` Merge pull request #124 from loopandlearn/fix/backport-connectondemand-self-disconnect-loop

PR #1407, opened 2026-08-11
Bump OmnipodKit to 792dc19

trioneer-dev:
Moves the `OmnipodKit` submodule to the current tip of [loopandlearn/OmnipodKit `dev`](https://github.com/loopandlearn/OmnipodKit/tree/dev). Follow-up to #1380.

`e31a8d1` → `792dc19`

### What's included

https://github.com/loopandlearn/OmnipodKit/pull/128:
`4162807` Fix re-entrant `managerQueue` deadlock surfacing a pod alert from an advert

`surfacePodConditionAndQuiet` runs on `managerQueue` (from `centralManager didDiscover`) and called `omnipodDidDetectAlert` inline. The host handles that synchronously via `getPodStatus` → `runSession` → `bleRunSession` → `peripheralManager(forIdentifier:)`, which does `managerQueue.sync` — a sync-to-self deadlock. It fired on a fresh launch when a DASH pod advertised a fault/alert transition before any connection had established `BlePodComms.manager`. The fix dispatches the delegate notification off `managerQueue`; alarm-scan quieting stays on it.

### DIFF
--- a/Config.xcconfig
+++ b/Config.xcconfig
@@ -19,7 +19,7 @@ TRIO_APP_GROUP_ID = group.org.nightscout.$(DEVELOPMENT_TEAM).trio.trio-app-group
 
 // The developers set the version numbers, please leave them alone
 APP_VERSION = 0.8.4
-APP_DEV_VERSION = 0.8.4.56
+APP_DEV_VERSION = 0.8.4.58
 APP_BUILD_NUMBER = 1
 COPYRIGHT_NOTICE =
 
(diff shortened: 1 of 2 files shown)


trioneer-dev:
See this comment for test of the applied fix for Trio. https://github.com/loopandlearn/OmnipodKit/pull/127#issuecomment-5244472666

MikePlante1 (approved):
Everything looks good. Confirmed to match loopandlearn/OmnipodKit:dev. Built it to a simulator for good measure.

PR #1417, opened 2026-08-16
LoopFollow Remote meal edit

aug0211:
- Allow remote editing & deleting of meals in Trio, from LoopFollow
- Paired with a PR for LoopFollow, which will submit the new remote event type with the needed data

### DIFF
--- a/Trio/Sources/APS/Storage/CarbsStorage.swift
+++ b/Trio/Sources/APS/Storage/CarbsStorage.swift
@@ -8,6 +8,135 @@ protocol CarbsObserver {
     func carbsDidUpdate(_ carbs: [CarbsEntry])
 }
 
+let remoteMealMutationMaximumAge: TimeInterval = 12 * 60 * 60
+
+private struct RemoteMealImportSuppression: Codable {
+    let date: Date
+    let expiresAt: Date
+}
+
+private final class RemoteMealImportSuppressionStore: @unchecked Sendable {
+    private let lock = NSLock()
+    private let defaults = UserDefaults.standard
+    private let key = "trio.remoteMealImportSuppressions.v1"
+    private let maximumCount = 500
+
+    func record(dates: [Date], now: Date) {
+        lock.lock()
+        defer { lock.unlock() }
+
+        var suppressions = load().filter { $0.expiresAt >= now }
+        for date in dates {
+            suppressions.removeAll { abs($0.date.timeIntervalSince(date)) <= 1 }
+            suppressions.append(
+                RemoteMealImportSuppression(
+                    date: date,
+                    expiresAt: date.addingTimeInterval(24 * 60 * 60)
+                )
+            )
+        }
+        suppressions.sort { $0.expiresAt < $1.expiresAt }
+        save(Array(suppressions.suffix(maximumCount)))
+    }
+
+    func activeDates(now: Date) -> [Date] {
+        lock.lock()
+        defer { lock.unlock() }
+
+        let stored = load()
+        let active = stored.filter { $0.expiresAt >= now }
+        if active.count != stored.count {
+            save(active)
+        }
+        return active.map(\.date)
+    }
+
+    private func load() -> [RemoteMealImportSuppression] {
+        guard let data = defaults.data(forKey: key),
+              let suppressions = try? JSONDecoder().decode([RemoteMealImportSuppression].self, from: data)
+        else { return [] }
+        return suppressions
+    }
+
+    private func save(_ suppressions: [RemoteMealImportSuppression]) {
+        guard let data = try? JSONEncoder().encode(suppressions) else { return }
+        defaults.set(data, forKey: key)
+    }
+}
+
+struct MealMutationValues: Codable, Equatable, Sendable {
+    let date: Date
+    let carbs: Decimal
+    let fat: Decimal
+    let protein: Decimal
+}
+
+struct MealFPUEntrySnapshot: Codable, Equatable, Sendable {
+    let id: UUID
+    let date: Date
+    let carbs: Decimal
+}
+
+struct MealMutationSnapshot: Codable, Equatable, Sendable {
+    let id: UUID
+    let fpuID: UUID?
+    let values: MealMutationValues
+    let note: String?
+    let fpuEntries: [MealFPUEntrySnapshot]
+}
+
+enum MealMutation: Equatable, Sendable {
+    case edit(MealMutationValues)
+    case delete
+}
+
+enum MealMutationDisposition: String, Codable, Equatable, Sendable {
+    case unchanged
+    case edited
+    case deleted
+}
+
+struct MealMutationResult: Equatable, Sendable {
+    let disposition: MealMutationDisposition
+    let before: MealMutationSnapshot
+    let after: MealMutationSnapshot?
+}
+
+enum MealMutationError: LocalizedError, Equatable {
+    case invalidExpectedValues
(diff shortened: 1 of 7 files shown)


t1dude:
Could you explain the use case for this? I've been on Oref for years and I can't see how editing or deleting a past meal would make any big difference in automatic dosing decisions. It would change COB for sure, but that's already handled by UAM (both ways)?

aug0211:
The use case is the **same as editing or deleting a meal directly in Trio, but from the caregiver’s device.**

For example, a parent enters a meal remotely, then their child texts to say they only ate half of it. Or maybe the pizza serving doubled at the birthday party. The parent should be able to update that entry. The same applies to correcting a typo, removing an accidental duplicate, or correcting fat and protein amounts so that future FPU entries are based on what was actually eaten.

**Would we remove meal editing and deletion from Trio because UAM can compensate? If not, remote caregivers should have access to those same capabilities.** The need to correct an entry does not change depending on which device someone uses to manage it. Caregivers should not have to demonstrate a different or greater need for functionality we already recognize as useful locally (via Trio).

I understand the question about whether this materially changes automatic dosing - rather than debating how FPUs are impacted by doubling a pizza serving or cutting it altogether, I'm sure we can agree that the value of correcting an entry is not limited to its dosing impact. Even where UAM compensates, caregiver teams still have reason to keep the information they entered accurate - for themselves and their partner caregivers making decisions based on information entered into the system.

You mentioned several years of experience with oref. The perspective I’m adding is several years of managing T1D with oref **remotely**, which is the experience this PR addresses. **When Trio runs on a child’s phone and LoopFollow is the parent’s primary interface, remote controls are not just a convenience. They are the tools the parent uses for day-to-day management, often without physical access to the Trio device.** Being able to enter a meal but not correct or delete it leaves that workflow incomplete.

A useful way to evaluate this is to consider life managing through LoopFollow exclusively for an extended period of time, without opening Trio to make changes. Which capabilities would you need available remotely? That is the perspective behind this and other requests for caregiver parity. No one needs to personally adopt that workflow to contribute to the discussion, but the experience of caregivers who rely on it every day should carry meaningful weight in establishing the need.

T1D is already hard. This is an opportunity to make everyday management easier for caregivers by allowing them to correct information through the same interface they used to enter it.

**More broadly, I’d like us to treat remote caregiving as a first-class use case. When functionality exists in Trio, the starting point should be making it available to remote caregivers too**. If there are specific concerns, let’s identify and address them directly. **Parity should be the starting point, with limitations justified by concrete concerns, rather than requiring caregivers to repeatedly establish the value of existing functionality.** To be clear, I'm not saying that you _personally_ are requiring caregivers to repeatedly establish the value - rather, that the remote caregiver experience has often felt like an afterthought and secondary, in my years of experience as a remote caregiver and code contributor. I'd like us to shift that perspective to consider remote use cases as first class citizens in the DIY T1D community ❤️ After all, many people in the T1D community are not managing controls themselves alone - many are managed remotely.

This PR addresses one of the remote capability gaps, and the Facebook request below is another example of caregivers asking for that capability.

https://www.facebook.com/share/p/19XqZCXT46/

image

dnzxy:
Thanks for your elaborate comment. I think the collective work by @bjorkert, @codebymini, other caregivers on the Trio team and myself prove that we see remote control as a first class citizen.
The remote commands as they exist today exist in the feature set of LoopFollow <> Trio first for a reason.

I disagree with large parts of this code as it introduces very vibe-code-esque concepts and duplicate state and stateful behavior, but conceptually we can look into it. The basic need of deleting a remotely entered meal is a valid one; but to also be very open here, we also need to step in and disallow certain caregiver types from obsessively monitoring and editing the tiniest carb amounts all the time, for their mental burden’s sake and their caregivee’s. Because at the end of the day that *can* influence dosing outcomes and system behavior.

Edit to add:
Since your work touches both the apps, we’ll need to work with the LF team on this, too. Thus far the PR there looks stale. Is the functionality one that is something the LF team wants to add? If no, we are somewhat at a standstill here. If yes, let’s get details sorted.

It did not make the cut for v1.0, I’m sorry. To some degree because you did not reply to the question asked here so it was labeled as on hold.

We can look into this, but that will need to happen *collaboratively*, and that may mean change requests or not accepting certain parts of the proposal.

Edit again to fix typos…

aug0211:
Not sure I followed all that. PR was submitted about a month back. Got a response 6 days ago, about 3w after PR going up, and this was "put on hold" after that 3 week comment. I replied within the week of getting that comment. Trio PR still merges without conflicts (I'm using it with 1.0.4 currently, built hours ago), and also with current LF (also built latest a few hours ago with the accompanying PR). Happy to collaborate, that's why I shared the code out via PR.

Ultimately I'm building stuff I use as a caregiver. When someone publicly requested this, it reminded me the PR was even up. I felt bad not sharing to them that the functionality indeed exists and is in use by people today. Didn't want to be selfish and have wanted to share code to help others.

As for people "doing too much" - sorry, this falls flat. There's literally a user who this week was hoping for a way to delete an accidental duplicate entry for her son's _safety_. And if this functionality exists on local Trio (it does of course), it certainly makes sense for remote capability as well. I built this for a reason.

This merges without conflicts, works consistently, and meets a real safety need that real remote caregivers have use for. Your call here. I'm hoping someone else can benefit from this functionality 🤞

aug0211:
Sorry can't hold back on this one:

> Thanks for your elaborate comment. I think the collective work by @bjorkert, @codebymini, other caregivers on the Trio team and myself prove that we see remote control as a first class citizen.
The remote commands as they exist today exist in the feature set of LoopFollow <> Trio first for a reason.

I'm sure you'll recall that modern day remote commands including a LF interface exist in LF & Trio today because of Daniel's and my work. We built the first remote functionality from LoopFollow and Trio 🙂 That reason you're citing is me and Daniel.

dnzxy:
This likely won’t merge as-is for what I tried to point at in my initial comment, re: it introducing a weird state machine for pseudo / temp / intermittent carb handling. Thats an implementation detail. Conceptually, i.e. deleting a remotely entered carb entry, it makes sense, I agreed in my very comment :) 

Just because a proposal merges without conflicts does not mean it’s mergeable, and I‘m very sure you know this from own experience. 

As for your other comment: are you not "other caregivers on the Trio team"? My bad. I didn’t mean to not give credit to anyone :) 
You guys built an initial proof of concept using various concepts that were not scalable. So while
you guys absolutely pushed the idea, others made it work in a way that works for non-techy users, is robust, doesn’t rely on third party services (other than Apple), and is scalable in a way that lets it be used today by thousands of caregivers. Collaborative effort. I don’t think we need to call
dips here; I certainly didn’t mean to. I simply wanted to emphasize on the fact that you keep repeating this old tale of caregivers not being seen as first class citizens and that’s a) not true, and b) not the point when we look at proposals like this. 

Please let me know how to best reach you, so we can get detail sorted between this proposal and the LoopFollow proposal.

aug0211:
I understand you have a different conceptual implementation you'd like to pursue.


Since there are no functional or UX issues using this version daily, I will continue to use mine as it is, and will consider updating once your refactor ships 👌 Excited to see you're interested in the functionality as well ❤️


My dev cycles for now will go toward implementing new functionality that's helpful for our situation. Just a matter of time available, and it's tough to prioritize a refactor when the functionality and UX nail the use case, and there are other valuable features to spend time on.

Looking forward to seeing a future conceptual implementation of this land one day for broader availability and glad to have provided a starting ground to accelerate that path.

PR #1420, opened 2026-08-17
Report Dexcom sensor states to Nightscout as notes

bjorkert:
When a G6 or G7 reading carries no reliable glucose, Trio uploads a Note treatment with the kit's own name for the sensor state, for example "CGM: sensorFailed", "CGM: questionMarks" or "CGM: .unknown(23)". Trio holds no state table, titles or localizations for this: the kit's hasReliableGlucose decides whether a reading is noted, and the kit's description names it.

A state is noted once and stays quiet while it persists. A reliable reading or a new sensor session clears the log, so a state that returns after recovery is noted again. A failed upload is retried on the next reading. The notes are controlled by a new Upload CGM Sensor States switch under Services, Nightscout, Upload, and in the onboarding upload step; it defaults to off in settings and requires Allow Uploading to Nightscout. Lifecycle states without glucose (warmup, stopped, session ended, G6 initial calibration) are noted too, roughly two to four notes per sensor session.

Scope is Dexcom G6 and G7, since the state lives on the kit-specific latestReading. Libre 3, AccuChek and Eversense issue LoopKit alerts and can follow by mirroring those to notes in PluginSource.issueAlert.

Tests cover the reliability and naming rule, the dedupe log, and G7 transmitter messages fed through the real G7CGMManager. The upload path was exercised end to end against a Nightscout instance with a scripted state sequence.


trioneer-dev:
Overall this feature is a great addition.

I wonder if the reworked `DeviceCatalog`, which carries all available CGM managers, or our existing `AlertCatalogRegistry`, which currently serves as a shim for the LoopKit Alert protocol as of LoopKit `dev` vs. `next-dev`, can be utilized for this feature, instead of manually implementing `CGMSensorStatus` which essentially carries over error messages / strings. 

Also noteworthy that this currently only covers Dexcom CGMs (G6/One, G7/One+) and misses to cover the other available CGMs as of Trio `dev`, i.e. FreeStyle Libre 3/3+, AccuChek SmartGuide, Eversense E3/E365.

t1dude:
This is a great feature @bjorkert :) Could you have a look at the comment from @trioneer-dev and let us know how you suggest to move this forward? I guess it's okay to launch this with support for only some CGM types, but it would definitely be better if they're all supported. Let us know what you think and I'll try to get this through.

bjorkert:
Thanks both. Personally, I'd prefer to merge this with Dexcom support as it is now and treat the other CGMs as follow-ups. I don't have hardware for other sensors, so I can't test them and I wouldn't be confident in code I write for them blind. The episode handling should be generic, so adding a sensor mostly means mapping its states in CGMSensorStatus, so someone who runs one of those sensors would be a much better fit to add it.

t1dude (approved):
LGTM

dnzxy:
I‘m okay with only having Dexcom support present for now.

However, I‘d like for us not to duplicate error strings and adding an additional mapping layer when we already have an existing mapping + shim layer using the device catalog and the alert catalog. 

Can we streamline those?

bjorkert:
I'll have a look!

bjorkert:
Currently working on a rewrite/simplification of this PR. Changing to "Draft".

LiroyvH:
Q: this looks rather nice, but will this show up as notes on the graph of every Nightscout Report and in apps like xDrip4iOS as well? Because in a period with some heavy sensor noise on and off (happens more often on Anubis I suppose), that might clutter the report/screen and be perceived as annoying if it does. (In which case being able to toggle it off would be nice, but I don't think it offers that now unless I overlooked a change.) (When retained indefinitely that may also use more storage, but that's probably rather negligible even with multiple events per sensor.)

MikePlante1:
> Q: this looks rather nice, but will this show up as notes on the graph of every Nightscout Report and in apps like xDrip4iOS as well? 

I didn't test this PR, but it says it's added as a "note" and notes shows up like this in Nightscout's main chart, where the text of the note is only shown if you put your cursor on it:

Screenshot 2026-09-14 at 20 59 35

And notes only show up in Day-to-Day reports if you check the box to display notes:

Screenshot 2026-09-14 at 21 00 20

It shows up in the Treatments report, but that's how it should be

LiroyvH:
@MikePlante1 Yeah I think that's why it might be nice if you can disable it. Notes I make to, for example, write down the patch has failed or fell off or something else unexpected happened is nice to have included, but an automated note for every time my sensor does something weird: not so much and can pollute the chart with a (temporarily) very noisy sensor. (plus it'll be obvious from no readings in the chart that something ain't right with the sensor). But you unfortunately can't selectively choose what notes to include so it dumps it all in the report, all or nothing. As such it'd be nice if it could be switched off as to not getting noise whenever there's noise. ;)

But that's just a personal preference that I'd figure to share for consideration :)

bjorkert:
Just to be clear, there is one note per error episode, not one per 5 minutes. The state is noted once when it first appears, and stays silent for as long as it persists.

The main driver here is being able to see if the stopped readins is because of a sensor error. Dexcom Follow doesn't report this, so caregivers and remote monitors currently have no way to know. Transmitter errors won't produce notes.

On a toggle: I'm happy to add one if that's the consensus.

LiroyvH:
@bjorkert Yup, for that it's a really nice addition. To illustrate why I was looking for a toggle: past 2 days I was on a lengthy trip (1000KM+) when the sensor was starting to give out; but had no good opportunity to change it out. Today only, looking at the amount of times it went back and forth between "OK" and "Temporary Sensor Issue" in the car: I probably would've had approximately 9 to 10 notes (18 to 20 if it also puts a note when it recovers) on my report chart.

Don't get me wrong though, its not exactly a terrible issue and heavy sensor issues like that are relatively rare. Its just if you don't really have a usecase for the notes then its an inconvenience it gets printed on the reports like that; and the amount of notes quickly increases with such multiple events of the sensor bravely fighting its inevitable demise. :)

bjorkert:
Toggle added: Services > Nightscout > Upload > "Upload CGM Sensor States", and also on the onboarding upload step. Notes are only uploaded when both it and Allow Uploading are on. No note is written on recovery (a successful bg value is confirmation enough), only when a state first appears, so the car trip would have been the 9 to 10, not 18 to 20.

dnzxy:
Good change. Let me know when you are happy with it for another round if review :)

bjorkert:
Example: 
CleanShot 2026-09-18 at 15 23 09@2x

bjorkert:
It has been running now on my test subject and it works just as described. Ready for review.

PR #1427, opened 2026-08-19
fix for contact backgrounds having unstable background colours

AndreasStokholm:
Over the last while I've noticed that the contacts on my Apple Watch would randomly switch between white and black backgrounds. White backgrounds stick out a lot and make it hard to read the values too. This PR fixes it by simply setting the background to black.

I've run this with my setup for 2-ish days now and have not seen any ill effects.

The fix applies to the top and left most circular complications on this screenshot:

incoming-B264DCAC-0C4D-403F-8CE2-3D88172D0953


trioneer-dev (approved):
LGTM from code view.

AndreasStokholm:
I've been running this newer version for 3 days, and it works well.

PR #1429, opened 2026-08-19
Fix Pump Basal Rate Rounding

### LINKED ISSUE #1424: `roundBasal` uses hardcoded rate bands that only match one pump
[`TempBasalFunctions.roundBasal`](https://github.com/nightscout/Trio/blob/38f3b369b717273e1d0e7e0f9b10e710aeac3597/Trio/Sources/APS/OpenAPSSwift/DetermineBasal/TempBasalFunctions.swift#L16-L36)

### Issue description

The three bands — 0.025 below 1 U/h, 0.05 to 10, 0.1 above — are the basal rate table of
a Medtronic x23/x54 and nothing else. They are applied to every pump, and the function
takes no pump input beyond `profile.model`. The `// FIXME: Should we just call the
pumpManager here?` on the first line of the body is the same observation.

The pump's real table is already available and already used elsewhere in the app:

- [`BasalProfileEditorStateModel`](https://github.com/nightscout/Trio/blob/38f3b369b717273e1d0e7e0f9b10e710aeac3597/Trio/Sources/Modules/BasalProfileEditor/BasalProfileEditorStateModel.swift#L51) offers the user exactly `supportedBasalRates`
- [`APSManager.enactTempBasal`](https://github.com/nightscout/Trio/blob/38f3b369b717273e1d0e7e0f9b10e710aeac3597/Trio/Sources/APS/APSManager.swift#L700) applies it at delivery via `roundToSupportedBasalRate`
- [`TDDStorage`](https://github.com/nightscout/Trio/blob/38f3b369b717273e1d0e7e0f9b10e710aeac3597/Trio/Sources/APS/Storage/TDDStorage.swift#L216) already takes that function as an injected closure

So the profile editor offers rates from the pump, and `determineBasal` cannot reproduce
them. Rates each pump supports that `roundBasal` alters, checked against every entry of
the kit's own `supportedBasalRates`:

| pump | unreachable | example |
|---|---|---|
| Dana (0.01 flat, to 3 U/h) | **240 / 300** | 0.01 → 0, 0.03 → 0.05 |
| Minimed pre-x23 (0.05 flat) | 250 / 700 | 10.05 → 10.1 |
| Omnipod / Medtrum (0.05 flat) | 200 / 600 | 10.05 → 10.1 |
| Minimed gen ≥23 (0.025/0.05/0.1) | 20 / 470 | 0.025 → 0.05 |

A Dana user can set 1.23 U/h in the basal editor;
[`computeAdjustedBasal`](https://github.com/nightscout/Trio/blob/38f3b369b717273e1d0e7e0f9b10e710aeac3597/Trio/Sources/APS/OpenAPSSwift/DetermineBasal/DetermineBasal%2BHelpers.swift#L174)
turns it into 1.25 whenever the sensitivity ratio applies. Resolution the pump supports is
discarded inside the algorithm, so the delivery-side lookup cannot recover it.

### Secondary: the model branch is dead

`profile.model` comes from `settings/model.json`, read at
[`OpenAPS.swift#L500`](https://github.com/nightscout/Trio/blob/38f3b369b717273e1d0e7e0f9b10e710aeac3597/Trio/Sources/APS/OpenAPS/OpenAPS.swift#L500)
and written nowhere in the codebase. Every install uses the
[bundled default `"722"`](https://github.com/nightscout/Trio/blob/38f3b369b717273e1d0e7e0f9b10e710aeac3597/Trio/Resources/json/defaults/settings/model.json),
so `lowestRateScale` is always 20 and the x23/x54 branch never executes — a Medtronic
gen ≥23 is rounded to 0.05 below 1 U/h despite delivering 0.025.

### Suggested fix

Either we answer the FIXME: pass `pumpManager.roundToSupportedBasalRate` into the algorithm as a
closure, the pattern `TDDStorage` already uses, and drop the bands. Exact for every pump.
 OR
Easy fix is to just drop the bands, the algo might use increments too fine, but they will be floored in pumpManager.

trioneer-dev:
## Summary

As reported in #1424, `TempBasalFunctions.roundBasal` erroneously considered every pump to one hardcoded set of supported basal rates, 0.025 below 1 U/hr, 0.05 to 10, 0.1 above, which is actually a Medtronic x23/x54’s rate... Further, `profile.model` was read from `settings/model.json`, a file that was never ever written to disk. As a consequence, every algorithm run fell back to the bundled `"722"`

  ### Changes

  1. **Drop the unused pump model setting.** `Profile.model`, both `ProfileGenerator parameters, the `createProfiles` load, `Settings.model`, and the bundled `settings/model.json`.
  2. **Round with the actual pump's own supported rates.** `roundBasal` floors against a new non-persisted
     `Profile.supportedBasalRates`, injected per algorithm run from the paired driver in `OpenAPS.determineBasal`. This mirrors LoopKit's `PumpManager.roundToSupportedBasalRate`, so the rate the algorithm commits to is the rate the pump delivers / can actually deliver.
  3. **Round the closed-loop dose before delivery as an additional safety net.** `performBasal` and `performBolus` handed the determination to the driver unrounded, unlike their `enactTempBasal` / `enactBolus` twins in APSManager.

Adjusted unit tests accordingly.

Fixes #1424.

### Update

@mountrcg found that point 3 was wrong: `Foundation`'s `Decimal` → `Double` conversion can land just below the value (0.07 becomes 0.06999999999999999), and since every pump table floors, re-rounding at delivery dropped a full increment on 41 of Dana's 301 rates. Two follow-up commits:

- **Hand the pump a bit-exact rate.** `performBasal` / `performBolus` now convert via the decimal string, which is bit-identical to the driver's own table entry, and no longer re-floor — the algorithm already floored against the same table. The `roundToSupportedBasalRate` comparison stays as a log-only canary. The algorithm side was never affected: it stays in `Decimal` end to end.
- **Fix a test type-check timeout on Xcode 26.2**, which CI pins. Both jobs are green.

`BasalRoundingDriverParityTests` now walks every entry of all five kit tables through the real delivery path, with a regression pin on 0.07 / 0.14 / 1.99 against `Double(truncating:)`.


mountrcg:
Before the algo rounding was done to nearest, I think you changed it to flooring now.

mountrcg:
I have increased testing to all pumps and bands and compared the algorithm against what APSManager actually hands over. You compared to driver fed a raw Double(truncating:) . For Dana with the very small increment <code>Foundation</code>'s <code>Decimal</code> → <code>Double</code> conversion rounds low. Measured my tests

| value | pump's table holds | what a Decimal becomes
| -- | -- | --
| 0.07 | 0.07 | 0.06999999999999999
| 0.14 | 0.14 | 0.13999999999999999
| 1.99 | 1.99 | 1.9899999999999998

Every pump table floors (<code>supportedBasalRates.last { $0 &lt;= requested }</code>), and a floor is unforgiving in this direction

trioneer-dev:
> Before the algo rounding was done to nearest, I think you changed it to flooring now.

@mountrcg , yes like the PR summary describes:
> `roundBasal` floors against a new non-persisted Profile.supportedBasalRates, injected per algorithm run from the paired driver in OpenAPS.determineBasal. This mirrors LoopKit's PumpManager.roundToSupportedBasalRate […]

The other comment I cannot understand among all the LLM word barf.  Could you clarify, please :)

trioneer-dev:
Updated PR summary, pushed 2 fixes.

mountrcg:
Tests fine with me now.

kingst (approved):
LGTM

PR #1430, opened 2026-08-19
Fix pump battery not updating after pumpmanager swap

jwoglom:
BaseAPSManager registers itself as a PumpManagerStatusObserver at init. This means if the pump manager is replaced without an app relaunch (i.e. switching pumps) the observer stays bound to the previous instance.

The delegate is re-attached in DeviceDataManager's didSet, so reservoir/dose/status data kept flowing, but the observer-driven storeBatteryStatus never ran for the new pump -- no OpenAPS_Battery rows were written and the home battery icon stayed hidden until relaunch.

Fix by having DeviceDataManager emitspumpManagerDidChange from the pumpManager didSet (where the delegate is already re-attached), covering every assignment path. APSManager re-registers its status observer on that signal.


jwoglom:
I've cherry picked this onto my active trio build and can confirm it doesn't horrifically break any behavior. But I haven't specifically replicated the bug (want to avoid re pairing mobi)

dnzxy:
@jwoglom this will work, but it adds a 3rd way of getting this info. 

`PumpManagerDelegate` already conforms to `PumpManagerStatusObserver` and `BaseDeviceDataManager` is the delegate that already implements `pumpManager(_:didUpdate:oldStatus:)` over in DeviceDataManager.swift:492-580, where it persists the reservoir and expiry. The delegate gets re-attached in `didSet` on the very same line the new subject would fire from, so there's no way for it to go stale. All five PMs notify the delegate and observers from a single call site (OmniPumpManager.swift:287, MedtrumPumpManager.swift:885, DanaKitPumpManager.swift:1424, MinimedPumpManager.swift:172, MockPumpManager.swift:253) already.

My suggestion: if we move the battery and `Monitor.status` write into that callback, you close both issues at once. You also get to delete APSManager.swift:203 along with the entire `PumpManagerStatusObserver` conformance, and you skip the protocol change entirely. As a bonus, it lets us fold in the duplicate MockPump write at DeviceDataManager.swift:160-192.

If that feels like too much change, we culd also subscribe to the existing `pumpDisplayState` instead of adding a new PassThrough subject. It already emits on every assignment, it's already on the APSManager protocol, and Trio's HomeState is already using it as this kind of handler. We also have an odd comment saying as much at BatterySetup.swift:25.

trioneer-dev (approved):
Code-wise LGTM.

straykatz:
Battery is now showing and updating. 
(I hate to say, I don't know whether it is due to this change, though, because it was showing before I updated to this version. Initially - when I first installed feat-tandem, the battery was not showing.) sorry, not a very precise comment, but Dan asked :-)

PR #1431, opened 2026-08-20
Floor quick-pick bolus suggestions to one decimal place

t1dude:
Changing rounding to simplify bolus suggestions and at the same time avoid a UI issue with two decimals in the suggested amount.


trioneer-dev:
Tested on device.

| Before | After |
|--------|--------|
| image | image | 

Code LGTM too.

trioneer-dev (approved):

PR #1436, opened 2026-08-21
Revert "upgrade looping animation and remove looping text"

trioneer-dev:
Reverts nightscout/Trio#1251


LiroyvH:
The new looping animation is fun to see, but it currently causes some issues - see attached:

https://github.com/user-attachments/assets/58bca15a-3ecf-478c-917a-fd1684127f4a

https://github.com/user-attachments/assets/94cf7a02-5e19-450f-a963-fd8314d4b66c

https://github.com/user-attachments/assets/1625a0a1-f2bd-4841-85bf-f76f8bf60bef

If the number that its stuck on is identical to the current time, you only see it marquee from left to right - which is faintly annoying); but when it differs it starts looking rather bad. 

Note: none of these videos are looped.

EJM0:
Pushed a Fix in #1437

AndreasStokholm (approved):

PR #1439, opened 2026-08-21
Introduce Device Catalog, Refactor Device Pickers and Use Pump-Derived Basal Increments

### LINKED ISSUE #1413: Pump Increments Are Not Configurable During Onboarding
## Describe the bug
In the onboarding flow, the Pump Model selector for insulin delivery increments doesn't actually change the available basal rate increments.

### Trio Version:
* Version Number: *8.4.60*

trioneer-dev:
## Summary 
Consolidates CGM and pump metadata into one `DeviceCatalog`, and replaces the four hardcoded `confirmationDialog`s with a single grouped picker sheet.

**Why:** adding a CGM meant editing four hand-synced lists; a pump, seven. They had already drifted — `cgmOptions` advertised "Dexcom Share" and "FreeStyle Libre Demo", which matched nothing and could never be selected.

**No submodule changes.** Trio links the kits statically, so the picker reads each kit's existing artwork via `Bundle(identifier:)` rather than needing `pickerImage` added upstream (LoopKit#604 et al. — 17 PRs across 15 repos, all still open).

Also **fixes #1413**: the onboarding Pump Model selector never changed the basal increments — every option used a hardcoded 0.05 step. Bounds now come from each kit's `onboardingSupportedBasalRates`, so Dana gets its real 0.01. Zero is excluded there and in therapy settings, because oref rejects a schedule containing a 0 rate.

## Screenshots
| Pump Options | CGM Options |
|--------|-------|
| image | image |
| image | image |

## User-visible
- Dexcom Share and FreeStyle Libre Demo no longer appear in the CGM hint list (they were never selectable).
- Basal rate minimum is no longer 0, in onboarding and therapy settings.

27 unit tests; four artwork gaps fall back to a category glyph and are pinned by test.


munchyk:
Nice improvements.

2 questions: 

Would it be better to add support for a 0 U/hr basal rate schedule rather than removing it as an option? I don't really know how important or prevalent this is so it may be a non-issue. (im assuming this would prob require separating out `current_basal` into something like `scheduled_basal` and `dosing_basal` to account for the 0 multiplier problem, which may add unnecessary complexity)

Also, would it make more sense to limit the onboarding pump selection to 2 options - "Dana" & "All Other Pumps"? I can see people worrying about the implications of selecting a specific pump and the difficulty of adjusting settings if switching to a different pump later, when in reality all the other pumps are limited to the same basal rate ranges/increments (unless keeping support for a 35 U/hr basal rate on minimed is more important than a consistent 30 U/hr across all pumps) and Dana is the only one with a meaningfully different basal rate range/increments. Not that important but figured I’d mention

trioneer-dev:
Re Q1: Algorithmically, we do not allow 0 rates, so we are also not going to allow them to be selected. 

Re Q2: There are more pumps going to be added to OS-AID in the future, one currently being in closed testing (Tandem Mobi), and not all of them will use “All other pumps" rates. So, thinking ahead here, we’re grabbing rates from the Pump Manager Kits and are surfacing them appropriately.

Trio/Sources/APS/Devices/DeviceCatalog.swift
```diff
@@ -0,0 +1,530 @@
...
+import MinimedKitUI
+import MockKit
+import MockKitUI
+import ObjectiveC
+import OmnipodKit
+import UIKit
+
+/// Single source of truth for every CGM and pump Trio can pair with.
+///
+/// Loop discovers this metadata at runtime from plugin bundle `Info.plist` keys. Trio statically links every
```
jwoglom, line 29: Is there a con to doing this instead?

jwoglom (commented):
Overall just checking the code this seems good, I'm curious as to why we differ from loop in having a static manager list vs. runtime discovery, though

g5WS:
Perhaps consider explicitly listing the MiniMed models (ie, 515, 715) instead of x15. Not all users would feel confident with this abstraction, and it may prevent some users from completing onboarding.

Trio/Sources/APS/Devices/DeviceCatalog.swift
trioneer-dev: Yes, having to commit to using Loop's plugin approach, which we intentionally decided not to, as it does not conform to Swift 6, modern parallelized, dynamic Xcode builds and over complicates things that we simply don’t need. So we embed the PM frameworks vs. the plugins.

trioneer-dev:
> Perhaps consider explicitly listing the MiniMed models (ie, 515, 715) instead of x15. Not all users would feel confident with this abstraction, and it may prevent some users from completing onboarding.

@g5WS thanks for the feedback. If we were to go that route, we’d have to follow this detailed list, including listing firmware and I personally feel that’s going down a lane of overcomplicating things for a pump used by a shrinking minority of the iOS OS-AID user base. 
image

What do you think? :)

Trio/Sources/APS/Devices/DeviceCatalog.swift
trioneer-dev: Plus, if we did this, then we would have to make changes in _every single_ pump and CGM manager submodule, if you take a peek at the Loop PR that introduces a similar device picker rework (single PR changing things, plus 17 dependent sub-PRs).

g5WS:
> What do you think? :)

Listing firmware specs seems to have an educational purpose - that responsibility lies with the user.  If they build Trio and it doesn’t work with their pump/firmware, that’s on them. 

Not being able to identify your pump in an abstracted list is a UX issue that may prevent users from completing onboarding. I don’t see the user being responsible for this. Not sure if all countries, languages, and cultures use x to indicate variable place value. I think it’s likely to cause issues for some users, especially in the future where  the X in Tandem’s X2 would have a literal meaning.

g5WS:
What about:

“Minimed
515, 522, 523, 554, 715, 722, 723, 754”

or 

“Minimed (with suitable firmware)
515, 522, 523, 554, 715, 722, 723, 754”

trioneer-dev:
Thanks for the input @g5WS . Let me try it out and see how it looks with text zoom (a11y) and normal when we extend the subline by that much. 

Side note: the tslim x2 will likely never become loopable with OS-AID as you cannot remotely issue boluses, only TBRs. For the foreseeable future we‘ll only have the tubed Mobi and (likely) the patch version of Mobi (tbd).

jwoglom (approved):
built on my test phone and looks good

jwoglom:
> > Perhaps consider explicitly listing the MiniMed models (ie, 515, 715) instead of x15. Not all users would feel confident with this abstraction, and it may prevent some users from completing onboarding.
> 
> @g5WS thanks for the feedback. If we were to go that route, we’d have to follow this detailed list, including listing firmware and I personally feel that’s going down a lane of overcomplicating things for a pump used by a shrinking minority of the iOS OS-AID user base. image
> 
> What do you think? :)

@trioneer-dev why not have the first intro page of the Medtronic setup be a list of all the supported models, before asking the insulin type

trioneer-dev:
> @trioneer-dev why not have the first intro page of the Medtronic setup be a list of all the supported models, before asking the insulin type

@jwoglom that’d be a change to `MinimedKit` and out of scope for this PR, plus not up to the Trio team as we do not maintain that pump manager. We can potentially record an issue over at https://github.com/LoopKit/MinimedKit and make a feature / change request.

trioneer-dev:
> What about:
> 
> “Minimed 515, 522, 523, 554, 715, 722, 723, 754”
> 
> or
> 
> “Minimed (with suitable firmware) 515, 522, 523, 554, 715, 722, 723, 754”

Got to admit, not a fan. 
image

I think we’ll stick with the `x15, x22, x23, x54` subline. 

Been tested and approved by the team.

AndreasStokholm (approved):

Issue #1438, opened 2026-08-21
Loop-time-ago shows two numbers


hummelstrand:
## Describe the bug
*The "time ago" pill fades between to different values. One of them seems to always be "4 m"*

BTW, should probably be "min" and not "m" as in meters. :)

## Attach a Log
[log.24.txt](https://github.com/user-attachments/files/31323226/log.24.txt)
[log_prev.17.txt](https://github.com/user-attachments/files/31323229/log_prev.17.txt)

## To Reproduce
Not sure what I did to trigger it. 

## Expected behavior
A single value.

## Screenshots
https://github.com/user-attachments/assets/7d88f28c-5883-425d-ad67-973ae70f6dbc

### Smartphone:
* Hardware: *iPhone 15 Pro Max*
* OS Version: *iOS 26.6*

### Pump:
* Manufacturer: *Medtrum*
* Model: *Nano 300*

### CGM:
* Device: Dexcom G7
* Manager app: Dexcom App

### Trio Version:
* Version Number: 0.8.4.68 (72)
* Repo: *nightscout/trio*
* Git Reference: Trio v0.8.4.68 (72) dev a93d195

## Additional context
*Add any other context about the problem here.*

dnzxy:
`m` is used everywhere if you take a look at the glucose bobble, the running override, etc. 

There's 2 PRs open tackling this (one just a reversal (#1436), the other a potential fix that has to be verified (#1437).

Issue #1435, opened 2026-08-21
Trio man stat history keeps endlessly scrolling left and right until Trio restarted


kfahling:
## Describe the bug
I have recorded an instance where the main Trio stats continues to scroll left and right. I clicked on the screen a couple of times but the scrolling continued. I restarted Trio which corrected the situation. I probably should have tried changing to a different screen but didn't think about that until I restarted Trio.

## Attach a Log
Will attach from my laptop.

## To Reproduce
*Steps to reproduce the behavior:*
1. I don't remember doing anything special.

## Expected behavior
Only scroll when I request it.

## Screenshots
Will attach from laptop.

### Smartphone:
* Hardware: iPhone 14 Pro
* OS Version: iOS 26.6.1

### Pump:
* Manufacturer: Insulet
* Model: Omnipod 5

### CGM:
* Device: Dexcom G7
* Manager app: Dexcom App

### Trio Version:
* Version Number: v0.8.4.68
* Repo: *nightscout/trio*
* Git Reference: Trio v0.8.4.68 (99) dev a93d195

## Technical Details
Nothing I can think of.

## Additional context

Nothing I can think of

kfahling:
[log 39.zip](https://github.com/user-attachments/files/31319752/log.39.zip)

[ScreenRecording_08-21-2026 13-30-29_1.zip](https://github.com/user-attachments/files/31319765/ScreenRecording_08-21-2026.13-30-29_1.zip)

dnzxy:
Related to #1438 , same cause of issue.

kfahling:
Ok, I saw that but didn't make the connection.  Would that also cause the CGM arrow to flip right and down continuously?  I just saw that happen a few minutes ago.  If that's the case I'll close this one out and track https://github.com/nightscout/Trio/issues/1438.

PR #1440, opened 2026-08-22
revision 2 of Feat/update loop indicator 

EJM0:
this revision tries to fix the previously unnoticed bug reverted in #1436 introduced in #1251 and discribed in:
#1438, #1435
it tries to capture the repeatforever modifier from applying onto other changing elements in on transaction
 
Update:
The logic that produced the bug is still removed. Other potential animation faliures have been hardened and let component rendering decide when to start animations, to prevent dropping them.
Reworked CapsuleSpinner to be a view modifier for easier handling.

https://github.com/user-attachments/assets/93e6d162-128a-491d-b37a-0c49e78e2233


EJM0:
testers would be appreciated since the bug occurs rather spontaneously 🙏

Trio/Sources/Views/CapsuleSpinnerView.swift
```diff
@@ -0,0 +1,162 @@
...
+                    isAnimating = true
+                }
+
+                // 3. Wait for transition. Also keeps the reset and the spin target in
+                //    separate updates, so SwiftUI can't collapse 1.0 -> 0.0 -> 1.0 into
+                //    no change at all
+                try? await Task.sleep(for: .seconds(0.3))
+
+                // 4. Drive the normalized 0.0 -> 1.0 spin loop
+                self.spinAnimation = .linear(duration: 1.333).repeatForever(autoreverses: false)
```
MikePlante1: It would be nice to have this honor Reduce Motion. This should make it just look like this if the user has Reduce Motion on in settings:



    @Environment(\.accessibilityReduceMotion) private var reduceMotion
    ...
    self.spinAnimation = reduceMotion ? nil : .linear(duration: 1.333).repeatForever(autoreverses: false)

EJM0: I implemented a solution where the displayed content handles alternative active-processing displaying.(reverting back to just showing looping text)
What do you think?

Trio/Sources/Modules/Home/View/Header/LoopView.swift
trioneer-dev: dead code.

Trio/Sources/Modules/Home/View/Header/LoopView.swift
```diff
@@ -20,26 +21,49 @@ struct LoopView: View {
...
-                if isLooping {
-                    ProgressView()
-                }
+                    .symbolEffect(.pulse, options: .repeating, isActive: showLooping && !reduceMotion)
             }
-            if isLooping {
-                Text("looping")
+            if showLooping, reduceMotion {
+                // neither the spinning border nor the pulse runs here, so say it in words
+                Text("Looping", comment: "Loop status label while a loop cycle is running")
```
trioneer-dev: By adding the comment here, you are orphaning the existing 20 localizations for this label

trioneer-dev (changes requested):
Dead code, and dropped localizations in this PR.

This PR seems like adding a lot of code and conditional polling logic for 0 gain in the overall app UI/UX. Leaning toward not accepting this feature.

EJM0:
Removed dead code and corrected the text.
At the end, I can only say that the code is "only" a visual improvement.
Condenced to the most visual pleasing and code-efficient version I could come up with.
But I think it improves the UI visually and makes it more pleasing to look at. Also, maintainability, reliability are given and its more space efficient through the missing native iOS progressview.
love to see this in the project.

Ps. Ui/Ux gain:
- space efficiency for smaller phones with old display form factor (less overlapping of the bobble) 
- coherent Ui/ux desing language

EJM0:
No animation spillage trought testing over close to a week. Confirmed by [other testers](https://discord.com/channels/1020905149037813862/1239982915199434833/1542881339840139434)

t1dude:
This does not add any substantial improvement to the app. I will close it for now as we are approaching a Main release.

Issue #1441, opened 2026-08-23
Trio dev crashing soon after opening


pauloalexander:
After switching to latest Trio dev build (from latest main) app crashes soon (but not immediately) after opening. My impression is the crash is triggered as soon as a connection to the dash pod is made. 

Tried with two dev builds with latest code, one on 19th Aug (62) and other 13th(59). Tried to switch multiple times over a week, but had to revert back to the main build each time. Tries once during a pod change, app seemed stable until priming was complete, then crashed. Also tried re-starting the phone. Dev app is stable on my phone with a simulator pump, crashes are on my son’s phone.

[log.txt](https://github.com/user-attachments/files/31345372/log.txt)
[log_prev.txt](https://github.com/user-attachments/files/31345373/log_prev.txt)

ips file wasn’t showing up to attach to this ticket but can be found on Trio Discord, Omnipod Dash channel 22 Aug: https://discord.com/channels/1020905149037813862/1239983179931324507/1540724287449858168

iPhone 17e
iOS 26.6.1
Omnipod dash (atlas) - consistently screaming during day 3 hence the desire to switch to dev
Dexcom G7

Browser builds

trioneer-dev:
Fixed via #1443.

PR #1443, opened 2026-08-24
Fixes numerical issue with DynamicISF

kingst:
If DynamicISF is using sigmoid and has a autosensMin value == 1 then
there are numerical issues with log10 returning -Inf, causing a crash
loop (https://github.com/nightscout/Trio/issues/1441).

The fix is to add a fudge factor to avoid this case, similar to what
JS oref had for a divide by zero when autosensMax == 1.

Important: This change will cause Swift's oref to behave differently
from JS. JS will simply default to a ratio of 1 for all cases when
autosesnsMin == 1, with this (very hacky) fix the Swift implementation
will product ratios that adjust as glucose changes (see unit tests).

Moving forward we should strongly consider replacing sigmoid with a
straight line -- they just aren't that different in practice and a
straight line is much easier to implement and understand, which is a
big deal given two hacks to work around numerical issues and some
random 0.0555 in the algorithm. But that will be for a future PR.


trioneer-dev (approved):
LGTM.

PR #1449, opened 2026-08-26
Fix non-atomic append in the log file writer

jwoglom:
SimpleLogReporter seeks to end-of-file and then writes, so anything appended between the two is overwritten. Harmless while it is the only writer of logs/log.txt, but pump manager plugins run in-process and loggerLock does not cover them. O_APPEND plus a single write(2) makes it atomic. appendToWatchLog uses the same extension.


jwoglom:
Validated by building on my test phone with a pump simulator that trio continues to write logs properly with this code. Didn't specifically catch the edge case this is fixing around missing logs, but given that it still writes ok in the normal case I think this is safe to merge

trioneer-dev (approved):

PR #1450, opened 2026-08-26
Add watch "Log Carbs" fallback when recommended bolus is zero

ssabo:
In the watch "Meal & Bolus" flow, the bolus screen's only action was "Enact Bolus", disabled whenever the amount was 0 U. A 0 U recommendation is a normal outcome (low glucose, stale loop, or IOB already covering the meal), so the user hit a dead end: back out to the root, choose "Meal", and re-enter the carbs.

Now, when carbs were carried in from the combo flow and the insulin dial sits at zero, the button becomes "Log Carbs" and logs the carbs directly (matching the standalone "Meal" path). Dialing insulin above zero flips it back to "Enact Bolus" and the existing crown-confirm path is unchanged. The zero check is taken against the increment-snapped amount so floating-point residue from dialing the crown up and back down does not read as a non-zero bolus.

To keep a fast tap from logging carbs before the phone's recommendation arrives, requestBolusRecommendation() now drives the existing (previously unused) "Calculating Bolus..." progress state, cleared on reply, on send error, or after a 3s timeout.

Demo video on simulator: https://www.youtube.com/shorts/o-QPaloI03A


Trio Watch App Extension/Views/BolusInputView.swift
```diff
@@ -23,6 +23,19 @@ struct BolusInputView: View {
...
     var body: some View {
+        let bolusIncrement = Double(truncating: state.bolusIncrement as NSNumber)
+        // Snap to the increment grid so sub-increment floating-point residue left by
+        // dialing the crown up and back down doesn't read as a non-zero bolus.
+        let adjustedBolusAmount = floor(bolusAmount / bolusIncrement) * bolusIncrement
+
+        // In the "Meal & Bolus" flow the user can dial insulin down to zero (or the
+        // recommendation itself is zero). In that case there is nothing to bolus, so
+        // offer a plain "Log Carbs" action instead of a dead-end disabled button.
+        let isCarbsOnly = state.carbsAmount > 0 && adjustedBolusAmount <= 0
```
trioneer-dev, line 32: adjust this to check for bolusAmount

Trio Watch App Extension/Views/BolusInputView.swift
```diff
@@ -23,6 +23,19 @@ struct BolusInputView: View {
...
+        let actionButtonLabel = isCarbsOnly
+            ? String(localized: "Log Carbs", comment: "Button Label to Log Carbs on Watch")
```
trioneer-dev: Semantically, this label makes little sense in the view that is presented to the user. 
The user sees the bolus view, and is presented with “Log Carbs”. 

You’re probably looking for something like “No bolus, log carbs” or something along those lines.

Trio Watch App Extension/Views/BolusInputView.swift
```diff
@@ -23,6 +23,19 @@ struct BolusInputView: View {
     )
 
     var body: some View {
+        let bolusIncrement = Double(truncating: state.bolusIncrement as NSNumber)
+        // Snap to the increment grid so sub-increment floating-point residue left by
```
trioneer-dev: let’s omit the prosa-like, unnecessary comment here.

Trio Watch App Extension/Views/BolusInputView.swift
```diff
@@ -112,13 +122,19 @@ struct BolusInputView: View {
...
+                            navigationPath.append(NavigationDestinations.acknowledgmentPending)
+                        } else {
+                            state.bolusAmount = min(bolusAmount, effectiveBolusLimit)
+                            navigationPath.append(NavigationDestinations.bolusConfirm)
+                        }
                     }
                     .buttonStyle(.bordered)
-                    .tint(Color.insulin)
-                    .disabled(!(bolusAmount > 0.0) || bolusAmount > effectiveBolusLimit)
+                    .tint(isCarbsOnly ? .orange : Color.insulin)
```
trioneer-dev: We’re still in the bolus view. Similar comment to the label one above.
Introducing a color always associated with carbohydrates in Trio makes no sense semantically.

trioneer-dev (dismissed):

t1dude:
Tested with iPhone Air and Apple Watch Ultra 3. Works as expected, letting me log carbs when the bolus recommendation is 0.

t1dude:
@ssabo please have a look at the comments from @trioneer-dev, make some code changes and resolve the issues. Then we can move this to approval :) Thanks for the fix!

Trio Watch App Extension/Views/BolusInputView.swift
ssabo: @trioneer-dev there were conditions I was finding where if the calculation determines a 0u bolus, you dialed up by ~0.30, and then back down to 0.00, the dial increments were not actually occurring in units of 0.05 so you could have a dial value that read 0.00 but still trigged a bolus action because it's marginally larger than 0.00. 

When these conditions happen, it also allows the dial to tick down to -0.05 as a minimum. The later condition was present on `main` without my patch but has probably gone unnoticed because it wouldn't have mattered. 

Since we use `adjustedBolusAmount` when making determinations about how to display the insulin quantity to the user, i was thinking it makes more sense to tie that to how we communicate the action as well. Otherwise we might find confusing situations where the user is rarely offered "enact bolus" instead of "log carbs" for a 0.00u bolus.

t1dude:
Thanks for addressing everything @ssabo ! The only thing missing now before a merge, is to update the string catalog. The changes you've made add at least one new string that needs to be translated. Before it can be translated, it must become part of the string catalog. The way to do that, is to use Xcode to build to a simulator, then wait for all the indexing to finish. You will see that the Localizable.xcstrings file will get a M mark in Xcode. When that is done, you can commit the Localizable.xcstrings to this PR and we will be good to go.

ssabo:
@t1dude localization strings added

t1dude:
> @t1dude localization strings added

Great, let me do a test build and then I'll approve and merge.

t1dude (approved):
Tested, works as expected.

t1dude:
@trioneer-dev or @dnzxy please approve and merge to dev.

PR #1453, opened 2026-08-28
Remove Enlite CGM

MikePlante1:
Medtronic stopped production on their MiniMed Enlite CGM back in 2020, which had a 6-month warranty... so if anyone has any of these sensors today they are very expired and likely very untrustworthy. Furthermore, its functionality has never been validated in Trio, as no Trio developer has had access to the hardware to test with it.


trioneer-dev:
Code-wise LGTM. Can you please pull latest dev into this, build again, to make sure we get an updated String catalog? We just had a ton of new localizations land.

Thanks.

MikePlante1:
> Code-wise LGTM. Can you please pull latest dev into this, build again, to make sure we get an updated String catalog? We just had a ton of new localizations land.
> 
> Thanks.

Latest dev is pulled in. Built to a simulator and there were no String catalog updates. I think the only auto-corrections were thanks to Xcode 27, so I didn't commit them.

trioneer-dev (approved):
LGTM. Merging 🚢

PR #1455, opened 2026-08-28
Feat/chart selection popover

EJM0:
This introduces a new style of the detailed popover that the chart has. It now uses GlassChrome and lives in the MealPanel. The clock now is showen in the x-axis where normaly the hour label live.
- More coherent UI/UX design language with the use of GlassChrome.
- Surrounding data points (BG's, SMB etc . ... ) around the probe in the chart are better viewable.
- Cleaner, more minimalist 

Changes in MainChart architecture:
1. The floating SelectionPopoverView card is deleted; the readout renders as ChartSelectionRow in the Home meal slot instead, so it no longer covers the data under the finger.
2. The selection rule and highlight dots moved to a shell overlay, so scrubbing never re-lays the canvas.
3. Hour labels left the pane's axis for the shell's xAxisOverlay, which lets a scrub swap them for its own time label.
4. The label strip became its own zone, taken off chartHeight before the panes split, with its height measured from a template label so it follows Dynamic Type.
5. xPosition(for:) now applies the live-pinch transform, since shell overlays sit outside the canvas's scaleEffect and would otherwise drift.
 
IMG_4157
IMG_4158

Iphone se (3rd gen iOS 17) (text is on older version only for demonstrating iOS 17 and small form factor)
image


munchyk:
@EJM0 nice!

Two personal preference suggestions:

1. Fade out the other graph components (notification snooze button, chart legend button, basal rate, y axis glucose values) when invoking the long press stats to better focus user attention. Looks like you already started this to some extent.

2. Move the time from the top popover to the bottom of the graph in place of the hour marks on the x axis.

EJM0:
Hi, good remark. But this pr soully concentrates on redesigning the popover and moving it to a nice location.
I think this already cleans things up really well, and don’t think it would be beneficial for the graph to undergo a whole
transformation when inspecting one data point. Hiding other components like the info button or the y-axis labelling is a too small factor, and others might want them to be still enabled while inspecting.

munchyk:
Is hiding the graph overlays a "whole transformation" or a "too small factor"? 😄

It may still be worthwhile to move the time from the popover to the graph’s x axis (in place of current hour markers) because this new popover will be too crowded for users with larger text sizes enabled. See attached mockup from a while back (don’t pay any heed to the graph layout. Was doing some experimentation.)

Just my personal preference. Do as you see fit!

Main Graph - Long Press Overlay

EJM0:
ah is see na you right, would safe space and putting the time label in the x axis would even be semantically more right.

EJM0:
but cost would be that the readability of the time gets worse

EJM0:
image

EJM0:
Bildschirmfoto 2026-08-29 um 18 21 31

dnzxy:
Spacing between elements in the popover is off. 
Imho time belongs in the popover. You are now changing the chart while accessing an overlay type control—why? Don’t do that.

EJM0:
Will change it back.
For the spacing of the elements, are you referring to the bg and smoothed one bc I clubbed them together Intentionally since normal cgm readings had no icon.
And also I kept icon positions static and let text left aligned in respect to them, so that I have a constant spacing while scrolling through values, this also prevents the popover from changing size rapidly.(think it's easier on the eyes)

EJM0:
Other alternatives for bg values

(drop.fill) "bg"|"smoothed bg"
(drop.fill) "bg"

(drop.fill) "bg" ("smoothed bg")
""

These could alow constant spacing but we would trash the smoothed icon

t1dude:
Two issues:

1. For some reason, at one specific point on the chart, the IOB value disappears.
2. The spacing between the glucose icon and the glucose value seems to be a bit large. Is it made for mg/dl?


https://github.com/user-attachments/assets/3a507218-ba07-400d-968a-25695648e10f

EJM0:
@t1dude 
1: I set it up do computer max space that the bg values can occupy. Need to check if it also checks for mmol. For mg/dl
icon 888(888) 
Probably also need to left align it.
2: seems like no iob was computed by oref on that point. How is the old popup showing data on that specific point for you? 
Could you share a screenshot or video in upstream/dev?


Will push a fix for that tmrw:)

t1dude:
For mmol/l values without decimals, we usually show .0, like 8.0. Currently you are not adding the .0 decimal.

Personally I'd add a space between the CGM value and the smoothed value. Now it reads "6.2(5.9)" but I'd change that to "6.2 (5,9)".

I know @dnzxy has some very specific input on spacing as well. 

Apart from that I find it _functionally_ okay :)

EJM0:
Okay, thanks for the feedback. I will implement the changes you mentioned. 
Happy to receive more feedback to follow up on.

t1dude:
Just waiting for some spacing input, then this will be good to go.

dnzxy:
I don't have super specific input.  I can point at* the existing "issues" I see with the current popover:
image

There is no left pad, massive right pad, spacing between iob and cob but not glucose, odd spacing between the blood drop(?) and the glucose value though, and the ✨ got lost  for smoothed values.

t1dude:
Thanks @dnzxy ! If you can address this @EJM0, I can test and approve.

EJM0:
Question: if i let the text determine with of the popover(via fixed spacing at trailing edge), this would result in the pop over rapidly changing width while scrolling wich could be irritating while scrubbing trough values.
With fixed max spacing this stays consistent in width, and thus easier on the eye.
How would you solve this dilemma?
One solution I'd say change width after values are debouched ...

dnzxy:
Since we don't run into this issue with the _current_ popover, I'd say that'll be something for you to figure out, since you're looking to propose a _new_ popover ;)

EJM0:
Damn

EJM0:
IMG_4176
IMG_4175

sould be good now

t1dude:
...testing

t1dude (approved):
Looks good, thanks for all the tweaking :)

PR #1466, opened 2026-09-03
Unify therapy settings editors

RobertDeJong1:
This change started as a small fix for scrolling in the therapy settings editors. After the transition to Liquid Glass, the editor’s hard-coded `.frame` height no longer matched the actual height of the list items, which caused scrolling and layout issues.

While addressing that issue, I noticed that a significant amount of code was duplicated across the different therapy settings editors. I therefore decided to unify them as much as possible.

I moved the shared code into a new `TherapySettingsEditor` module. The existing onboarding steps and editor views now use this shared code.

The change ended up becoming quite a substantial refactor. Since the refactor was already in place, I decided to submit it as a PR rather than leave the changes unused. I'm creating this as a draft PR so we can decide if we want to move forward with this or not.

The resulting structure:
```
Onboarding
- BasalProfileStepView
- CarbRatioStepView
- InsulinSensitivityStepView
- GlucoseTargetStepView
    -> use TherapySettingsEditor.EditingView

Standalone editor root views
- BasalProfileEditorRootView
- CarbRatioEditorRootView
- ISFEditorRootView
- TargetsEditorRootView
    -> use TherapySettingsEditor.RootView

Editor state models
- BasalProfileEditor.StateModel
- CarbRatioEditor.StateModel
- ISFEditor.StateModel
- TargetsEditor.StateModel
    -> conform to TherapySettingsEditor.StateModel

TherapySettingsEditor.RootView
    -> uses TherapySettingsEditor.EditingView

TherapySettingsEditor.EditingView
    -> uses TherapySettingsEditor.ChartView
    -> handles the shared editing logic and data flow

TherapySettingsEditor.ChartView
    -> renders the shared chart for therapy settings
```

Key changes include:
- Moved the shared code into the `TherapySettingsEditor` module, which contains `RootView`, `EditingView`, and `ChartView`.
- Added a Liquid Glass save button to the therapy settings editors through the shared `RootView`.
- Moved the hint "Swipe to delete a single entry. Tap on it to edit its time or value." into `EditingView` so it is also shown during onboarding.
- Fixed the double-scrolling behavior between the root view and the editor’s list view.


trioneer-dev (changes requested):
If you want to truly make this reusable, then I’d suggest to truly remove the four modules basal, CR, ISF, target and only keep a single view module with a therapy settings type based logic handler. I’m not a big fan of this, as it does remove setting-specific logic and forces it into a combined approach.

Right now, you’ve moved from an already existing shared view component – which you deleted – to mixing Trio’s existing approach, and adding a shared state into the mix. 

Either keep what we currently have and move to using just a shared view component, i.e., building on and extending the current `TherapySettingEditorView` (which is my personal recommendation), or move toward a single view module that surfaces different therapy setting UI/UX with shared logic, i.e., similar to what’s done in trio’s Onboarding module.

RobertDeJong1:
Hi,

Please forgive me if I’m misunderstanding your concern, but I believe the current implementation does build on the existing `TherapySettingEditorView`.

The functionality of that view was not removed. Its shared UI and interaction logic were extracted and moved into `TherapySettingsEditor.EditingView`. Because the original file appears as deleted in the diff and the implementation was moved into a new module, this may be difficult to see when reviewing the diff.

I also don’t believe that an additional state layer was introduced. `TherapySettingsEditor.StateModel` is a protocol that allows the existing setting-specific state models to be handled through a common interface by `TherapySettingsEditor.RootView`.

The individual state models remain separate, so they can preserve the necessary differences between targets, basal rates, carb ratios, and ISF. Only the common editor UI and interaction logic have been consolidated.

It was not my intention to remove the separate target, basal, carb ratio, and ISF modules. I agree that setting-specific logic should remain separate so that each type of therapy setting can preserve its own nuances and behavior.

The main goal was to avoid duplicating the same view and interaction code across those modules, while keeping their setting-specific state and logic in place. This also means that the `.frame` height fix only needs to be applied in `TherapySettingsEditor.EditingView`, rather than separately in each editor.

That said, I may be missing an architectural concern behind your comment, so please correct me if I’m interpreting it incorrectly. I also understand that this PR was not originally requested. If we decide that this is not the direction we want to take, I’m fine with that outcome as well.

trioneer-dev:
@RobertDeJong1 apologies for potentially having written this up in a confusing way. 

You introduced this module:
image

Then you extend every single one of the existing four therapy setting modules (basal, CR, ISF, targets) by this `    @Observable final class StateModel: BaseStateModel<Provider>, @MainActor TherapySettingsEditor.StateModel { … }` which we don’t do.

Which brings me to back to my initial comment: either omit the four existing modules, introduce an enum with 4 types, then provide 4 distinct invocations of a single therapy editor module where e.g. you pass said type in the constructor and it handles both view construction, data fetching and state machine accordingly, or you keep a single view `struct` like the existing `TherapySettingEditorView ` and further extend that with your iOS 26/27 fixes and some clean up, and simply update the existing four view modules by potentially a “bigger” instantiation of said struct.

Either/or.

trioneer-dev:
PS: This PR now has conflicts. Please sync with latest dev and make sure to properly incorporate the new accessibility enhancements in your proposal.

RobertDeJong1:
@trioneer-dev thanks for clarifying, I'll have a look into it

PR #1468, opened 2026-09-04
Fix targets editor potentially showing incorrect value of 4.0mmol/L when switching from md/dL to mmol/L

### LINKED ISSUE #1452: When changing glucose unit, the glucose targets are displayed wrong
## Describe the bug
I use mg/dL because I'm used to it but I live in a country that uses mmol/L. On my last doctor's appointment I changed the unit to give my Dr my glucose targets and I noticed that they where wrong. If I change back to mg/dL they original values are displayed.

## To Reproduce
1. Go to Settings
2. Click on Therapy Settings
3. Click on Units and Limits
4. Set Glucose Units to mg/dL
5. Go Back to Therapy Settings
6. Click on Glucose Targets
7. Take note of the entries and convert them, on paper or whenever, to mmol/L
8. Go Back to Therapy Settings
9. Click on Units and Limits
10. Set Glucose Units to mmol/L
11. Go Back to Therapy Settings
12. Click on Glucose Targets
13. Compare the entries displayed with your notes from step 7. Mine were different

## Screenshots
Image
Image

### Smartphone:
* Hardware: *iPhone 15*
* OS Version: *iOS 26.6*

### Pump:
* Manufacturer: *Insulet*
* Model: *Omnipod Dash*

### CGM:
* Device: *Dexcom G7*
* Manager app: *Dexcom App*

### Trio Version:
* Version Number: *0.8.4*
* Repo: *nightscout/trio*
* Git Reference: *29350e3*

## Technical Details

## Additional context

RobertDeJong1:
Fixes #1452

When a target of 83mg/dL is set, this value cannot be found in the rateValues list because 82mg/dL and 83mg/dL both round to 4.6 mmol/L. Only 82md/dL exists in the list, this results in a default rate index of 0, which corresponds to 4.0mmol/L.

This fix adopts the same code `ISFEditor.StateModel` uses to fix this problem, since the issue does not occur there.


Trio/Sources/Modules/TargetsEditor/TargetsEditorStateModel.swift
```diff
@@ -135,5 +135,21 @@ extension TargetsEditor {
                 }
             }
         }
+
+        func resolveRateIndex(rate: Decimal) -> Int? {
+            if let index = rateValues.firstIndex(of: rate) {
+                return index
+            }
+
+            // When using mmol/L, value 83mg/dL cannot be found because 82mg/dL and 83mg/dL both round to 4.6 mmol/L
```
trioneer-dev: unnecessary comment, please remove.

trioneer-dev (dismissed):
Since you’re moving to (partially) copy over logic from ISFEditorStateModel, but chose to extend it (partially), instead of invoking the helper function twice, you could have it return a tuple and then, in a one-liner, have the tuple be assigned to two variables, `lowIndex` and `highIndex`.

Since there is also a draft for a more wide ranging refactor via #1466, I wonder whether this PR is needed.

trioneer-dev:
Since this fixes an intermittent issue, let’s get this in, plus please remove the commented in-line comment @RobertDeJong1 .
Will likely be replaced when/if #1466 gets ready.

t1dude (approved):
I updated the branch and performed the changes requested by @trioneer-dev.

dnzxy (approved):

PR #1469, opened 2026-09-04
Enter remote meals without a note

bjorkert:
Carbs entered from a remote meal command get the note "Remote meal command". Trio uploads the carb note in both notes and foodType, and Nightscout prints foodType as text next to the carb entry on the chart, so every remotely entered meal ends up labeled on the graph.

This drops the note, so remote meals render like any other carb entry.


t1dude (approved):
Tested in simulator + NS as I don't log carbs. 

LGTM.

dnzxy:
Changing the message to something shorter vs. no longer reporting the message seems like a good compromise, otherwise we just drop the carb announcement source (for caregivers) entirely.

bjorkert:
The message is now replaced with a 📡 - satellite antenna as a symbol for remote command.

dnzxy (approved):

PR #1470, opened 2026-09-05
Plot glucose on the top chart layer so carbs and boluses no longer hide it 

### LINKED ISSUE #1451: UI: Plot BG readings on highest layer of main graph
Trio v0.8.4.77 dev 632772a

Here, CGM data is hidden under the carb triangle. The trend is difficult to observe, and it looks like data is missing. Please plot CGM data on top of other chart elements. 

Image

---

This graph is more zoomed in, and the pattern can be observed. (That is, the data is/was available.)
Image

trioneer-dev:
Self explanatory.

Fixes #1451.


t1dude (approved):
Tested in simulator as I don't log carbs. LGTM.

MikePlante1 (approved):
Code and simulator test look good. 

I added some l18n missed from #1375

PR #1472, opened 2026-09-06
fix: store calibrations as manual fingersticks

bastiaanv:
Currently, Eversense and AccuChek do not emit the calibrations to the app. I'm planning on changing this, but Trio records the manual readings as CGM readings.

This PR changes that by looking at the `wasUserEntered` property of [NewGlucoseSample](https://github.com/LoopKit/LoopKit/blob/dev/LoopKit/GlucoseKit/NewGlucoseSample.swift#L28). If this value is true, it will mark that reading as a manual glucose (adding the nice red waterdrop icon to the Trio main chart)


dnzxy:
Code-wise this looks good to me. 
How would this currently work with CGMs that are not directly integrated into Trio like Dexcom’s CGMs or xDrip4iOS (which doesn’t use the NewGlucoseSample pathway)?

t1dude:
@bastiaanv could you address the comment from Dan, then we can maybe move this one into dev? :)

bastiaanv:
CGM list:
- G7SensorKit -> uses [the same Bool](https://github.com/LoopKit/G7SensorKit/blob/main/G7SensorKit/G7CGMManager/G7CGMManager.swift#L503) for `isDisplayOnly` and `wasUserEntered`. This is lead to the glucose being dropped. No calibrations are emitted
- Libre CGM's cannot be calibrated
- xDrip does not emit BGCheck to OS-AID, only directly to Nightscout via the API
- NightScout remote CGM -> I made a small change, fixing this integration. Trio already uses the nightscout api model with the extra glucose property. The glucose property is now either sgv or mbg

dnzxy:
~~Might be display bug on GH, but has all indentation on the added code been dropped somehow?~~
Never mind, Github having issues apparently.

bastiaanv:
@dnzxy Any concerns on this PR?

dnzxy:
@bastiaanv , I briefly reviewed it and found issues with this with the passive consumption CGMs (Dexcom) where the assumption of using `wasUserEntered ` as a flag to identify MBG values wasn't true, but ran out of time to pursue it further with v1.0.0, the following fixes / clean-up etc. and today's release (and prior preparation of 1.0.1) — apologies :(

PR #1473, opened 2026-09-06
Adding Glucose Bobble as an optional "Contact Trick" image design.

t1dude:
While testing out the direct Dexcom driver for Trio, I found myself missing the Dexcom widget on my Apple Watch. I usually have it positioned as a very visible glucose badge. I enabled the "Contact Trick" feature and saw that the existing design templates could use some love.

As an _optional feature_ I added the possibility to select a new layout called "Glucose Bobble" **(image 1)**. When selected, the rest of the options are hidden and replaced with separate options: Color Mode, show/hide minutes since last reading, and show/hide delta value. **(image 2)** After testing it for a while I found that colored glucose values have lower readability at night (without glasses...) so I added an option to change between colors and monochrome. **(image 3)**. Finally I added some information to the (i) sheet. **(Image 4)**

This might all be replaced by a future direct-to-watch integration, but for now I think it deserves a place as an optional design choice.

1 - Layout

2 - Options

3 - Color modes

4 - Help

Apple Watch Screenshot


Model/TrioCoreDataPersistentContainer.xcdatamodeld/TrioCoreDataPersistentContainer.xcdatamodel/contents
```diff
@@ -26,6 +26,8 @@
         </fetchIndex>
     </entity>
     <entity name="ContactImageEntryStored" representedClassName="ContactImageEntryStored" syncable="YES" codeGenerationType="class">
+        <attribute name="bobbleShowDelta" optional="YES" attributeType="Boolean" defaultValueString="YES" usesScalarValueType="YES"/>
```
dnzxy, line 29: Should these really default to `YES` rather as they are optionals? Not nil?

dnzxy (commented):
Functionality wise, works! 👏 

I wonder if the triangle of the bobble for the watch face is too small and needs to be made slightly bigger.

If we compare the watch-contact-bobble vs. actual home screen or watch app bobble, you can see the difference in the triangle, and I wonder if that's too small and also just looks a little odd.

| Phone App | Watch App | Watch Contact |
|--------|--------|--------|
| Image | Image | Image |

Model/TrioCoreDataPersistentContainer.xcdatamodeld/TrioCoreDataPersistentContainer.xcdatamodel/contents
t1dude: Same pattern as isSMB in line 6: optional="YES" + defaultValueString="YES" + scalar. 

With usesScalarValueType="YES" the generated Swift property is a non-optional Bool, so it can never be nil anyway?

t1dude:
incoming-EAB03299-8B75-4D27-87DC-9E34A1DBE83F


Arrow is now slightly bigger, making it easier to see the direction. I'm trying to keep the circle as large as possible, that's why the arrow is still a bit smaller compared to the aspect ratio in the phone app HUD.

Trio/Sources/Services/ContactImage/ContactPicture.swift
```diff
@@ -301,6 +305,62 @@ struct ContactPicture: View {
...
+            return 0
+        }
+    }
+
+    private static func makeGlucoseBobbleImage(
+        contact: ContactImageEntry,
+        state: ContactImageState,
+        pixelSize: CGFloat
+    ) -> UIImage {
+        let hasReading = state.glucose != nil
```
dnzxy, line 346: This can never be false after the first CGM reading, so the `– –` / gray stale path will get stuck when CGM drops out, and the contact photo (will) keep(s) rendering a stale value in full glucose color.

dnzxy (commented):
Triangle is better now, thanks for changing that.

Trio/Sources/Services/ContactImage/ContactPicture.swift
t1dude: It looks like this issue was there already in ContactImageManager.swift and also affected the other contact image styles. Tried to fix by clearing state.glucose around line 223. Should work for all styles.

dnzxy (approved):

PR #1475, opened 2026-09-07
Add new zoom gesture and other tweaks

EJM0:
This pr reworks some of the gestures controlling the main chart.

1. I want to propose to change the anchoring of the double-tap stepped zoom (6h-12h-24h) to the centre where the double taps were evoked, instead of the trailing edge of the chart. (currently in the code)
2. Added a new gesture double-tap hold and swipe which is an alternative to the pinch zoom gesture for easier one-handed use. (example: https://www.reddit.com/r/iphone/s/T4kmwUrFkq) Vanilla gestures all remain the same. If the user does not confirm on zoom (stays in 10p bound) after 0.5s, the app also still reverts to inspection.
3. Zoom rate now varies across the zoom range. A zoom no longer costs the same finger travel per doubling everywhere: ~180 pt halves or doubles the window at the 1 h end, tapering to ~72 pt at the 24 h end. That puts the whole 1 h → 24 h range within a single pull — roughly 500 pt of travel instead of the ~825 pt a flat rate needed — so zooming all the way out no longer takes a re-grab. Fine control at the tight end is unchanged.
~~4. Added gesture info sheet to the graphs info button~~


t1dude:
Personally I think these changes are not necessary, but the gestures explanation is a good addition. Maybe close this PR and add a new one that only adds gesture help to the existing (i) button?

EJM0:
I’d like to see the double tap swipe in trio, for people who know the gesture(viewable in info), this could really improve navigation comfort. I'll open a separate pr for the gesture info. But maybe some more opinions on whether this is worth the added code "complexity" would be great.
Also I could reduce the pr only to the new zoom gesture to make it even leaner.

EJM0:
Closing in favor of keeping framework driven approach.


... praying for apple charts to get better 🥲

PR #1476, opened 2026-09-07
Fix fractional Nightscout glucose dates

premnirmal:
Fixes https://github.com/nightscout/Trio/issues/1478

## Summary

- round stored glucose timestamps to integer milliseconds before Nightscout upload
- use the same timestamp for `date` and `dateString`
- cover both sensor and manual glucose records

## Problem

`getGlucoseNotYetUploadedToNightscout()` converted `Date.timeIntervalSince1970` to `Decimal` before multiplying by 1000. This preserved sub-millisecond precision and produced values such as `1788785143203.7996` in Nightscout.

Nightscout can return that value as `lastModified.collections.entries`, while clients expect an integer millisecond timestamp.

## Testing and verification
- Added a regression test with a fractional timestamp. It verifies that sensor and manual Nightscout records use the rounded integer millisecond value and keep the same `dateString` instant.
- I also verified this by running my Trio build and inspecting the entries uploaded to my own nightscout instance


dnzxy:
Hi,

thanks for this proposal. What _issue_ is this solving?

premnirmal:
Apologies, linked the issue https://github.com/nightscout/Trio/issues/1478

dnzxy (approved):

dnzxy:
Thanks for the added context. 
So this is actually a cross-platform issue. 
Make sense to change this like proposed.
Thanks!

dnzxy:
Please stop merging dev over an over again for sync 😅 it always requires a new CI test run and I was just waiting for that to go green (which the last one did).

dnzxy (approved):

premnirmal:
Sorry about that!

dnzxy (approved):

PR #1477, opened 2026-09-07
Fix: register UserNotificationsManager in container scope

nicolevanelstvandenhoek:
Without .inObjectScope(.container), Swinject did not keep a persistent reference to UserNotificationsManager after it was resolved once in TrioApp.swift, causing ARC to deallocate it and silently drop its Combine subscription to new glucose readings (no further badge updates). This matches the scope already used for the sibling background services (TrioAlertManager, GlucoseAlertCoordinator, NotLoopingMonitor).


dnzxy:
Hi,

thanks for this proposal. What _issue_ is this solving? 

Note: I don’t mean the LLM summary of what it thinks is wrong and what it then sought out to change, but an actual issue you are attempting to fix here. 

Thanks!

nicolevanelstvandenhoek:
Hi and thanks for your reply! 

The app badge on the Trio icon does not appear. Not even when I select it at onboarding. 

Kind regards Nicole

dnzxy:
Hi Nicole,

sorry, which badge are you referring to? Glucose app badge?

nicolevanelstvandenhoek:
Yes, the one on the Trio icon

dnzxy:
Okay, and what's your issue with that?
If I enable that in settings, I see it on my app badge.

t1dude:
If there is an issue with the glucose badge on the Trio app icon, please register it as an **issue** on GitHub, describing what is wrong and steps to reproduce the error.

Issue #1146, opened 2026-05-11
TTs and overrides in the past (>24 hours)


bjornoleh:
## Describe the bug
Trio don’t fetch TTs or exercise events (Overrides) >24 hours old.

When a TT start date is > 24 hours in the past, the TT can not be cancelled from the Trio main screen. Possibly TTs can still be terminated from Adjustments/TT, but not after a restart of Trio.

The temp target is still active as seen in loop status and NS OpenAPS pill. The TT is rendered in both NS and Trio charts (removed from chart after restart of Trio).

A new TT in Trio has the maximum duration of 23:59, but TTs can be started from NS with no such limits for duration.

When testing with a TT with duration of seven days (10 000 minutes), the folowing temptargets.json persists, but cannot be terminated from the UI:
[
  {
    "duration" : 10000,
    "_id" : "6a01bb34b464ba8a0482ff9e",
    "created_at" : "2026-05-10T11:30:00.000Z",
    "targetBottom" : 144.12472,
    "reason" : "Activity",
    "targetTop" : 144.12472,
    "enteredBy" : "test"
  }
]

After deleting temptargets.json:
Profile target is used, and not the temporary target.

## To Reproduce
*Steps to reproduce the behavior:*
1. Start a temp target from NS with duration > 24 hours. It can be backdated with a start date close to 24 hours in the past to set up for quick and easy testing without the need to wait 24 hours.
2. Verify that Trio picks up the TT
3. Notice how the TT adjustment can no longer be terminated from Trio, although the temporary target still is applied (check loop info or OpenAPS pill in NS)
4. The TT is still rendered in the chart, but a restart of the Trio app will remove the TT rendering
5. Delete Trio/settings/temptargets.json to terminate the TT

Similar issues exist with long running (infinite/30 days) overrides. The associated `exercise`NS event is not parsed and hence can not be edited when the start date is > 24 hours in the past. As a concequence, the Override appears to be active in NS althpough it has been terminated in Trio. 

In summary, TTs and are Overrides are affected in opposite ways:
- TTs could still be applied in Trio although they are canceled from NS
- Overrides might appear to be active in NS, but are cancelled in Trio
- Overrides can also not be terminated from LF remote when the start date is older than 24 hours (I am unsure about the mechanism here)

## Screenshots
*If applicable, add screenshots to help explain your problem.*

### Trio Version:
* Version Number: *[e.g. 0.7.0.14]*
* Repo: *nightscout/trio*
[...]

PR #1482, opened 2026-09-08
Centralize override and temp target activation

bjorkert:
This rewrite was done because overrides activated from the watch were not sent to Nightscout until they were cancelled. The change has successfully been tested in real-world use for several days with no issues.

Shortcuts, remote control and the watch route activation and cancellation through AdjustmentManager, which performs the whole sequence in one Core Data transaction: end every enabled adjustment and record each as a run entry, enable the requested one, then apply the side effects in a fixed order. Commands are serialized, so two arriving at once cannot interleave. The Adjustments view keeps its own write path in this PR.

The run entry carries the elapsed duration to Nightscout and draws the finished band on the home chart. uploadOverrideRuns deletes the entry at the run's own created_at and re-posts it there, so the cleanup holds for a preset row whose date moves on every activation, and the run upload window keys on endDate so an adjustment that ran for more than a day still retracts its placeholder. deleteNightscoutOverride filters on event type.

APSManager.determineBasalSync() is the single entry point for a determination outside the loop. It waits for a loop in flight to finish, collapses concurrent calls into one, never delays a loop start, and keeps throwing so the treatment views still see algorithm errors. The manager and the Adjustments view models go through it.

A custom, non-preset adjustment is stored disabled and activated by object ID, so storeTempTarget returns the inserted row's NSManagedObjectID.

The watch acknowledges a command from the returned outcome, and its failure acknowledgment applies only to messages carrying none of the recognized payloads. A remote cancel with nothing running is the state the command asked for.

### DIFF
--- a/Trio/Sources/APS/APSManager.swift
+++ b/Trio/Sources/APS/APSManager.swift
@@ -30,6 +30,10 @@ protocol APSManager {
     func enactTempBasal(rate: Double, duration: TimeInterval) async
     func determineBasal() async throws
     func determineBasalSync() async throws
+    /// Recomputes the determination after something the algorithm reads has changed — an override,
+    /// a temp target. Skips itself while a loop is running, because that loop determines from the
+    /// same state, and never delays a loop start.
+    func recomputeDetermination() async
     func simulateDetermineBasal(
         simulatedCarbsAmount: Decimal,
         simulatedBolusAmount: Decimal,
@@ -76,6 +80,7 @@ enum APSError: LocalizedError {
 /// Ensures only one loop runs at a time via actor isolation
 private actor LoopGuard {
     private var isRunning = false
+    private var isRecomputing = false
 
     /// Atomically checks whether a new loop can start and marks it as running if so.
     func tryStart(minInterval: TimeInterval, lastLoopDate: Date, lastLoopStartDate: Date) -> Bool {
@@ -90,6 +95,19 @@ private actor LoopGuard {
 
     func finish() {
         isRunning = false
+    }
+
+    /// Claims the guard for a determination-only recompute. Refuses while a loop runs, and while
+    /// another recompute runs — a second one would compute the same state. `tryStart` ignores this
+    /// claim: a skipped loop has a therapy cost, an overlapping recompute only wastes work.
+    func tryStartRecompute() -> Bool {
+        guard !isRunning, !isRecomputing else { return false }
+        isRecomputing = true
+        return true
+    }
+
+    func finishRecompute() {
+        isRecomputing = false
     }
 }
 
@@ -556,6 +574,19 @@ final class BaseAPSManager: APSManager, Injectable {
 
     func determineBasalSync() async throws {
         _ = try await determineBasal()
+    }
+
+    func recomputeDetermination() async {
+        guard await loopGuard.tryStartRecompute() else {
+            debug(.apsManager, "Determination recompute skipped: loop or recompute in flight")
+            return
+        }
+        do {
+            try await determineBasal()
+        } catch {
+            debug(.apsManager, "Determination recompute failed: \(error)")
+        }
+        await loopGuard.finishRecompute()
     }
 
     func simulateDetermineBasal(
--- a/Trio/Sources/APS/APSManager.swift
+++ b/Trio/Sources/APS/APSManager.swift
@@ -29,11 +29,11 @@ protocol APSManager {
     var isSuspended: Bool { get }
     func enactTempBasal(rate: Double, duration: TimeInterval) async
     func determineBasal() async throws
+    /// Runs a determination outside the loop, after a treatment or adjustment changed what the
+    /// algorithm reads. Waits for a loop in flight to finish first, since that loop determined
+    /// before the change landed. Concurrent calls collapse into the one already running. Algorithm
+    /// errors propagate to the caller.
     func determineBasalSync() async throws
-    /// Recomputes the determination after something the algorithm reads has changed — an override,
-    /// a temp target. Skips itself while a loop is running, because that loop determines from the
-    /// same state, and never delays a loop start.
-    func recomputeDetermination() async
     func simulateDetermineBasal(
         simulatedCarbsAmount: Decimal,
         simulatedBolusAmount: Decimal,
@@ -80,7 +80,8 @@ enum APSError: LocalizedError {
 /// Ensures only one loop runs at a time via actor isolation
 private actor LoopGuard {
     private var isRunning = false
-    private var isRecomputing = false
+    private var isDeterminingStandalone = false
+    private var loopWaiters: [CheckedContinuation<Void, Never>] = []
 
     /// Atomically checks whether a new loop can start and marks it as running if so.
     func tryStart(minInterval: TimeInterval, lastLoopDate: Date, lastLoopStartDate: Date) -> Bool {
@@ -95,19 +96,28 @@ private actor LoopGuard {
 
     func finish() {
         isRunning = false
+        let waiters = loopWaiters
+        loopWaiters.removeAll()
+        waiters.forEach { $0.resume() }
     }
 
-    /// Claims the guard for a determination-only recompute. Refuses while a loop runs, and while
-    /// another recompute runs — a second one would compute the same state. `tryStart` ignores this
-    /// claim: a skipped loop has a therapy cost, an overlapping recompute only wastes work.
-    func tryStartRecompute() -> Bool {
-        guard !isRunning, !isRecomputing else { return false }
-        isRecomputing = true
+    /// Returns once no loop is running.
+    func waitForLoop() async {
+        guard isRunning else { return }
+        await withCheckedContinuation { loopWaiters.append($0) }
+    }
+
+    /// Claims the guard for a determination that runs outside the loop. Refuses while another one
+    /// runs, since it computes the same state. `tryStart` ignores this claim: a skipped loop has a
+    /// therapy cost, an overlapping determination only wastes work.
+    func tryStartStandaloneDetermination() -> Bool {
+        guard !isDeterminingStandalone else { return false }
+        isDeterminingStandalone = true
         return true
     }
 
-    func finishRecompute() {
-        isRecomputing = false
+    func finishStandaloneDetermination() {
+        isDeterminingStandalone = false
     }
 }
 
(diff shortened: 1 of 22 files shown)


bjornoleh:
>  ... the run upload window keys on endDate so an adjustment that ran for more than a day still retracts its placeholder. deleteNightscoutOverride filters on event type.
> 

Not sure if I read this correctly, but does this mean it will also fix https://github.com/nightscout/Trio/issues/1146?

bjorkert:
Partially... the endDate window means a run that started more than a day ago now gets uploaded when it ends, so the stale Nightscout entry is deleted and re-posted with the real duration. That is the "override cancelled in Trio but still shown as running in NS" half of #1146. The rest is untouched.

TrioTests/CoreDataTests/TestAssembly.swift
```diff
@@ -40,5 +40,11 @@ class TestAssembly: Assembly {
         container.register(OverrideStorage.self) { r in
             BaseOverrideStorage(resolver: r, contextProvider: { self.testContext })
         }.inObjectScope(.container)
+
+        // Override AdjustmentManager registration for tests: same context as the storages, and the
```
dnzxy: Let's shorten this please.

TrioTests/AdjustmentsTests/AdjustmentManagerTests.swift
dnzxy, line 1: You've changed the predicate for both "not yet uploaded" fetches of the "xxxRunStored" entities.
Is this change, and especially that the new predicate still covers everything the old one did, and beyond fixes apparent issues, covered by automated testing? If not, it probably should.

Trio/Sources/Services/Adjustments/AdjustmentManager.swift
```diff
@@ -0,0 +1,487 @@
+import CoreData
+import Foundation
+import Swinject
+
+// MARK: - Public surface
+
+/// How a caller names the adjustment to activate.
+enum AdjustmentRef: Equatable {
+    /// A row the caller already holds. Saved rows only: a row that has never been saved carries a
```
dnzxy: stop sign comment

Trio/Sources/Services/Adjustments/AdjustmentManager.swift
```diff
@@ -0,0 +1,487 @@
...
+    /// temporary object ID, which resolves in no other context.
+    case objectID(NSManagedObjectID)
+    /// The preset's own identifier — `OverrideStored.id` / `TempTargetStored.id`.
```
dnzxy: stop sign comment

Trio/Sources/Services/Adjustments/AdjustmentManager.swift
```diff
@@ -0,0 +1,487 @@
...
+    case presetID(String)
+    /// The preset's display name, compared with surrounding whitespace trimmed on both sides.
```
dnzxy: stop sign comment

Trio/Sources/Services/Adjustments/AdjustmentManager.swift
```diff
@@ -0,0 +1,487 @@
...
+    /// Names are free text and may repeat, so the lowest order position wins.
+    case presetName(String)
+}
+
+/// The entry point a command came from. Travels into the log line and the Nightscout upload
```
dnzxy: self explanatory

Trio/Sources/Services/Adjustments/AdjustmentManager.swift
```diff
@@ -0,0 +1,487 @@
...
+/// request, and is the only thing the manager knows about its caller.
+enum AdjustmentSource: String {
+    case app
+    case remote
+    case watch
+}
+
+/// What one command started or ended. Callers build their own acknowledgment, log line or toast
```
dnzxy: self explanatory

Trio/Sources/Assemblies/ServiceAssembly.swift
```diff
@@ -36,5 +36,9 @@ final class ServiceAssembly: Assembly {
         }
         container.register(IOBService.self) { r in BaseIOBService(resolver: r) }
         container.register(BolusSafetyValidator.self) { r in BaseBolusSafetyValidator(resolver: r) }
+        // Container scope: the manager serializes adjustment mutations across every entry point,
```
dnzxy: not needed

Trio/Sources/APS/Storage/OverrideStorage.swift
```diff
@@ -289,7 +289,7 @@ final class BaseOverrideStorage: @preconcurrency OverrideStorage, Injectable {
             ofType: OverrideRunStored.self,
             onContext: context,
             predicate: NSPredicate(
-                format: "startDate >= %@ AND isUploadedToNS == %@",
+                format: "endDate >= %@ AND isUploadedToNS == %@",
```
dnzxy, line 292: side effects?

Trio/Sources/APS/Storage/TempTargetsStorage.swift
```diff
@@ -336,7 +339,7 @@ final class BaseTempTargetsStorage: TempTargetsStorage, Injectable {
             ofType: TempTargetRunStored.self,
             onContext: context,
             predicate: NSPredicate(
-                format: "startDate >= %@ AND isUploadedToNS == %@",
+                format: "endDate >= %@ AND isUploadedToNS == %@",
```
dnzxy, line 342: side effects?

Trio/Sources/APS/APSManager.swift
```diff
@@ -30,6 +30,10 @@ protocol APSManager {
     func enactTempBasal(rate: Double, duration: TimeInterval) async
     func determineBasal() async throws
     func determineBasalSync() async throws
+    /// Recomputes the determination after something the algorithm reads has changed — an override,
+    /// a temp target. Skips itself while a loop is running, because that loop determines from the
+    /// same state, and never delays a loop start.
+    func recomputeDetermination() async
```
dnzxy: `func determineBasalSync() async throws` essentially does the same as the newly introduced method `func recomputeDetermination() async`, less the added logging, plus it throws, so it propagates algorithm errors to super layers. 

I like the proposed new logic, and also the `tryStartRecompute()` check, but it no longer throws.
The condition function name may need some polishing, because I'm personally a fan of self-descriptive code, and here I may ask myself in a few months "tryStartRecompute of what ?", it catches the algo run's error and only log-surfaces it. 

I'd suggest to omit one of the two functions invoking the algo here, streamlining the approach, but ensuring we still throw, because said throwing is needed elsewhere for reactive UI behavior.

dnzxy (commented):
Some minor comments about – no pun intended – comments, some questions, and a slightly bigger ask about streamlining potentially duplicate logic.

Overall great proposal and deduplicating A LOT of logic, kudos. 

Thanks for tackling this, Jonas!

Trio/Sources/APS/Storage/OverrideStorage.swift
bjorkert: The new predicate is a superset of the old one: a run's endDate is never before its startDate, so every run the startDate window matched is still matched. The only new behavior is that a run which started more than a day ago but ended within the last day is now uploaded. uploadOverrideRuns then deletes the Nightscout entry at that run's created_at (its startDate) and re-posts it with the real duration. That is exactly the case where a long override was cancelled in Trio but kept showing as running in Nightscout (#1146). Nothing else reads these fetches. Covered by testGetOverrideRunsNotYetUploadedToNightscout and testGetTempTargetRunsNotYetUploadedToNightscout.

Trio/Sources/APS/Storage/TempTargetsStorage.swift
bjorkert: The new predicate is a superset of the old one: a run's endDate is never before its startDate, so every run the startDate window matched is still matched. The only new behavior is that a run which started more than a day ago but ended within the last day is now uploaded. uploadTempTargetRuns then deletes the Nightscout entry at that run's created_at (its startDate) and re-posts it with the real duration. That is exactly the case where a long override was cancelled in Trio but kept showing as running in Nightscout (#1146). Nothing else reads these fetches. Covered by testGetOverrideRunsNotYetUploadedToNightscout and testGetTempTargetRunsNotYetUploadedToNightscout.

Trio/Sources/APS/APSManager.swift
bjorkert: Done: recomputeDetermination is gone and determineBasalSync is the single entry point. It keeps throwing, so the treatment views still get algorithm errors. It now claims the guard as tryStartStandaloneDetermination, and instead of skipping while a loop is in flight it waits for the loop to finish and then runs, because a loop mid-enact has already published a determination that predates the change the caller just made, and the views that show a spinner only clear it on the next determinationDidUpdate. Concurrent standalone calls still collapse into one, and a loop start never waits on it.

TrioTests/AdjustmentsTests/AdjustmentManagerTests.swift
bjorkert: Added testGetOverrideRunsNotYetUploadedToNightscout and testGetTempTargetRunsNotYetUploadedToNightscout in the storage test suites; each seeds a run inside the old window, one that started three days ago and ended an hour ago, one that ended two days ago, and one already uploaded.

t1dude:
Could you resolve the branch conflicts, @bjorkert ? Let's try to get this through.

dnzxy (approved):

bjornoleh (commented):
@bjorkert  I notice this is already merged, cool. Just wondering if temp targets are also covered now, it looks that way from recent commits?

bjorkert:
Yes. AdjustmentManager handles overrides and temp targets the same way.

bjornoleh:
> Yes. AdjustmentManager handles overrides and temp targets the same way.

Good! And does this then resolve all of https://github.com/nightscout/Trio/issues/1146, or should this be kept open?

bjorkert:
Keep it open. This PR only fixes the first of the three problems in #1146: an override or temp target that Trio started, ran for more than a day and then ended is now re-posted to Nightscout with its real duration, so it stops showing as running there.. the other two are untouched. A temp target started from Nightscout with a duration over 24 hours still cannot be cancelled from Trio once its start is more than a day old, and temptargets.json still persists in that case, because the active-adjustment lookups and the oref file handling were not part of this change. Nightscout exercise entries older than 24 hours are still not parsed, so they still cannot be edited from Trio. Let's leave #1146 open for those two, and note in it that the "cancelled in Trio, still running in NS" half is done?

PR #1489, opened 2026-09-11
Fix empty PluginSource fetch results

andrewtcombs:
## Summary
- Return the first plugin fetch result even when it is empty, instead of filtering out the sole callback value.
- Emit `[]` on empty completion or timeout using `replaceEmpty(with: [])`; keep first-result cancellation, five-minute timeout, and defensive error replacement.
- Add a standalone macOS native-Combine contract regression suite and correct the fetch documentation.

## Why
`fetchIfNeeded()` currently resolves a single-result Future with `[]`. Its callback intentionally ignores share-client values (see #703); actual CGM readings/backfill arrive separately through the manager delegate. Filtering the callback's only output makes the fetch complete without a value, rather than delivering the documented empty result. With `customError: nil`, timeout also completes without a value unless explicitly replaced.

This intentionally changes empty-result observability. The timer subscriber now enters its existing semaphore/task path, including starting/ending a background task. Source review shows its empty-data guard returns before glucose storage or pump heartbeat scheduling. Delegate delivery, share-client suppression, and timer configuration are unchanged. Device-level scheduling still needs maintainer/integration validation.

## Verification
On macOS, Xcode 26.6 / Apple Swift 6.3.3, with warnings as errors:
- `python3 scripts/tests/plugin-fetch/run.py` — 12 tests, 0 failures.
- `python3 scripts/tests/plugin-fetch/run.py --full-timeout` — 12 tests, 0 failures; silent-publisher test exercised the unchanged production timeout for 300.007 seconds.
- Same suite against pre-change `a708381782825c03745db1c6377efe632b5f2161` source — 7 expected failures out of 12 tests, covering the changed empty-result contract.
- `git diff --cached --check` — passed.

Coverage includes empty array/completion, nonempty result, defensive error replacement, timeout, first-result/upstream cancellation, downstream cancellation, finite demand, late values/errors, and asynchronous Future delivery.

## Boundaries / review attention
The suite extracts the current production method into a test-only host with synthetic integer elements and an injected publisher. It is not an app/driver integration test and is not wired into TrioTests or CI. The generic error case probes a defensive operator; production `fetchIfNeeded()` has `Failure == Never`.

The reported `Collection.isEmpty` / Combine Filter `EXC_BAD_ACCESS` has **not** been reproduced. This removes that closure and verifies empty-fetch semantics; it does not establish that an underlying memory-lifetime or concurrency defect is fixed. No full iOS app build, device testing, or deployment was performed.

AI-assisted preparation and manual source/diff review were used for this narrow change. No dosing, bolus, dependency, signing, or deployment changes are included.

### DIFF
--- a/Trio/Sources/APS/CGM/PluginSource.swift
+++ b/Trio/Sources/APS/CGM/PluginSource.swift
@@ -47,18 +47,18 @@ final class PluginSource: GlucoseSource {
         cgmProgressHighlight.value = cgmManager?.cgmLifecycleProgress
     }
 
-    /// Function that fetches blood glucose data
-    /// This function combines two data fetching mechanisms (`callBLEFetch` and `fetchIfNeeded`) into a single publisher.
-    /// It returns the first non-empty result from either of the sources within a 5-minute timeout period.
-    /// If no valid data is fetched within the timeout, it returns an empty array.
+    /// Returns the first fetch result, including an empty array, within five minutes.
+    /// Empty completion, timeout, or failure produces an empty array.
+    /// Plugin readings and backfill arrive separately through the CGM manager delegate;
+    /// the fetch callback is not a second source of glucose readings.
     ///
     /// - Parameter timer: An optional `DispatchTimer` (not used in the function but can be used to trigger fetch logic).
     /// - Returns: An `AnyPublisher` that emits an array of `BloodGlucose` values or an empty array if an error occurs or the timeout is reached.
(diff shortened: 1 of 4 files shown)


t1dude:
Both the code and the PR text have typical AI generated patterns. Could you please explain in plain language what is the issue you are trying to solve, and how you are solving it?

andrewtcombs:
Yes, this was AI-assisted. We traced the crash to the filter but haven’t found the root cause; the tests verify a behavior change, not a crash fix. I submitted it too early.

PR #1496, opened 2026-09-12
Merging alpha branch to dev branch

yahyakaptan0617-bit:


yahyakaptan0617-bit:
So let’s change the UI & UX and add more pumps like equil and delete medtronic 780g

PR #1499, opened 2026-09-12
Stop the loop pill claiming to loop in open loop

### LINKED ISSUE #643: UX: confusing/incorrect display of loop status in open loop
v0.5.0.38

State: Trio is looping, and (in my case) was 3 minutes into a 30-minute temporary rate
UX displays a green circle and (in my case) "3 m".

Image

***
Action: Set Trio to open loop

***
State 1 minute later: Trio is not looping
UX displays: a grey circle with a horizontal line through its centre... and, say "4 m".

❓Question: is it incorrect that the time shows 4m?  

Image

***
State another 1 minute later: Trio is not looping
UX displays: a grey circle with a horizontal line through its centre... and the word "looping" (because a new BG was just received).

Image

❓Question:  Should the word "looping" be surpressed?  Because Trio is't actually looping?

***
State yet another minute later: Trio is not looping
UX displays: a grey circle with a horizontal line through its centre... and "<1 m" because a BG was received less 
than a minute ago.

Image

***
Trio continued to cycle through these states (4 m, 5 m, looping, <1m, 1 m, 2 m, 3 m) *until the temporary basal rate Trio set 27 minutes ago expired*.  

***
State: last temporary basal rate set by Trio has expired.
UX displays: a grey circle with a horizontal line through its centre... and "--".

❓Question:  Should Trio display "--" from the moment the loop is opened?  Or, perhaps better still, calculate and display the length of time the user has been in open loop?

Image

trioneer-dev:
Closes #643. Stacked on #1300 and the PR below it.

The pill was permanently grey and never signalled staleness, while the status sheet said "Enacted at …" when nothing had been enacted — both caused by a guard on `determination.timestamp`, which only `reportEnacted` sets, so it is never set in open loop.

Open loop now reports device health: green while CGM and pump are reporting, red on a device issue. The time caption is dropped outside closed loop, where it would imply an action that did not happen.

The glyph becomes a drawn ring with horizontal gaps — closed only in closed loop — with a centre symbol for the constrained modes. It scales with Dynamic Type.


trioneer-dev:

PR #1500, opened 2026-09-12
Add Low Glucose Suspend and Basal Testing modes

### LINKED ISSUE #644: Trio basal rates seem inappropriate/are not logged to history when Closed Loop is toggled off
v0.5.0.38

State: Trio is looping, and set a 30-minute temporary basal rate (0.95 u/hr) three minutes ago.

Image

***
Action: Toggle Closed Loop off.

***
Observe that Trio continues to run (states it is running?) the Trio-determined temporary basal rate.

❓Question: should this happen?  Or should the basal rate revert to the programmed rate?

Image

***
Wait out Trio's last-programmed temporary basal rate.  Observe that Trio displays *no* basal rate in its graph, and no basal rate in its history. (This one is concerning, as the graph and history make it look like Trio has stopped delivering insulin.)

❓Question: If not when toggling off Closed Loop, shouldn't the basal rate revert to the programmed rate here, and be entered in the log?  And the basal rate should be shown (perhaps in another color?) in the main screen?

Image

***
Re-enable Closed Loop.  Observe that Trio now backfills the basal rate in its graph, but it does so at the last temporary basal rate Trio calculated.  The history is not updated to show the basal rate used between when the previous temporary rate expired and Closed Loop was toggled back on.

Image

Image

trioneer-dev:
Closes #644. Supersedes the docs' manual "set Max IOB to 0, Autosens to 100%" recipe.

**Low Glucose Suspend** clamps Max IOB to 0. oref's reduction stages all run before the `iobGreaterThanMax` guard, so reductions stay intact and only correction and SMB are neutered.

**Basal Testing** adds pinned Autosens, Dynamic ISF off, and a 72 mg/dL hypo threshold. Past the low-glucose suspend it recommends nothing — leaving `rate`/`duration` unset rather than commanding the scheduled rate, which would record as a temp basal and litter the history the test exists to read. Entering the mode clears a running Trio-set temp.

Clamps are **derived overrides** applied at two `OpenAPS` call sites. Stored preferences are never written, so switching back restores everything and manual dosing keeps honouring the real Max IOB.

Verified on device: with stored Max IOB 7 / Autosens 0.7–1.2 / Dynamic ISF on, `profile.json` shows `max_iob 0`, `autosens 1–1`, `useNewFormula false`, `threshold 72`, and `preferences.json` is untouched.


trioneer-dev:

PR #1501, opened 2026-09-12
Name the dosing mode in the loop status sheet

trioneer-dev:
The sheet explained closed-loop behaviour in every mode. It now names the active mode and states what it did with the determination — open loop enacted nothing, LGS may only lower basal, Basal Testing holds unless glucose goes low. Closed loop gets no caveat.

The glossary defines Sigmoid and Logarithmic Dynamic ISF, which Basal Testing switches off, so it now says when those terms cannot appear.

Also fixes black-on-grey badge text in dark mode.


trioneer-dev:

PR #1502, opened 2026-09-12
Report the dosing mode to Nightscout

trioneer-dev:
Open loop and a failed enactment both send no `enacted` block, so followers could not tell a system deliberately not dosing from one that tried and failed.

Devicestatus now carries `openaps.dosingMode` as the raw value, not the localized name, since followers parse it. Not added to the profile upload, which mirrors the standard NS profile schema.

Tests pin the four raw values so a rename cannot silently break followers.


trioneer-dev:

PR #1503, opened 2026-09-12
Show what the algorithm would give in open loop

trioneer-dev:
Open loop computes `insulinReq` every cycle and never enacts it, and nothing surfaced it.

The bolus screen now shows it alongside the calculator's own number, discounted by the user's Recommended Bolus Percentage and capped at Max Bolus — not capped again at Max IOB, which oref already applies internally.

Never `smbToDeliver`: oref pairs an SMB with a compensating low temp that open loop does not enact, so showing it alone would be half a dosing decision. The row hides when `insulinReq` is negative, which is oref asking for basal reduction, not a bolus.

Also removes `insulinForManualBolus` — read in two places, written nowhere, so both always returned 0.


trioneer-dev:

PR #1504, opened 2026-09-12
[Dosing Modes – Part 1 of 6] Replace the Closed Loop toggle with a Dosing Mode

dnzxy:
Stacked on #1300 — diff includes it until that merges.

`closedLoop: Bool` becomes `DosingMode`: `closed`, `open`, `lowGlucoseSuspend`, `basalTesting`.

Each mode exposes an `AutomationLevel` (`off` / `hypoSuspendOnly` / `reductionsOnly` / `full`), so nothing lumps the constrained modes in with closed loop. Only `closed` is `full`.

Existing users migrate via the settings decoder's legacy-key path: `true` → closed, `false` → open. Unknown values fall back to open.

Also removes the unreachable `ManualTempBasal` module, and pre-fills Max IOB during onboarding with a third of the entered basal schedule instead of 0.

Closed and open behave exactly as the toggle did.


t1dude (approved):
Mode selection works as expected. My live setup went into closed loop mode. A simulation starting imwith closed loop off, went into open loop mode.

PR #1507, opened 2026-09-12
[Dosing Modes – Part 4 of 6] Name the dosing mode in the loop status sheet

dnzxy:
The sheet explained closed-loop behaviour in every mode. It now names the active mode and states what it did with the determination — open loop enacted nothing, LGS may only lower basal, Basal Testing holds unless glucose goes low. Closed loop gets no caveat.

The glossary defines Sigmoid and Logarithmic Dynamic ISF, which Basal Testing switches off, so it now says when those terms cannot appear.

Also fixes black-on-grey badge text in dark mode.

| Open loop | Basal Testing |
|---|---|
|  |  |

Simulator has no pump, hence "No recent determination".


t1dude (approved):
Tested in live setup. Works as expected.

PR #1512, opened 2026-09-13
Fix for Libre 3/3 Plus: Stop dropping display-only glucose readings

### LINKED ISSUE #1495: Libre 3 Plus readings received but marked notActionable/display-only, causing No Recent Glucose
## I am using a FreeStyle Libre 3 Plus directly with Trio dev.
The Libre connects successfully by Bluetooth and LibreLoop continues receiving glucose values. However, Trio eventually shows “No Recent Glucose” and the loop goes red/stops looping.
The diagnostic log shows actual glucose values being received with dq=good and cond=ok, but they are classified as:
act=notActionable
LibreLoop then forwards the readings as:
isDisplayOnly: true
Trio subsequently drops the display-only reading, so the glucose value is not available for looping.
Bluetooth may still show Connected while Trio shows No Recent Glucose.
Setup
CGM: FreeStyle Libre 3 Plus
Trio: v0.8.4.93 (3)
Branch: dev
Trio commit: a708381
LibreLoop submodule: b66138f
The problem has continued despite Bluetooth reconnecting. The Libre is still transmitting numeric glucose values, so this does not appear to be simply loss of Bluetooth communication.

## Attach a Log
*Tap the Trio settings icon at the bottom of the screen, then tap 'Share logs' on the bottom of the list and attach it to this ticket.*

## To Reproduce
1. *Go to '...'*
2. *Click on '....'*
3. *Scroll down to '....'*
4. *See error*

## Expected behavior
*A clear and concise description of what you expected to happen.*

## Screenshots
*If applicable, add screenshots to help explain your problem.*

### Smartphone:
* Hardware: *[e.g. iPhone 15 Pro]*
* OS Version: *[e.g. iOS 17.5]*

### Pump:
* Manufacturer: *[e.g. Insulet]*
* Model: *[e.g. Omnipod Dash or Eros]*

### CGM:
* Device: *[e.g. Dexcom G7]*
* Manager app: *[e.g. Dexcom App or xDrip4iOS]*

### Trio Version:
* Version Number: *[e.g. 1.9.2]*
* Repo: *nightscout/trio*
* Git Reference: *[e.g. commit hash]*

## Technical Details

## Additional context

dnzxy:
`isDisplayOnly` means the reading was shifted for visual consistency after calibration (LoopKit), not that it is unsafe to dose on: Loop keeps these samples and excludes them only from momentum and counteraction. The gate was LibreLoop-scoped, so Trio already dosed on Dexcom's display-only samples.

It never held either way: LibreLoop's backfill paths forward the same sensor minute with isDisplayOnly false, so a dropped reading re-entered unflagged. On Libre 3 Plus the actionability bit is stuck clear, so every reading was discarded and those users could not loop at all.

**Readings the sensor reports as faulty are still withheld by the driver.**

Fixes #1495 and similar reports.


MikePlante1 (approved):
I could not produce a isDisplayOnly reading to test this, but approving via code review as the fix is simple and targeted.

PR #1513, opened 2026-09-13
Update CODEOWNERS to use github groups

jwoglom:
Modify the .github/CODEOWNERS file to change the codeowners automatically assigned to PRs. Moved from a single global list to rough per-component groupings, which are defined in github teams rather than hardcoded in the file.


Sjoerd-Bo3 (approved):

t1dude:
I'm okay with this as long as we are all put in the right groups.

dnzxy (approved):

PR #1515, opened 2026-09-13
Fix index-out-of-range crash when deleting a schedule entry

### LINKED ISSUE #1497: Crash: iPhone 16/26.5 deleting a time segment for ISF or glucose target
## Describe the bug
Crash
OP 5 / iphone 16 / 26.5
commit hash id fb2b136 https://github.com/gviny234-org/Trio/commit/fb2b1360bd64b8b9017f1a9e84bc008a9e7d0d9b

## Attach a Log

[log.txt](https://github.com/user-attachments/files/32148360/log.txt)
[Trio-2026-09-10-084636.ips.zip](https://github.com/user-attachments/files/32148508/Trio-2026-09-10-084636.ips.zip)

## To Reproduce
My apologies for being a newbie - it was either a glucose target or ISF, I *think* it was ISF. 
I had two, one from 12a 118, one from 7a 110.
I swiped left to delete the 7am one and it crashed.

### Smartphone:
* Hardware: iPhone 16
* OS Version: iOS 26.5

### Pump:
* Manufacturer: Insulet
* Model: Omnipod 5

### CGM:
* Device: Dexcom G7
* Manager app: Dexcom app

### Trio Version:
* Version Number: v0.8.4 
* Repo: nightscout/trio
* Git Reference: fb2b136 

## Additional context
First bug report, I'm sure I missed something with formating or other. I'm a former engineering director but with brainfog. Just guide me on your preferences for optimium efficiency. #nerd.

dnzxy:
## Summary

Deleting a time segment in the ISF, glucose target, carb ratio or basal editor could kill the app with `Fatal error: Index out of range`. 

The fix takes index bindings out of the therapy settings item row entirely:

  - `ForEach(items)` iterates values, so `item.id` in an action is a captured `UUID` rather than a binding read
  - `deleteItem(with:)` resolves the index by id at call time — used by the swipe action, the context menu and the VoiceOver delete action (which previously also read `item.id` through the binding)
  - the picker row gets an id-keyed binding whose setter no-ops if the entry is gone
  - `validateTherapySettingItems()` keeps ids stable; selection falls back to time-matching only when the selected entry
  actually disappeared

This fix touches the shared editor for all therapy settings, so it covers ISF, glucose target, carb ratio and basal editors as well as the four onboarding therapy-setting steps.

Fixes #1497


t1dude (approved):
Works as expected

PR #1518, opened 2026-09-14
Add option to hide home stats panel metrics

trioneer-dev:
- Spell out Avg. Glucose instead of the ambiguous ⌀ symbol (#1474)
- Hide Statistics face: static View Statistics shortcut, no data

image

PR #1520, opened 2026-09-14
Add more options for statistics panel

jwoglom:
The home statistics panel offered three faces, all hard-wired to the current calendar day. This adds three more and makes the reported time configurable.

New options
- Looping Performance: the share of glucose readings Trio looped on, drawn as a green bar — the same pair the Statistics screen charts as "Successful Loops" against "Glucose Count".

IMG_2336

- Total Daily Dose: insulin delivered and carbs entered, styled like the Averages face.

IMG_2337

- None: just "View Statistics", for people who would rather not see numbers on the home screen.

IMG_2335

New setting, Features > User Interface > Statistics Range: Today, Last day, Last week, Last month or Last 3 months. It drives the panel's numbers and its wording ("Today's average" becomes "Last week's average"), and the Statistics screen now opens on the same span:
IMG_2338

The panel's data moves off the chart's 72 h fetch into its own background computation (HomeStatsPanelStats), since the longer ranges read months of history.


t1dude:
This overlaps a bit with #1518

Sjoerd-Bo3:
Looks good, only the tdd one feels off. It doesnt adhere to the other ones markup? Maybe making the title (total daily dose) the first line and big makes it better?

dnzxy:
This PR was superseded by #1518 and holds the same issue with insulin accounting that 1518 had, which is why it was removed and only the hide option was kept. 

I‘ll close this one. We can revisit a reworked and corrected version in a new PR.

bjornoleh:
I hope the selectable time windows for stats will be added in a separate PR.

PR #1523, opened 2026-09-15
Fixing Loop Mode help sheet

t1dude:
Picking up the fix from Elias (thank you!) that solves the Loop Mode help sheet opening blank the first time it is opened.

Adding one sentence to clarify how LGS works.


trioneer-dev (approved):

MikePlante1 (approved):
Tested in-vivo. Works as expected. 👍

PR #1539, opened 2026-09-17
Fix x-axis label alignment in stats and main charts for SDK 27

### LINKED ISSUE #1525: x-axis labels misaligned in stats
## Describe the bug
The x-axis labels in statistics are not aligned properly.

## Attach a Log
n/a

## To Reproduce
1. Open stats to a Glucose (by day) or insulin or meals chart
2. Open your eyes

## Expected behavior
x-axis labels should appear and be aligned as they used to

### Smartphone:
* Hardware: iPhone 17 Pro Max
* OS Version: iOS 27.0 (full release)

### Pump:
* Manufacturer: Tandem
* Model: Mobi

### CGM:
* Device: Dexcom G7
* Manager app: DexKit

### Trio Version:
* Version Number: 0.8.4.109
* Repo: *mikeplante1/trio*
* Git Reference: [93fe2e8
](https://github.com/MikePlante1/Trio/commit/93fe2e83da8bb08e3946c0c6ed85cc69278bb936)

## Technical Details
Built with Xcode 27 (I presume it'd look fine building with Xcode 26, but have not tested)

MikePlante1:
Fixes #1525 

## Summary

Building Trio with Xcode 27 to iOS 27 shifts x-axis labels out of alignment across the
statistics charts and the main chart, and hides them entirely on the glucose
Distribution (by day) chart.

Three Swift Charts behaviours changed in the iOS 27 SDK. Each is gated on the
**linked SDK**, not the OS: the same source built with Xcode 26 renders
correctly on an iOS 27 device, so this only started affecting users once Trio
was built with Xcode 27.

| Swift Charts API | Behaviour under SDK 27 |
|---|---|
| `AxisValueLabel(centered: true)` | Centres across the span to the next *labelled* mark instead of one stride unit |
| `AxisValueLabel(anchor:)` | Silently ignored unless `preset: .aligned` is also set |
| `.chartYAxisLabel` | Consumes the x-axis label row when combined with `.chartScrollableAxes` |

## Changes

**SDK 27 fixes**

- **Stats charts** — every chart strides finely and labels sparsely (stride by
  day, label weekly), so the `centered:` change pushed labels right by half the
  *label* spacing: ~3.5 days in month view, ~half a month in 3M, 3 hours in day
  view. Five near-duplicate axis blocks are replaced with one
  `StatChartUtils.dateAxisMarks(for:)` that emits an `AxisValueLabel` on every
  stride value — empty where no label is wanted — so "next labelled mark" once
  again equals "next stride value". That is correct under both the old and the
  new semantics, and a no-op on iOS 26. The mark closing the domain is left
  unlabelled so `preset: .aligned` stops pinning its centred label on top of the
  preceding one (the `TueWe` overlap in week view, the stray clipped `0` in day
  view).
- **Main chart** — already passed `anchor: .top`; adding `preset: .aligned`
  makes it take effect again. Verified the glucose, basal and COB/IOB panes stay
  column-aligned: gridline positions are identical across all three panes with
  and without the preset.
- **Distribution (by day)** — this is the one chart that is both scrollable and
  used `.chartYAxisLabel`, so it is the only one whose x-axis labels disappeared.
  Removing the label fixes it; a y-axis label cannot be reinstated on that chart
  while it is scrollable.

**Deliberate changes (not regressions)**

- **Centre on-mark labels** — Percentile/Distribution by time and Looping
  Performance had left-aligned labels sitting ~5pt right of their mark on *every*
  SDK. They are now centred via `preset: .aligned` + `anchor: .top`.
- **Label cleanup** — dropped the redundant "Percentage" labels and the
  "Glucose Distribution" heading, and added the glucose units above the y-axis of
  Percentile (by day), which had none.

Bar charts keep `centered: true` throughout: their labels measure 0px from bar
centre, which is the right target for a bar. Only point/area charts centre on the
mark.

## Testing

Each configuration was measured against rule marks placed at the exact target
position, on the iOS 26.5 and iOS 27.0 simulators, built against both SDK 26.5
and SDK 27.0:

| | SDK 26.5 build | SDK 27 build, before | SDK 27 build, after |
|---|---|---|---|
| `centered:` span | 0px | +115px | 0px |
| `anchor: .top` | 0px | +32px | 0px |
| main chart | 0px | +32 to +36px | 0 to −1px |

The old and new label-selection logic were also compared exhaustively — 90 days
of axis marks × all four intervals × both original code paths — with zero
unintended differences; the only changes are the intended trailing-mark trims.


dnzxy (approved):
Still looks good on SDK 26 (Xcode 26) and iOS 26, so LGTM.

t1dude:
...testing on MacOS 27 + Xcode 27.

t1dude (approved):
Works as expected, tested using MacOS 27, Xcode 27, iOS 27.

PR #1548, opened 2026-09-18
Add Dexcom G7, ONE+ and Stelo Direct Sensor Connection

dnzxy:
## Summary

This PR introduces new connectivity method for Dexcom CGMs: It lets Trio connect straight to G7-family sensors over Bluetooth. No Dexcom app, no Dexcom account, no second phone relaying readings.

### What this gives users
  - Pairing with the 4-digit applicator code, typed or scanned from the Data Matrix on the applicator.
  - Full onboarding: site selection, insertion and overpatch guidance, then pairing.
  - Migration path: the existing read-through-the-Dexcom-app (i.e., passive consumption) mode stays for anyone mid-session without their code to hand. Settings shows when a direct connection is available and offers the switch, keeping the current session and its readings.
  - Session management: per-cycle reads, backfill of any gap, sensor model, serial, firmware, and 10- vs 15-day session length detected from the sensor itself.
  - Lifecycle alerts from Trio: expiring (24 h / 2 h), expired, session ended, sensor failed, signal loss, connection refused. All alerts get retracted when the sensor is replaced or the CGM removed.
  - Calibration with guidance on when it is and isn't appropriate, and a warning before sending a value far from the sensor's own.
  - Settings: lifecycle bar, last reading and trend, sensor details, previous-sensor record, and a device log of connections, commands and handshake steps.

  ## Credit

This exists because of Erik Tolboom (@ETolboom) and Pete Schwamb (@ps2).

The direct G7 stack is Erik's work, built in `DexKit`, a hybrid project combining `G6SensorKit` work on direct G6/ONE integration with his own on G7, ONE+ and Stelo. `DexKit` ran in closed testing with a group of Trio users from early September, and the protocol was established the hard way: reverse-engineering sessions, BLE captures, and findings dug out of tester logs.

Later, `DexKit` development was halted. The decision was made to enhance `G7SensorKit` rather than adopt `DexKit` as a new, stand-alone CGM manager. Pete went through the effort of collecting the needed parts of the working implementation – i.e., the pairing planner and service, advertisement parsing, handshake sequencing in the authenticator, the BLE manager's active-peripheral and recovery handling, the applicator scanner, several message types, and the sensor application guide's steps and images – and refactored and extended the existing `G7SensorKit` so that we retain the passive pathway while adding this new, direct one. We are following this decision because we believe in a unified iOS OS-AID device infrastructure.

## For Testers

  - One display per sensor. A G7-family sensor admits one display at a time. The Dexcom app must be deleted (or disabled) before pairing. Please follow the onboarding instructions. Switching from the app to Trio needs the established connection ("sensor lease") to expire – force-quit or uninstall the official Dexcom app and wait ~15 minutes.
  - Alerts move to Trio. With direct pairing, the Dexcom app no longer alerts you. Onboarding checks notification permissions.
  - Dexcom Stelo is supported (DX01 advertisements, its own sensor model) but has had the least hardware validation. **We are actively looking for Stelo users to provide test feedback.**
  - Dexcom G6/ONE functionality is untouched by this PR. Where the G6 work from DexKit lands is undecided. Please retain using CGMBLEKit and the passive consumption mode.
  - Verified so far: builds clean, installs, launches, direct onboarding reachable from the existing picker entry. On-device pairing validation is what we need testers for.


LiroyvH:
Question, I cannot build this right now to look for myself and thus have to ask here:
> Migration path: the existing read-through-the-Dexcom-app (i.e., passive consumption) mode stays for anyone mid-session without their code to hand. Settings shows when a direct connection is available and offers the switch, keeping the current session and its readings.

Does this only stay for people "mid-session" or does it stay available as a choice? Eg: users who prefer to use the Dexcom-app (eg: due to different alarms, Direct to Watch features, Dexcom Share/Clarity uploads, whatever), will they continue having the ability to use these sensors in passive mode or will they be forced to use the direct connection and abandon the Dexcom app?

loudestnoise:
At present, there's no requirement to do a swap like you're asking. Can't speak for development plans, but I can't see why removing the long-standing way OS-AID apps passively get sensor readings. Some people (like myself) just prefer to not have to use the Dexcom app.

And you can do a brand new sensor pair, or switch over. The key is disabling (bluetooth access specifically) for Dexcom and then wait up to 15 minutes after entering the 4-digit code in Trio. I've done one mid-sensor swap, and one new sensor and both work just like you'd expect them to.

munchyk:
In my limited testing, “offloading” and “reinstalling” the Dexcom app through iOS storage settings allows one to quickly switch back and forth between the Dexcom app and Trio’s direct connect without needing to wait 15 min. The benefit of “offloading” over “uninstalling” is that it retains all of the app data. After you reinstall the app, everything appears just as you left it (including the last paired sensor code which allows for the quick reconnect). Though, disabling Bluetooth for the Dexcom app may be simpler/easier

loudestnoise:
Disabling Bluetooth & **Notifications** because otherwise it is annoying and sends you Critical Alerts 😜

LiroyvH:
@loudestnoise Yes I'd argue in favour of a similar option as xDrip4iOS offers, which asks the user if they want to connect to the sensor directly or passively ("co-existence mode"). Some people prefer direct connections to get rid of the Dexcom-app, some prefer to keep the Dexcom-app (for various reasons, including direct to watch). Offering that choice to the user would be nice. :) We already have the code for passive connections anyway. 

Of course you could have Dexcom + xDrip4iOS (co-existence) and then have Trio read from xDrip4iOS... But that forces users to install another app to retain their Dexcom specific features, whilst they might otherwise not have needed or wanted xDrip4iOS in between. This is also more resource intensive and another app that could crash or lock up.  

Putting this all together, I'd argue allowing the user to choose between the direct (new option) and passive (existing option) connection would be the best way forward to ensure the best user experience and freedom of choice. 

Again, I haven't been able to build this yet so apologies if its already designed like this and offers users a choice.

ETolboom:
Hey @LiroyvH .

Pete (@ps2) has implemented G7SensorKit in the exact way you mention right now. Existing users will continue to use G7SensorKit as usual (passive). When pairing a new sensor, however, G7SensorKit now has two modes: direct pairing (new), as well the passive eavesdropping mode. 

Like you mentioned, there are features like Dexcom Share/Clarity as well as Direct-To-Watch are currently not supported yet. As such, both options should (and are) being offered. 

For those suffering frequent connectivity issues with the eavesdropping (and that do not mind sacrificing on Share/Clarity/DTW), this PR is being offered already.

munchyk:
This _is_ G7SensorKit. Not sure where anyone got the idea that passive eavesdropping was being removed (though this may eventually change).

This simply adds another choice for users to connect directly to the G7 (instead of eavesdropping) which results in a more reliable and consistent connection.

A side point was that, as of the current implementation, direct connection requires deleting/disabling the Dexcom app. A direct connection can only be maintained by one app at time.

LiroyvH:
@ETolboom Splendid, that's great - thank you for clarifying! :) 

@munchyk Didn't necessarily get that idea, but was unsure - hence the question _if_ its being removed or not (due to the wording in the PR). Apparently not, so all good! :) -edit- I don't think it should be removed in the future either, users must always have freedom of choice and its not problematic to leave the passive mode in there imo. :)

dnzxy:
Thanks for adding context here @ETolboom .

@LiroyvH @loudestnoise what you see here right now , and so properly described also by @munchyk , is what's been implemented as of https://github.com/LoopKit/G7SensorKit/pull/67, not necessarily the final end result.

There is work being done to not require a user to delete the OEM's app when going for the direct-connection path, and I personally believe that we cannot expect users to do this. However, for now, this is what this _in-development_ feature ships with. Similar thing for current implementation limitations like direct-to-watch and upload to Clarity/Share.

Since this is cross-project work it involves coordination with multiple people, and testing across apps, bear with it and us, it may take a moment or two. We'll collectively get there ☺️ .

LiroyvH:
@dnzxy Yup, understood and no worries - was just concerned, due to the wording in the PR, that passive mode was going to be removed as an option; thereby forcing users to go to direct mode and removing freedom of choice to use the Dexcom app (for whatever personal motivation someone may have for doing so at this time (including the aforementioned features)). Very happy to hear that is, at the very least at this time, not the case and people will continue to be able to use the "eavesdropping" mode. :) Thanks!

dnzxy:
Yes absolutely. 

Just a side note, if you'll allow it: freedom of choice is non-existent with all new CGMs added to Trio (Eversense, AccuChek SmartGuide, Libre 3/3+), largely due to limitation posed by the sensors themselves. So, while I can see the appeal to what you are saying here, the reality is that we are going toward – and also actively working towards – a future where the AID app will hold the _active_ connection to sensors.

PR #1552, opened 2026-09-18
Adjust Not-Looping Alerts Escalation

dnzxy:
## Summary

Changes the not-looping alert from a single 20 min critical alarm into a six-step escalation scheme:

* Time-Sensitive at 20, 40, 60, 80, and 100 min
* Critical at 120 min
* All six are armed up front as OS notifications, so escalation still works if iOS ever suspends Trio
* Any successful loop retracts the full ladder and starts fresh

Each warning gets its own alert ID (`w1`…`w5`) because pending notifications with the same identifier replace each other rather than stack.

Also updates device alarm defaults:

* Critical defaults to sound on, using `high_chimes.caf`
* Time-Sensitive and Normal default to silent
* Decode fallback now follows the per-tier default
* Glucose alarm sound defaults are unchanged

Tests were updated for the new ladder and sound defaults, and the unused `AlertMuter.allowedDurations` was removed.

### Note

Existing persisted device alarm settings are left untouched, so current users keep their existing `playsSound` / `soundFilename` values. Moving them to the new defaults would need an explicit migration.


AndreasStokholm (approved):
After @dnzxy's fix in [ec00e90](https://github.com/nightscout/Trio/pull/1552/commits/ec00e90da1e204b40d23e6121da608aec31f630f) this works wonderfully 👏

It's also got a much less poop-your-pants-in-public alert sound by default, which I appreciate a lot!

